PR #7, opened 2020-12-10
[webview_flutter] Add webview_flutter initial package

bwikbs:


packages/webview_flutter/README.md
```diff
@@ -0,0 +1,48 @@
+
+# webview_flutter_tizen
```
bbrto21, line 2: Does it only work on arm targets, not x86(emulator)? If so, could you add some guide?

bwikbs: This is a commit to work on the dev branch. The parts you mentioned will be worked on before entering the master.

bwikbs: Since it is still in the early stages of development, there are many areas to be decided.(Including what you mentioned above)

bwikbs: BTW, how did you know this PR came up? Do you get notified by email or something :question:

bbrto21: Yes, Gmail forwards notices to Knox mail

seungsoo47 (approved):
LGTM

PR #9, opened 2020-12-13
Add a workflow for static analysis of Dart code

swift-kim:
- Use concrete version dependencies on platform interface packages (because platform interfaces are not necessarily backward compatible)
- Update analysis_options.yaml and resolve warnings
- Add a workflow (Dart)

The workflow verifies Dart code only. More checks will be added later for full analysis of other requirements.


bbrto21 (approved):
LGTM

PR #11, opened 2020-12-21
Handle key events for tizen webview

seungsoo47:
* Implement dispatchKeyDownEvent and dispatchKeyUpEvent
* Translate ecore key values to KeyValues for webview

### DIFF
--- a/packages/webview_flutter/lib/webview_flutter_tizen.dart
+++ b/packages/webview_flutter/lib/webview_flutter_tizen.dart
@@ -275,7 +275,11 @@ class PlatformViewsServiceTizen {
       case 'viewFocused':
         final int id = call.arguments as int;
         if (_focusCallbacks.containsKey(id)) {
-          _focusCallbacks[id]();
+          if (_focusCallbacks[id] != null) {
+            _focusCallbacks[id]();
+          } else {
+            throw FlutterError('FocusCallbacks[$id] must not be null.');
+          }
         }
         break;
       default:
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -18,6 +18,9 @@
 #include "lwe/PlatformIntegrationData.h"
 #include "webview_factory.h"
 
+#include <Ecore_Input_Evas.h>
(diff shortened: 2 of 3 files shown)


packages/webview_flutter/lib/webview_flutter_tizen.dart
```diff
@@ -275,7 +275,9 @@ class PlatformViewsServiceTizen {
       case 'viewFocused':
         final int id = call.arguments as int;
         if (_focusCallbacks.containsKey(id)) {
-          _focusCallbacks[id]();
+          if (_focusCallbacks[id] != null) {
```
bbrto21, line 278: Can an item in ```_focusCallbacks``` be ```null```? Should we allow this?
if this case is unacceptable, do we have to find out why it is null?

seungsoo47: Its original code was *_focusCallbacks[id]!()*. I guess the _focusCallbacks' items may be null. Anyway, because current our Flutter SDK does not support the syntax, I changed it to above code. Do you have any good idea?

bbrto21: if original code use postfix operator '!' , maybe we should throw an exception.
The result of the below is ```Error: Null check operator used on a null value``` 

``` dart
void main() {
  int? aNullableInt = null;
  try {
    print( aNullableInt!.isOdd);
  }catch(e){
    print('Error: '+  e.toString());
  }
 }

```

bbrto21: Well, this could be a problem later on, but it's still on the devel branch and the null safety language feature is experimental so I think we can move on at this time.

bwikbs: same here!

bwikbs (approved):

packages/webview_flutter/lib/webview_flutter_tizen.dart
seungsoo47: Fixed

bbrto21 (approved):
LGTM

PR #15, opened 2021-01-06
[webview_flutter_dev] Change webview interface && embedding lwe.so file

bwikbs:
Now, we can run webview test app without any preparations

Just run ``flutter-tizen run`` at example.

### DIFF
--- a/packages/webview_flutter/tizen/project_def.prop
+++ b/packages/webview_flutter/tizen/project_def.prop
@@ -16,12 +16,12 @@ USER_CPP_UNDEFS =
 
 # Compiler/linker flags
 USER_CFLAGS_MISC =
-USER_CPPFLAGS_MISC = -c -fmessage-length=0
-USER_LFLAGS = -llightweight-web-engine
+USER_CPPFLAGS_MISC = -c -fmessage-length=0 
+USER_LFLAGS = -llightweight-web-engine.mobile -Wl,-rpath=/opt/usr/globalapps/org.tizen.webview_flutter_tizen_example/lib/arm
 
 # Libraries and objects
 USER_LIB_DIRS = lib
-USER_LIBS = lightweight-web-engine
+USER_LIBS = lightweight-web-engine.mobile
 USER_OBJS =
 
 # User includes
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -21,6 +21,16 @@
 #include <Ecore_Input_Evas.h>
 #include <Ecore_IMF_Evas.h>
 
+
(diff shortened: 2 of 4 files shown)


packages/webview_flutter/tizen/project_def.prop
```diff
@@ -16,12 +16,12 @@ USER_CPP_UNDEFS =
 
 # Compiler/linker flags
 USER_CFLAGS_MISC =
-USER_CPPFLAGS_MISC = -c -fmessage-length=0
-USER_LFLAGS = -llightweight-web-engine
+USER_CPPFLAGS_MISC = -c -fmessage-length=0 
+USER_LFLAGS = -llightweight-web-engine.mobile -Wl,-rpath=/opt/usr/globalapps/org.tizen.webview_flutter_tizen_example/lib/arm
```
bbrto21, line 20: I think that this path is for example app only

bwikbs: As you said, that's right. This is just temporary code for convenience and we need an extensive discussion of the lwe distribution method.

bbrto21: ok, thank you for explanation

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -438,23 +448,16 @@ void WebView::InitWebView() {
   }
   float scaleFactor = 1;
```
bbrto21, line 449: There is some routine to calculate the scale factor on the engine side, Should it be shared by the engine? 
This is just my curiosity.

bbrto21 (approved):

seungsoo47:
The liblightweight-web-engine.mobile.so file is on example directory. However, I think this library has to be located on plugin side. How do you think of this? And how do you think of removing the word "mobile" in the name?

bwikbs:
Actually. so file is located on both sides.(plugin & example). Only reason it exists in the example side is purely for the convenience of execution.  So it seems like the problem of naming is not big deal at this moment I guess. we should think about how to distribute lwe first.

packages/webview_flutter/tizen/src/webview.cc
bwikbs: I guess.. we should use engine's value. we need to modify platform view's interface for that.

bwikbs:
Here are the results of running this for each profile.

- TW3 -> There is no big problem.
- TV -> We can't load so of apps, ``crash`` occurs after short running, re-compilation required
- TM1 -> The screen flickers and is very unstable.
- FHUB ->  Prertty much the same with ``TM1``

seungsoo47 (approved):
LGTM

PR #19, opened 2021-01-20
[webview_flutter] Update webview to be focused

seungsoo47:
* A webview has to act like a FocusNode. So, when it is clicked or released, a FocusCallback function should be called.


swift-kim:
Could you add a plugin name prefix like [webview_flutter] in the PR title? Thanks!

seungsoo47:
I have updated it. Thank you.

bwikbs (approved):

PR #36, opened 2021-02-02
[webview_flutter] Add NavigationDelegate and onWebResourceError

seungsoo47:
* The NavigationDelegate handler decides how to handle navigation actions
* When the handler is null, all navigation actions are allowed
* Implement the onWebResourceError callback to report web resource loading error
* The callback's error codes use ones provided by lightweight-web-engine on tizen

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -26,6 +26,99 @@ extern "C" size_t LWE_EXPORT createWebViewInstance(
     const std::function<void(::LWE::WebContainer*, bool isRendered)>&
         renderedCb);
 
+template <typename T = flutter::EncodableValue>
+class NavigationRequestResult : public flutter::MethodResult<T> {
+ public:
+  NavigationRequestResult(std::string url, WebView* webView)
+      : url_(url), webView_(webView) {}
+
+  void SuccessInternal(const T* shouldLoad) override {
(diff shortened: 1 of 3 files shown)


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -82,6 +177,14 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
     currentUrl_ = "about:blank";
   }
 
+  auto settings = params[flutter::EncodableValue("settings")];
+  if (std::holds_alternative<flutter::EncodableMap>(settings)) {
+    auto settingList = std::get<flutter::EncodableMap>(settings);
+    if (settingList.size() > 0) {
+      ApplySettings(settingList);
```
bbrto21, line 184: ApplySettings can throw an error and This is in the constructor. Where should the error be handled?

bwikbs: Why did you mention the constructor? I can't get your point

bwikbs: https://stackoverflow.com/questions/5743013/exception-handling-a-constructor.Now I got your point.

seungsoo47: I fixed that it is handled in webview_factory.cc. Thanks

bbrto21 (approved):
LGTM

bwikbs (approved):

PR #41, opened 2021-02-23
[url_launcher] Update platform interface to 1.0.9

swift-kim:
The `linkDelegate` property has been added in 1.0.9 to support the newly proposed Link widget. For more information on it, see the relevant changes (#3154, #3177) in the [flutter/plugins](https://github.com/flutter/plugins) repo and the [source code](https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher/lib/src/link.dart).

`null` can be returned to enable a fallback to `DefaultLinkDelegate` on non-web platforms.

### DIFF
--- a/packages/url_launcher/CHANGELOG.md
+++ b/packages/url_launcher/CHANGELOG.md
@@ -5,3 +5,9 @@
 ## 1.0.1
 
 * Migrate to Tizen 4.0
+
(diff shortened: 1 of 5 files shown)


bbrto21 (approved):
LGTM

PR #42, opened 2021-02-23
[webview_flutter] Optimize tbm allocation

bwikbs:


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -719,17 +720,22 @@ void WebView::InitWebView() {
       0, 0, width_, height_, scaleFactor, "SamsungOneUI", "ko-KR", "Asia/Seoul",
       [this]() -> LWE::WebContainer::ExternalImageInfo {
         LWE::WebContainer::ExternalImageInfo result;
-        tbmSurface_ = tbm_surface_create(width_, height_, TBM_FORMAT_ARGB8888);
+        if (needTbmAllocate_) {
```
bbrto21, line 723: I think that it can be replaced by checking whether ```tmbSurface_``` is ```nullptr``` or not

bwikbs: @bbrto21 ^^ good point :clap:

bbrto21 (approved):
LGTM

PR #46, opened 2021-03-03
[camera] Initial commits for camera plugin

bbrto21:
This PR includes:
* [camera] Initial commit for camera
* [camera] Copy an example from original
* [camera] Add tizen host for exmple
* [camera] Establish camera channel and cameraEvents channel
* [camera] Fix build error


packages/camera/example/integration_test/camera_test.dart
```diff
@@ -0,0 +1,238 @@
...
+        await controller.initialize();
+        final bool presetExactlySupported =
+            await testCaptureImageResolution(controller, preset.key);
+        assert(!(!previousPresetExactlySupported && presetExactlySupported),
+            'The camera took higher resolution pictures at a lower resolution.');
+        previousPresetExactlySupported = presetExactlySupported;
+        await controller.dispose();
+      }
+    }
+  }, skip: !Platform.isAndroid);
```
wiertel, line 93: does this mean the test will always be skipped? or the example will be multiplatform?

packages/camera/tizen/src/camera_plugin.cc
```diff
@@ -0,0 +1,159 @@
...
+        bool enable_audio;
+        if (!resolution_preset_value.IsNull()) {
+          enable_audio = std::get<bool>(enable_audio_value);
+        }
+
+        LOG_DEBUG("camera_name[%s], resolution_preset[%s], enableAudio[%d]",
+                  camera_name.data(), resolution_preset.data(), enable_audio);
+        flutter::EncodableValue reply = InitializeCameraDevice(
+            camera_name, resolution_preset, enable_audio);
+        result->Success(reply);
```
wiertel, line 92: isn't camera initialization asynchronous in Tizen? I mean, user should wait for a state change. If that's true, couldn't we postpone reporting success until the camera is really ready?

packages/camera/README.md
```diff
@@ -0,0 +1,18 @@
+# camera
+
+A new flutter plugin project.
```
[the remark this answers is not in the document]
bbrto21: You're right. this plugin in the devel branch and I'm not really concerned about it yet. I'll do it later~

packages/camera/example/integration_test/camera_test.dart
bbrto21: This is a test that we will use when our plugin implementation is working, and it came from the original plugin. I think we don't have to worry about this until then.

packages/camera/tizen/src/camera_plugin.cc
bbrto21: This plugin works with asynchronous communication over channels. so I think that it doesn't matter whether Tizen camera APIs are asynchronous or not. I think you don't have to seriously consider the future work at this stage since concrete implementations will be added over and over again. thank you for review

bwikbs:
Camera plug-ins are now in the early stages of development, and this branch is intended for code sharing among developers.

bwikbs:
I will merge this for further development. 
comments or contributions are always welcome :smile

bwikbs (approved):

PR #53, opened 2021-03-29
[webview_flutter] Update README and pubspec

seungsoo47:
* Update README
* Update pubspec

### DIFF
--- a/packages/webview_flutter/example/pubspec.yaml
+++ b/packages/webview_flutter/example/pubspec.yaml
@@ -1,23 +1,19 @@
 name: webview_flutter_tizen_example
 description: Demonstrates how to use the webview_flutter_tizen plugin.
 
-# The following line prevents the package from being accidentally published to
-# pub.dev using `pub publish`. This is preferred for private packages.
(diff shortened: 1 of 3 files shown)


packages/webview_flutter/example/pubspec.yaml
```diff
@@ -1,29 +1,23 @@
 name: webview_flutter_tizen_example
 description: Demonstrates how to use the webview_flutter_tizen plugin.
 
-# The following line prevents the package from being accidentally published to
-# pub.dev using `pub publish`. This is preferred for private packages.
-publish_to: 'none' # Remove this line if you wish to publish to pub.dev
-
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.2.2 <3.0.0"
```
bwikbs, line 5: Is there any special reason for this? ``2.7.0`` -> ``2.2.2 ``

bwikbs (approved):
LGTM

packages/webview_flutter/example/pubspec.yaml
seungsoo47: There is no reason. I referenced just other plugin's version.

PR #57, opened 2021-04-02
[permission_handler] Add permission_handler_tizen package

swift-kim:
Originally written by @XuWanchaoS, and updated by me based on new changes in the [platform interface](https://github.com/Baseflow/flutter-permission-handler/tree/master/permission_handler_platform_interface). Please see the commit history for the detailed changelog.

Notable change: The "undetermined" status has been [removed](https://github.com/Baseflow/flutter-permission-handler/commit/aa4819a4e4928ffd04ec2f65771b0a78b3686760). The status is now either "denied" or "granted". If a permission is not supported by Tizen, the status is "granted". If a permission is not yet granted to an app, the status is "denied". The behavior is the same as other platforms.

As far as I can see this plugin doesn't support Tizen 4.0. Should we support Tizen 4.0?


swift-kim:
`ppm_request_permissions` is not available on Tizen 4.0 and only `ppm_request_permission` is available instead. Is there a simple way to change the current code to use the singular request API?

HakkyuKim:
> `ppm_request_permissions` is not available on Tizen 4.0 and only `ppm_request_permission` is available instead. Is there a simple way to change the current code to use the singular request API?

Is iterating with `ppm_request_permission` a last resort?

HakkyuKim:
> As far as I can see this plugin doesn't support Tizen 4.0. Should we support Tizen 4.0?

Shall we discuss about this?
FYI, The enum value: `PRIVACY_PRIVILEGE_MANAGER_ERROR_PERMISSION_DENIED` is also missing in Tizen 4.0

I'm facing a similar issue in `audio_player_tizen`, should we use #defines?

swift-kim:
@HakkyuKim No. If your plugin is targeting Tizen 4.0, then you must use Tizen 4.0 APIs only. That's it.

For example, if `PRIVACY_PRIVILEGE_MANAGER_ERROR_PERMISSION_DENIED` is not defined in Tizen 4.0, then simply remove its use cases.

bbrto21 (approved):
I think It seems enough that you wrote that this plugin support 5.5 or later. LGTM

swift-kim:
I'll merge this as it is but please consider porting to Tizen 4.0 - it shouldn't be too hard as long as you have basic understanding of C++ asynchrony.

PR #60, opened 2021-04-07
[permission_handler] Hot fix for glibc dependency issue

swift-kim:
I found [this error](https://github.com/flutter-tizen/flutter-tizen/issues/71) during integration testing. Let me apply this workaround before the plugin is published on pub.dev.

### DIFF
--- a/packages/permission_handler/tizen/src/app_settings_manager.cc
+++ b/packages/permission_handler/tizen/src/app_settings_manager.cc
@@ -22,7 +22,8 @@ AppSettingsManager::AppSettingsManager() {}
 AppSettingsManager::~AppSettingsManager() {}
 
 void AppSettingsManager::OpenAppSettings(OnAppSettingsOpened success_callback,
-                                         OnAppSettingsError error_callback) {
+                                         OnAppSettingsError error_callback)
+{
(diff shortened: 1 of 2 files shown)


packages/permission_handler/tizen/src/app_settings_manager.cc
```diff
@@ -12,6 +12,12 @@ static std::string ErrorToString(int error) {
       return "AppControl - Invalid parameter";
     case APP_CONTROL_ERROR_OUT_OF_MEMORY:
       return "AppControl - Out of memory";
+    case APP_CONTROL_ERROR_APP_NOT_FOUND:
```
bbrto21, line 15: I think we can use ```get_error_message``` instead of ```ErrorToString```

bbrto21 (approved):
I've written a tiny opinion(not important) and generally, it looks good to me

packages/permission_handler/tizen/src/app_settings_manager.cc
swift-kim: Brilliant. I didn't even know such function exists. We may make a separate PR to apply the same thing to other plugins (package_info, image_picker, share, ...) as well.

PR #63, opened 2021-04-08
[google_maps_flutter] Add initial commit for google_maps_flutter plugin

seungsoo47:
* We will start tizen implementation based on a webview because Tizen doesn't have
  a google map components on the platform.


packages/google_maps_flutter/lib/src/webview.dart
```diff
@@ -0,0 +1,840 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+part of google_maps_flutter_tizen;
```
bbrto21, line 6: Why should this plugin have their own Webview implementation?

seungsoo47: There is no special reason. However, this plugin uses platform view components (like AndroidView and UiKitView based on PlatformView) to present a google_map. However, because there is no TizenView on flutter engine, I added it in this plugin.

packages/google_maps_flutter/tizen/src/google_map_controller.h
```diff
@@ -0,0 +1,57 @@
...
+  FlutterTextureRegistrar* texture_registrar_;
+  LWE::WebContainer* webview_instance_;
+  double width_;
+  double height_;
+  tbm_surface_h tbm_surface_;
+  bool is_mouse_lbutton_down_;
+  std::unique_ptr<flutter::MethodChannel<flutter::EncodableValue>> channel_;
+};
+
+#endif  // FLUTTER_PLUGIN_GOOGLE_MAPS_FLUTTER_TIZEN_GOOGLE_MAP_CONTROLLER_H_
```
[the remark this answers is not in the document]
seungsoo47: Done!

packages/google_maps_flutter/example/README.md
```diff
@@ -0,0 +1,16 @@
+# google_maps_flutter_example
+
+Demonstrates how to use the google_maps_flutter plugin.
```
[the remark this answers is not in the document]
seungsoo47: Done!

packages/google_maps_flutter/lib/src/webview.dart
bbrto21: Can't we use webview_flutter tizen implementation?

bwikbs (approved):

packages/google_maps_flutter/lib/src/webview.dart
seungsoo47: To use webview plugin, first of all, it needs to be updated to latest version. Applying to the latest version for webview plugin may be done later.

bbrto21: Can we remove webView implementation from this plugin after migrating webview_flutter to the latest version?
Thank you for your answer.

seungsoo47: Though I am not sure because I have not tried it, it doesn't seems impossible.

PR #65, opened 2021-04-08
[audioplayers] Add audioplayers_tizen package

HakkyuKim:
Originally written by @XuWanchaoS. Please see the commit history for detailed changes.

- target Tizen 4.0 and above
- update license info
- upgrade sdk constraint: dart 2.12 and flutter 2
- update README.md in project root
- fix spelling and incorrect log msg
- organize dependencies
- apply Google C++ naming convention


bbrto21 (approved):
LGTM

bbrto21:
@HakkyuKim 
I've run this example but the behavior looks weird. also, I've seen this log while testing. Could you check one more time?
```
E/ConsoleMessage( 7316): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: PlatformException(player - Invalid parameter, player_set_volume failed, null, null)
E/ConsoleMessage( 7316): #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:581:7)
E/ConsoleMessage( 7316): #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:158:18)
E/ConsoleMessage( 7316): <asynchronous suspension>
E/ConsoleMessage( 7316): #2      AudioPlayer._invokeMethod.<anonymous closure> (package:audioplayers/audioplayers.dart)
E/ConsoleMessage( 7316): <asynchronous suspension>

```

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -0,0 +1,484 @@
+#include "audio_player.h"
+
+#include "audio_player_error.h"
+#include "log.h"
+
+static std::string StateToString(player_state_e state) {
```
swift-kim, line 6: Unused function? Even if it is used, the same thing can be done using just [stringification](https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html).

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -0,0 +1,484 @@
...
+      result = "PLAYER_STATE_PLAYING";
+      break;
+    case PLAYER_STATE_PAUSED:
+      result = "PLAYER_STATE_PAUSED";
+      break;
+  }
+  return result;
+}
+
+static std::string PlayerErrorToString(int code) {
```
swift-kim, line 28: What a long list of errors. You may apply @bbrto21's solution here. https://github.com/flutter-tizen/plugins/pull/60#discussion_r608352262

(and hopefully to other plugins as well if you have time)

README.md
```diff
@@ -12,6 +12,7 @@ The _"non-endorsed"_ status means that the plugin is not endorsed by the origina
 
 | Package name | Original package | Pub | Endorsed |
 |-|-|:-:|:-:|
+| [**audioplayers_tizen**](packages/audioplayers) | [audioplayers](https://github.com/luanpotter/audioplayers) (3rd-party) | [![pub package](https://img.shields.io/pub/v/audioplayers_tizen.svg)]() | No |
```
swift-kim, line 15: Broken hyperlink.

```suggestion
| [**audioplayers_tizen**](packages/audioplayers) | [audioplayers](https://github.com/luanpotter/audioplayers) (3rd-party) | [![pub package](https://img.shields.io/pub/v/audioplayers_tizen.svg)](https://pub.dev/packages/audioplayers_tizen) | No |
```

HakkyuKim:
> @HakkyuKim
> I've run this example but the behavior looks weird. also, I've seen this log while testing. Could you check one more time?
> 
> ```
> E/ConsoleMessage( 7316): [ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: PlatformException(player - Invalid parameter, player_set_volume failed, null, null)
> E/ConsoleMessage( 7316): #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:581:7)
> E/ConsoleMessage( 7316): #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:158:18)
> E/ConsoleMessage( 7316): <asynchronous suspension>
> E/ConsoleMessage( 7316): #2      AudioPlayer._invokeMethod.<anonymous closure> (package:audioplayers/audioplayers.dart)
> E/ConsoleMessage( 7316): <asynchronous suspension>
> ```

The exception is thrown when trying to set volume above 1.0.

At the time, I thought the exception made sense since `player_h` allows volume range between 0.0 and 1.0. The valid volume range is also same for Android and iOS, but Android doesn't throw exception(I haven't tested for iOS). 

Android: https://developer.android.com/reference/android/media/MediaPlayer#setVolume(float,%20float)
iOS: https://developer.apple.com/documentation/avfoundation/avplayer/1390127-volume

Should we not throw exception for consistency? I'm not sure how to handle this. Could you share your opinion on this?

bbrto21:
> Should we not throw exception for consistency? I'm not sure how to handle this. Could you share your opinion on this?

Well.. I think it's right to throw an exception :)

bbrto21:
One more thing, I've seen the call stack as below when I push the back-key to exit example.

```
E/ConsoleMessage(12693): FlutterApplication.cs: Run(39) > Unhandled exception: System.EntryPointNotFoundException: Unable to find an entry point named 'FlutterDestoryWindow' in shared library 'flutter_tizen.so'.
E/ConsoleMessage(12693):    at Tizen.Flutter.Embedding.Interop.DefaultEmbedder.FlutterDestoryWindow(FlutterWindowControllerHandle window)
E/ConsoleMessage(12693):    at Tizen.Flutter.Embedding.Interop.FlutterDestoryWindow(FlutterWindowControllerHandle window)
E/ConsoleMessage(12693):    at Tizen.Flutter.Embedding.FlutterApplication.OnTerminate()
E/ConsoleMessage(12693):    at Tizen.Applications.CoreBackend.UICoreBackend.OnTerminateNative(IntPtr data)
E/ConsoleMessage(12693):    at Tizen.Applications.CoreBackend.UICoreBackend.Run(String[] args)
E/ConsoleMessage(12693):    at Tizen.Applications.CoreApplication.Run(String[] args)
E/ConsoleMessage(12693):    at Tizen.Applications.CoreUIApplication.Run(String[] args)
E/ConsoleMessage(12693):    at Tizen.Flutter.Embedding.FlutterApplication.Run(String[] args)
E/ConsoleMessage(12693):    at Runner.App.Main(String[] args)
E/ConsoleMessage(12693): onSigabrt called

```

swift-kim:
> One more thing, I've seen the call stack as below when I push the back-key to exit example.

@bbrto21 That cannot happen unless you're using a locally built engine with an old version of the tool because the engine commit hash was also updated at the time https://github.com/flutter-tizen/flutter-tizen/pull/64 was merged.

bbrto21:
> > One more thing, I've seen the call stack as below when I push the back-key to exit example.
> 
> @bbrto21 That cannot happen unless you're using a locally built engine with an old version of the tool because the engine commit hash was also updated at the time [flutter-tizen/flutter-tizen#64](https://github.com/flutter-tizen/flutter-tizen/pull/64) was merged.

@swift-kim Oh! I forgot about it!, thank you :) 
@HakkyuKim Don't mind what I just said!

swift-kim:
Never mind :)

packages/audioplayers/tizen/src/audio_player.cc
HakkyuKim: I've updated the code with the approach mentioned in #60. Indeed the code is a lot shorter now. 
I will apply the same solution to other plugins after this pr gets merged.

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -3,111 +3,10 @@
...
-      break;
-    default:
-      result = "player - Unkonwn error";
-      break;
-  }
-  return result;
+static std::string ErrorToString(int code) {
+  const char *ptr = get_error_message(code);
+  std::string result = ptr ? std::string(ptr) : "";
+  return "player - " + (result.empty() ? "Unknown error" : result);
```
swift-kim, line 9: Wouldn't it be better to remove this function itself and just print errors like:

```cpp
LOG_ERROR("error blabla: %s", get_error_message(code));
```

HakkyuKim: If we didn't need a string type, then yes, but the returned string type is also used as a parameter in these places in `audio_player.cc`:

line 131 ~136

```c++
if (result != PLAYER_ERROR_NONE) {
  seeking_ = false;
  std::string error = ErrorToString(result);
  LOG_ERROR("player_set_play_position failed : %s", error.c_str());
  throw AudioPlayerError(error, "player_set_play_position failed");
}
```

line 313 ~319

```c++
void AudioPlayer::HandleResult(const std::string &func_name, int result) {
  if (result != PLAYER_ERROR_NONE) {
    std::string error = ErrorToString(result);
    LOG_ERROR("%s failed : %s", func_name.c_str(), error.c_str());
    throw AudioPlayerError(error, func_name + " failed");
  }
}
```

line 378 ~382

```c++
void AudioPlayer::OnErrorOccurred(int code, void *data) {
  std::string error = ErrorToString(code);
  LOG_ERROR("error occurred: %s", error.c_str());
  AudioPlayer *player = (AudioPlayer *)data;
  player->error_listener_(player->player_id_, "error occurred: " + error);
}
```

For me, having a helper function looks better than creating a string type of `get_error_message` multiple times.
Of course, there might be a better way, and I'm open to other suggestions.

swift-kim: Hmm. I think

```cpp
if (result != PLAYER_ERROR_NONE) {
  seeking_ = false;
  LOG_ERROR("player_set_play_position failed : %s", get_error_message(result));
  throw AudioPlayerError(get_error_message(result), "player_set_play_position failed");
}
```

or

```cpp
void AudioPlayer::OnErrorOccurred(int code, void *data) {
  std::string error(get_error_message(code));
  LOG_ERROR("error occurred: %s", error.c_str());
  AudioPlayer *player = (AudioPlayer *)data;
  player->error_listener_(player->player_id_, "error occurred: " + error);
}
```

should just work. There's no problem with the current style though. I just wanted to show another option in case you're going to apply this to all other plugins.

HakkyuKim: Actually those look good, I will apply similar fixes to other plugins. Thanks!

PR #67, opened 2021-04-14
[camera] Implement setFocusMode

bbrto21:
* Return result as a boolean type in camera device API wrapper function
* Use the capture file format explicitly as JPEG

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -73,9 +73,37 @@ static ExifTagOrientation ChooseExifTagOrientatoin(
   return orientation;
 }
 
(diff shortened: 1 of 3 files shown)


packages/camera/tizen/src/camera_device.cc
```diff
@@ -188,36 +222,39 @@ void CameraDevice::Dispose() {
   }
 }
 
-int CameraDevice::GetDeviceCount() {
-  int count = 0;
+bool CameraDevice::GetDeviceCount(int &count) {
```
bwikbs, line 225: It's a minor stuff.. but... Why is it changed like this? Do you plan to use the return value later?

bwikbs (approved):

PR #68, opened 2021-04-14
[google_maps] Update HandleMethodCall

seungsoo47:
* Implement map#getVisibleRegion()
* Add LatLng & Convert classes

### DIFF
--- a/packages/google_maps_flutter/tizen/project_def.prop
+++ b/packages/google_maps_flutter/tizen/project_def.prop
@@ -6,7 +6,7 @@ type = sharedLib
 profile = common-5.5
 
 # Source files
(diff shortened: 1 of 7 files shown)


bbrto21:
I think that we should seriously review a few things before proceeding further work.

First, do we really need a native implementation for map view? 
IMO, This will be necessary if the platform's map view is supported, but may not be necessary if you are using the flutter web view.
 
Second, is it possible to implement the MapView functions that we originally intended to support with via the flutter web view interface? If it is possible, we can more easily develop this plugin on Dart world

I am sorry that this review will have a great impact on future works, but I politely request to review my comment

seungsoo47:
As we discussed offline, I agree on your opinions. 
Even for future implementation of chrome-based webview, I think that it would be better to rethink current structure of the plugin. Thank you your review.

PR #70, opened 2021-04-16
[camera] Implement setExposureMode

bbrto21:
* Clean-up codes
* Fix a crash when Disposing camera
* ExposureMode.locked is not supported on TM1

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -73,19 +73,61 @@ static ExifTagOrientation ChooseExifTagOrientatoin(
   return orientation;
 }
 
+bool StringToCameraPixelFormat(std::string Image_format,
+                               CameraPixelFormat &pixel_format) {
+  if (Image_format == "bgra8888") {
+    LOG_WARN("bgra8888 is unsupported");
+    return false;
+  } else if (Image_format == "yuv420") {
+    pixel_format = CameraPixelFormat::kYUV420;
+    return true;
+  } else if (Image_format == "jpeg") {
+    pixel_format = CameraPixelFormat::kJPEG;
+    return true;
+  }
+  LOG_WARN("Unknown Image format!");
+  return false;
+}
+
+bool StringToExposureMode(std::string mode, ExposureMode &exposure_mode) {
+  LOG_DEBUG("mode[%s]", mode.c_str());
+  if (mode == "auto") {
+    exposure_mode = ExposureMode::kAuto;
+    return true;
+  } else if (mode == "locked") {
+    exposure_mode = ExposureMode::kLocked;
+    return true;
+  }
+  LOG_WARN("Unknown focus mode!");
+  return false;
+}
+
 bool StringToFocusMode(std::string mode, FocusMode &focus_mode) {
   LOG_DEBUG("mode[%s]", mode.c_str());
   if (mode == "auto") {
-    LOG_DEBUG("FocusMode::kAuto");
     focus_mode = FocusMode::kAuto;
     return true;
   } else if (mode == "locked") {
-    LOG_DEBUG("FocusMode::kLocked");
     focus_mode = FocusMode::kLocked;
     return true;
   }
   LOG_WARN("Unknown focus mode!");
   return false;
+}
+
+bool ExposureModeToString(ExposureMode exposure_mode, std::string &mode) {
+  switch (exposure_mode) {
+    case ExposureMode::kAuto:
+      mode = "auto";
+      return true;
+    case ExposureMode::kLocked:
+      mode = "locked";
+      return true;
+    default:
+      LOG_WARN("Unknown focus mode!");
+      return false;
+      break;
+  }
 }
 
 bool FocusModeToString(FocusMode focus_mode, std::string &mode) {
@@ -131,7 +173,7 @@ flutter::EncodableValue CameraDevice::GetAvailableCameras() {
 }
 
 CameraDevice::CameraDevice() {
-  CreateCameraHandle();
+  CameraCreate();
   GetState(state_);
 }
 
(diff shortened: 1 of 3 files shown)


packages/camera/tizen/src/camera_device.cc
```diff
@@ -73,21 +73,63 @@ static ExifTagOrientation ChooseExifTagOrientatoin(
   return orientation;
 }
 
+bool StringToCameraPixelFormat(std::string Image_format,
+                               CameraPixelFormat &pixel_format) {
+  if (Image_format == "bgra8888") {
+    LOG_WARN("bgra8888 is unsupported");
+    return false;
+  } else if (Image_format == "yuv420") {
+    pixel_format = CameraPixelFormat::kYUV420;
```
bwikbs, line 82: Are these types supported in preview?

bbrto21: Yes, ```camera_set_preview_format``` can an accept yuv420.
but I've checked result of ```camera_foreach_supported_preview_format```, TM1 only supports CAMERA_PIXEL_FORMAT_NV12.

bbrto21: I don't have a deep understanding of the pixel format, but this link seems useful. https://www.programmersought.com/article/47255462169/

bwikbs: Well.. what I was curious about this...I maen my intention is ... when we use this api, the tbm surface (for preview) format is not affected by this changes.

bwikbs: As I remember, the format of the external texture is fixed... so..

bwikbs: If it works properly, you don't have to worry about this. Never mind

bbrto21: OK :)

bbrto21: > Well.. what I was curious about this...I maen my intention is ... when we use this api, the tbm surface (for preview) format is not affected by this changes.

For further investigation, I've contacted a guy who is in charge of camera API(Unfortunately, the API docs were insufficient for me.)
He told to me that both callbacks(registered through ```camera_set_media_packet_preview_cb``` and ```camera_set_preview_cb respectively``` ) will be affected  by the pixel format specified through ```camera_set_preview_format```.

For example, if I set the pixel format to CAMERA_PIXEL_FORMAT_JPEG, both callbacks will receive jpeg data.

bbrto21:
@bwikbs @seungsoo47 
Please review and approve this PR or leave some comments

bwikbs (approved):

packages/camera/tizen/src/camera_device.cc
bwikbs: Hmm.. Then I think there will be a problem. Why does it works? That's weird...

bbrto21: On TM1, CAMERA_PIXEL_FORMAT_NV12 is only supported, any other format specified is handled as an error.

bbrto21: > Hmm.. Then I think there will be a problem. Why does it works? That's weird...

Please describe your concerns as detailed as possible and I will investigate as much as possible even after this PR is merged.

bwikbs: Only support ``CAMERA_PIXEL_FORMAT_NV12``?
I'm little confusing.. if we set CAMERA_PIXEL_FORMAT_jpeg, the surface of preview is not jpeg?
I mean tbm surface at media_packet.

bbrto21: > Only support CAMERA_PIXEL_FORMAT_NV12?

On TM1, yes. ```CAMERA_PIXEL_FORMAT_NV12```
If you try to set a format other than ```CAMERA_PIXEL_FORMAT_NV12``` as the preview format using ```camera_set_preview_resolution```, an error occurs.

As already mentioned, according to the results of ```camera_foreach_supported_preview_format```, only ```NV12``` is supported on TM1.

bwikbs: Here is our external texture code which is used at preview
```c++
  PFNGLEGLIMAGETARGETTEXTURE2DOESPROC glEGLImageTargetTexture2DOES =
      (PFNGLEGLIMAGETARGETTEXTURE2DOESPROC)eglGetProcAddress(
          "glEGLImageTargetTexture2DOES");
  glEGLImageTargetTexture2DOES(GL_TEXTURE_EXTERNAL_OES, eglSrcImage);
  if (eglSrcImage) {
    PFNEGLDESTROYIMAGEKHRPROC n_eglDestoryImageKHR =
        (PFNEGLDESTROYIMAGEKHRPROC)eglGetProcAddress("eglDestroyImageKHR");
    n_eglDestoryImageKHR(eglGetCurrentDisplay(), eglSrcImage);
  }
  opengl_texture->target = GL_TEXTURE_EXTERNAL_OES;
  opengl_texture->name = state_->gl_texture;
  opengl_texture->format = GL_RGBA8;
  opengl_texture->destruction_callback = (VoidCallback)DestructionCallback;
  opengl_texture->user_data = static_cast<void*>(this);
  opengl_texture->width = width;
  opengl_texture->height = height;
  mutex_.unlock();
  return true;
```
We can see it is fixed as ``GL_RGBA8``. That's why I'm thinking that someone should convert it...
It's already working, so it doesn't matter much, but... I'm just curious.

bbrto21: >  opengl_texture->format = GL_RGBA8;

This could be a problem. before getting the gl texture, we are already getting tbm_surface_info_s. I think we should check the format using this. here is related docs 
https://docs.tizen.org/application/native/api/mobile/4.0/struct__tbm__surface__info.html
https://docs.tizen.org/application/native/api/mobile/4.0/group__CAPI__UI__TBM__SURFACE__MODULE.html#ga13b29098467d27c545c78d35f141477d

PR #73, opened 2021-04-21
[camera] Send focusPointSupported as a true

bbrto21:
* Even if I send this to false when opening camera,
  the focus point request arrived to method handler normally


seungsoo47 (approved):
LGTM

PR #75, opened 2021-04-22
[camera] Implement setZoomLevel

bbrto21:
* Implement get{Max|min}ZoomLevel

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -8,6 +8,8 @@
 #include <flutter/encodable_value.h>
 #include <sys/time.h>
 
+#include <cmath>
+
 #include "log.h"
 
 static uint64_t Timestamp() {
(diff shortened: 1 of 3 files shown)


packages/camera/tizen/src/camera_device.cc
```diff
@@ -728,8 +761,8 @@ bool CameraDevice::SetPreviewFormat(CameraPixelFormat format) {
 
 bool CameraDevice::SetPreviewSize(Size size) {
   int w, h;
-  w = static_cast<int>(size.width);
-  h = static_cast<int>(size.height);
+  w = static_cast<int>(round(size.width));
```
bwikbs, line 764: Why do we need this?

bbrto21: The type of ```Size```'s memeber are double, I thought it was better than throwing away the decimal point.

bwikbs (approved):

PR #77, opened 2021-04-27
[camera] Request a recorder permission

bbrto21:
**This PR includes :**
* [camera] Use 'Camera' as a wrapper function name prefix
* [camera] Request permissions as async callchain

**Note**
This implementation uses an asynchronous request chain for multiple permissions, 
because multiple privileges in single API call are supported from Tizen 5.0


seungsoo47 (approved):
LGTM

PR #79, opened 2021-04-29
[camera] Implement video recording

bbrto21:
* Implement {start|stop|pause|resume}Recording
* Cleanup code

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -12,6 +12,10 @@
 
 #include "log.h"
 
+// These macros came from tizen camera_app
+#define VIDEO_ENCODE_BITRATE 40000000 /* bps */
+#define AUDIO_SOURCE_SAMPLERATE_AAC 44100
+
 static uint64_t Timestamp() {
   struct timeval tv;
   gettimeofday(&tv, NULL);
@@ -73,6 +77,28 @@ static ExifTagOrientation ChooseExifTagOrientatoin(
   }
 
   return orientation;
+}
+
+static RecorderOrientationTag ChooseRecorderOrientationTag(
+    OrientationType device_orientation) {
+  RecorderOrientationTag tag = RecorderOrientationTag::kNone;
+  switch (device_orientation) {
+    case OrientationType::kPortraitUp:
+      tag = RecorderOrientationTag::k90;
+      break;
+    case OrientationType::kLandscapeLeft:
+      tag = RecorderOrientationTag::k180;
+      break;
+    case OrientationType::kPortraitDown:
+      tag = RecorderOrientationTag::k270;
+      break;
+    case OrientationType::kLandscapeRight:
+      tag = RecorderOrientationTag::kNone;
+    default:
+      LOG_ERROR("Unknown RecorderOrientationTag!");
+      break;
+  }
+  return tag;
 }
 
 bool StringToCameraPixelFormat(std::string Image_format,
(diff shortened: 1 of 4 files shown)


packages/camera/tizen/src/log.h
```diff
@@ -8,9 +8,14 @@
 #endif
 #define LOG_TAG "CameraPlugin"
 
-#define LOG(prio, fmt, arg...)                                                 \
-  dlog_print(prio, LOG_TAG, "%s: %s(%d) > " fmt, __FILE__, __func__, __LINE__, \
-             ##arg)
+#ifndef __MODULE__
```
bbrto21, line 11: @swift-kim if you still have a plan what keep a log template for plugin, I think it would be nice to add ```__MODULE__``` macro

swift-kim: Ah, yes, I'm planning not to remove the log.h file from the plugin template and instead just wrap the plugin class with an unnmaed namespace:

https://github.com/swift-kim/flutter-tizen/commit/ac78aacac77d53f6f821daa5fb7fe2ff64ce3d55

Isn't `__MODULE__` defined by default?

bbrto21: Well, I thought there was, but there was no.

packages/camera/tizen/src/camera_device.cc
```diff
@@ -614,14 +795,31 @@ bool CameraDevice::Open(std::string image_format_group) {
...
   auto value = std::make_unique<flutter::EncodableValue>(map);
   camera_method_channel_->Send(CameraEventType::kInitialized, std::move(value));
+
   return true;
 }
 
+void CameraDevice::PauseVideoRecording(
+    std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> &&result) {
+  LOG_DEBUG("enter");
+  PauseRecorder();
```
seungsoo47, line 805: The PauseRecorder()'s return value is a boolean type. If it returns false, should not this function also return Error()?

packages/camera/tizen/src/camera_device.h
```diff
@@ -159,23 +224,45 @@ class CameraDevice {
...
+  bool CommitRecorder();
+  bool DestroyRecorder();
+  bool GetRecorderState(RecorderState &state);
+  bool GetRecorderFileName(std::string &name);
+  bool SetRecorderAudioChannel(RecorderAudioChannel chennel);
+  bool SetRecorderAudioDevice(RecorderAudioDevice device);
+  bool SetRecorderAudioEncorder(RecorderAudioCodec codec);
+  bool SetRecorderAudioSamplerate(int samplerate);
+  bool SetRecorderFileFormat(RecorderFileFormat format);
+  bool SetRecorderFileName(std::string name);
```
seungsoo47, line 252: How about changing std::string to const std::string& ?

packages/camera/tizen/src/camera_device.cc
bbrto21: You're right, I designed the return type of all wrapping methods as Boolean, but all public APIs in CameraDevice doesn't thrown any errors and I rarely checked the return value yet.
I have a plan to mimic the error handle behavior on Android after I finish implementing features roughly. 
I'd like to implement these things later If you don't mind.

packages/camera/tizen/src/camera_device.h
bbrto21: Oh! thank you

seungsoo47 (approved):

PR #81, opened 2021-04-30
[camera] Implement lockCaptureOrientation

bbrto21:

### DIFF
--- a/packages/camera/tizen/src/camera_plugin.cc
+++ b/packages/camera/tizen/src/camera_plugin.cc
@@ -175,14 +175,19 @@ class CameraPlugin : public flutter::Plugin {
       }
       result->Error("InvalidArguments", "Please check 'mode'");
     } else if (method_name == "setExposurePoint") {
+      LOG_WARN("setExposurePoint is not supported!");
       result->NotImplemented();
     } else if (method_name == "getMinExposureOffset") {
+      LOG_WARN("getMinExposureOffset is not supported!");
       result->NotImplemented();
     } else if (method_name == "getMaxExposureOffset") {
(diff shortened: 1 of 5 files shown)


packages/camera/tizen/src/camera_plugin.cc
```diff
@@ -241,9 +246,22 @@ class CameraPlugin : public flutter::Plugin {
...
+        std::string orientation;
+        OrientationType orientation_type;
+        if (GetValueFromEncodableMap(arguments, "orientation", orientation) &&
+            StringToOrientationType(orientation, orientation_type)) {
+          camera_->LockCaptureOrientation(orientation_type);
+          result->Success();
+          return;
+        }
+        result->Error("InvalidArguments", "Please check 'orientation'");
+      }
```
bwikbs, line 261: How about putting some logs here?  I mean for error case handling

bwikbs: It's just a suggestion... I guess you know the reason :imp:

bwikbs:
Thank you for your dedicated work! It's pretty awesome :+1:

packages/camera/tizen/src/camera_plugin.cc
bbrto21: Well... I want to reduce unnecessary logs as much as possible. in this case, throwing an error directly to ```flutter``` and leaving enough logs in the wrapper implementation. so I think it is enough

bwikbs: Got it.

bwikbs (approved):
LGTM

PR #88, opened 2021-05-07
[camera] Use enable_audio argument

bbrto21:
* If enable_audio is false, disable audio track when recording video


seungsoo47 (approved):

bwikbs (approved):
Thanks for your hard working :+1:

PR #91, opened 2021-05-07
[google_maps] Initialize with google map options

seungsoo47:
* Create a map through assets/map.html
* Parse initialization options
* Support only web-based options


bwikbs (approved):
Thanks for your hard working!

PR #96, opened 2021-05-13
[webview_flutter] Migrate example app host to latest

bwikbs:


swift-kim:
Regarding the migration from `PackageReference` to `ProjectReference`, I think it would be better to update all our plugins at once.

bwikbs:
@swift-kim got it. I think it would be better.

PR #102, opened 2021-05-17
[google_maps] Register listeners to google map

seungsoo47:
* Attach bounds_changed, idle, click, rightclick listeners to a google map's stream controller
* Add getZoomLevel API

### DIFF
--- a/packages/google_maps_flutter/lib/src/google_maps_controller.dart
+++ b/packages/google_maps_flutter/lib/src/google_maps_controller.dart
@@ -171,12 +171,17 @@ class GoogleMapController {
     return JavascriptChannel(
         name: 'Click',
         onMessageReceived: (JavascriptMessage message) async {
-          final Map<String, dynamic> event =
-              json.decode(message.message) as Map<String, dynamic>;
-          assert(event['latLng'] != null);
-          final LatLng position = LatLng(event['latLng']['lat'] as double,
-              event['latLng']['lng'] as double);
-          _streamController.add(MapTapEvent(_mapId, position));
+          try {
+            final dynamic event = json.decode(message.message);
+            if (event is Map<String, dynamic>) {
+              assert(event['latLng'] != null);
+              final LatLng position = LatLng(event['latLng']['lat'] as double,
+                  event['latLng']['lng'] as double);
+              _streamController.add(MapTapEvent(_mapId, position));
+            }
+          } catch (e) {
+            print('Javascript Error: $e');
+          }
         });
   }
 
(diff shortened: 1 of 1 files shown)


packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -27,66 +27,65 @@ class GoogleMapController {
...
-          print('Google map started loading.');
-        },
-        onPageFinished: (String url) async {
-          print('Google map finished loading.');
-        },
-        gestureNavigationEnabled: true,
-      );
-    }
-    return _widget;
+    String html = await rootBundle.loadString('assets/map.html');
```
bbrto21, line 42: Currently, map.html is owned by example. IMO, I think that the plugin implementation can have it directly as a string. It seems that you will have to think about how to get the API key from the outside of this plugin

packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -143,6 +142,58 @@ class GoogleMapController {
...
+        });
+  }
+
+  JavascriptChannel _onRightClick() {
+    // NOTE: LWE does not support a long press event.
+    return JavascriptChannel(
+        name: 'RightClick',
+        onMessageReceived: (JavascriptMessage message) async {
+          final Map<String, dynamic> event =
+              json.decode(message.message) as Map<String, dynamic>;
```
bbrto21, line 189: Can you always be sure of the format of the message?
if you didn't, how about this. I haven't checked whether the below works, but please refer to the meaning only.
```suggestion
              dynamic temp =  json.decode(message.message);
              if (temp is Map<String, dynamic> ) { 
                  event =  temp as Map<String, dynamic>;
              } 
```

packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -332,17 +357,48 @@ class GoogleMapController {
     }
   }
 
+  LatLng _convertToLatLng(String value) {
+    try {
```
bbrto21, line 361: or you can use try where you try to cast a type

packages/google_maps_flutter/lib/src/google_maps_controller.dart
seungsoo47: Yes. Even though I have no any good idea yet, we have to consider to solve this issue.

packages/google_maps_flutter/lib/src/google_maps_controller.dart
seungsoo47: It's a so good idea. However, I think it will be always returned into JSON style if it works properly.

packages/google_maps_flutter/lib/src/google_maps_controller.dart
seungsoo47: Done.

packages/google_maps_flutter/lib/src/google_maps_controller.dart
bbrto21: All right, let's find a way to solve it later.

bbrto21 (approved):
LGTM

PR #103, opened 2021-05-18
[google_maps] Set traffic layer to map

seungsoo47:
* Attach/Detach a traffic layer to google map
* Fix a parsing bug regarding a getCameraBound option


bwikbs (approved):
Thanks for your hard working! :+1:

PR #111, opened 2021-05-25
Use get_error_message() whenever possible

swift-kim:
- Avoid the error converter function and use `get_error_message()` instead.
- Use the new log.h template (`__FILE__` to `__MODULE__`).

### DIFF
--- a/packages/audioplayers/tizen/src/log.h
+++ b/packages/audioplayers/tizen/src/log.h
@@ -9,8 +9,7 @@
 #define LOG_TAG "AudioplayersTizenPlugin"
 
 #ifndef __MODULE__
-#define __MODULE__ \
-  (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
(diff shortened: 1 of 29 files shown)


bbrto21 (approved):
LGTM!  I feel comfortable after review this :smile:

rwalczyna (approved):
To get rid off condition when extracting file name, you can prepend `/` to `__FILE__`:

```cpp
#define __MODULE__ strrchr("/" __FILE__, '/') + 1
```
But it is not an issue and overall PR looks good to me.

swift-kim:
@bbrto21 How do you think about @rwalczyna's suggestion? The current way is a bit more readable, but the suggested way looks simpler and cleaner.

bbrto21:
I think @rwalczyna's suggestion looks better!
@swift-kim If you don't mind, could you update all macros of this PR and template of log.h ?

bwikbs (approved):
I didn't know that something like this ``get_error_message`` exist. 
I have been looking for a while where ``get_error_message`` is in your PR. :smile:

PR #112, opened 2021-05-25
[messageport] Add messageport plugin code

rwalczyna:
This plugin enables communication between two Flutter applications on Tizen platform.
It is a part of flutter-tizen/flutter-tizen#94

Signed-off-by: Rafal Walczyna <r.walczyna@samsung.com>


packages/messageport/tizen/src/messageport.cc
```diff
@@ -0,0 +1,224 @@
...
+    std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> sink,
+    bool is_trusted, int* local_port) {
+  LOG_DEBUG("RegisterLocalPort: %s, is_trusted: %s", port_name.c_str(),
+            is_trusted ? "yes" : "no");
+  int ret = -1;
+  if (is_trusted) {
+    ret = message_port_register_trusted_local_port(port_name.c_str(),
+                                                   OnMessageReceived, this);
+  } else {
+    ret = message_port_register_trusted_local_port(port_name.c_str(),
```
pkosko, line 112:
```suggestion
    ret = message_port_register_local_port(port_name.c_str(),
```

packages/messageport/tizen/src/messageport.cc
```diff
@@ -0,0 +1,224 @@
...
+  bundle* b = bundle_create();
+
+  bool result = ConvertEncodableValueToBundle(message, b);
+  if (!result) {
+    LOG_ERROR("Failed to parse EncodableValue");
+    bundle_free(b);
+    return CreateResult(MESSAGE_PORT_ERROR_INVALID_PARAMETER);
+  }
+
+  int ret = message_port_send_trusted_message(remote_app_id.c_str(),
```
pkosko, line 189: always send trusted? what about non-trusted ports?

packages/messageport/tizen/src/messageport.cc
```diff
@@ -0,0 +1,224 @@
...
+  bundle* b = bundle_create();
+
+  bool result = ConvertEncodableValueToBundle(message, b);
+  if (!result) {
+    LOG_ERROR("Failed to parse EncodableValue");
+    bundle_free(b);
+    return CreateResult(MESSAGE_PORT_ERROR_INVALID_PARAMETER);
+  }
+
+  int ret = message_port_send_trusted_message_with_local_port(
```
pkosko, line 210: always send trusted?

packages/messageport/tizen/src/messageport.cc
```diff
@@ -0,0 +1,224 @@
...
+            *port_check ? "true" : "false");
+
+  return CreateResult(ret);
+}
+
+MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
+                                           std::string& port_name,
+                                           flutter::EncodableValue& message) {
+  LOG_DEBUG("Send (%s, %s)", remote_app_id.c_str(), port_name.c_str());
+  bundle* b = bundle_create();
```
pkosko, line 180: bundle_create can return nullptr, we should check it

packages/messageport/tizen/src/messageport.cc
```diff
@@ -0,0 +1,224 @@
...
+  return CreateResult(ret);
+}
+
+MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
+                                           std::string& port_name,
+                                           flutter::EncodableValue& message,
+                                           int local_port) {
+  LOG_DEBUG("Send (%s, %s), port: %d", remote_app_id.c_str(), port_name.c_str(),
+            local_port);
+  bundle* b = bundle_create();
```
pkosko, line 201: bundle_create can return nullptr, we should check it

pkosko (changes requested):

packages/messageport/example/.metadata
```diff
@@ -0,0 +1,10 @@
+# This file tracks properties of this Flutter project.
```
swift-kim, line 1: Please remove this file.

packages/messageport/lib/messageport_tizen.dart
```diff
@@ -0,0 +1,163 @@
...
+    return _isTrusted;
+  }
+
+  final String _remoteAppId;
+  final String _portName;
+  final bool _isTrusted;
+}
+
+/// API for accessing MessagePorts in Tizen
+class TizenMessageport {
```
swift-kim, line 129: How do you think of removing the `Tizen` prefix from all class names? I think it's redundant.

packages/messageport/lib/messageport_tizen.dart
```diff
@@ -0,0 +1,163 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+
+final MethodChannel _channel = MethodChannel('tizen/messageport');
```
swift-kim, line 9: Please resolve all linter warnings. You can check them either in the IDE or using the `flutter analyze` command.

image

image

packages/messageport/lib/messageport_tizen.dart
rwalczyna: I can do it but `MessagePort` is already a part of [dart:html](https://api.dart.dev/stable/2.13.1/dart-html/MessagePort-class.html). If it's OK to have `Messageport` (from this plugin) and `MessagePort` (from html), then I can change the name. Or should I change it `MessagePort`?

packages/messageport/example/lib/main.dart
```diff
@@ -0,0 +1,239 @@
+import 'package:flutter/material.dart';
+import 'dart:async';
```
swift-kim, line 2: You can also apply the linter rules to the example files by commenting out the line
```yaml
- "**/example/**"
```
in the `analysis_options.yaml` file.

I excluded them only because we usually copy examples from the "original" package and we don't care how well they are written. Since this package is written by you from scratch, you can temporarily include the example files for analysis.

packages/messageport/lib/messageport_tizen.dart
swift-kim: Oops. In that case, it'd be better to use the name `TizenMessagePort` (capitalized P) to avoid confusion, but you may still consider removing `Tizen` from `TizenLocalPort` and `TizenRemotePort`.

rwalczyna:
Thanks for the reviews, I have updated the PR. Please check whether I correctly fixed all the issues :)

bbrto21:
Please click ```Resolve Conversation``` if  you done

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -0,0 +1,236 @@
...
+
+  virtual ~MessageportTizenPlugin() {}
+
+ private:
+  template <typename T>
+  bool GetValueFromArgs(const flutter::EncodableValue *args, const char *key,
+                        T &out) {
+    if (std::holds_alternative<flutter::EncodableMap>(*args)) {
+      flutter::EncodableMap map = std::get<flutter::EncodableMap>(*args);
+      flutter::EncodableValue value = map[flutter::EncodableValue(key)];
```
bbrto21, line 53: Unintended insertion occurs If the map doesn't have a key.
Please use a find like line 68

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    if (native_result) {
+      result->Success();
+    } else {
+      result->Error("Could not send message", native_result.message());
+    }
+  }
+
+  int nextId_ = 0;
```
bbrto21, line 221:
```suggestion
  int next_id_ = 0;
```

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -0,0 +1,236 @@
...
+    std::string port_name = "";
+    bool is_trusted = false;
+    if (!GetValueFromArgs<std::string>(args, "portName", port_name) ||
+        !GetValueFromArgs<bool>(args, "isTrusted", is_trusted)) {
+      result->Error("Could not create local port", "Invalid parameter");
+      return;
+    }
+
+    std::stringstream event_channel_name;
+    int id = nextId_++;
```
bbrto21, line 130: ```nextId_``` will be negative someday and you use -1 as initial value at line, it seems to intend invalid state.
I don't think it's gonna be a problem right now, but I don't think you intend to allow negative id

bbrto21 (commented):
In the example app, enable attach local port, send a message, then error log always occurs.
```bash
E/MessageportTizenPlugin (P24993, T24993): messageport_tizen_plugin.cc: GetValueFromArgs(58) > Key localPort not found

```

packages/messageport/tizen/src/messageport_tizen_plugin.cc
rwalczyna: `id` from method `Send` will be only used when there is a proper value sent from Dart layer. If next_id_ became -1 (after creating 2^32 ports in one run of application) then nothing wrong should happen.

bbrto21 (approved):
LGTM

packages/messageport/lib/messageport_tizen.dart
```diff
@@ -0,0 +1,164 @@
...
+  final bool _isTrusted;
+  final String _portName;
+  StreamSubscription<dynamic>? _streamSubscription;
+  late EventChannel _eventChannel;
+
+  static final Set<String> _registeredPorts = <String>{};
+}
+
+/// Remote port to send messages
+class RemotePort {
```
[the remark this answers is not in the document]
rwalczyna: You are right, I will add check method. In native guide there is suggestion to check for remote port before connecting and as I merged this into connect method, I forgot that user may want to check if the port is still OK 😶

packages/messageport/example/tizen/tizen-manifest.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="com.example.messageport_tizen_example" version="1.0.0" api-version="5.5" xmlns="http://tizen.org/ns/packages">
+    <profile name="common"/>
+    <ui-application appid="com.example.messageport_tizen_example" exec="runner" type="capp" multiple="false" nodisplay="false" taskmanage="true" hw-acceleration="on">
```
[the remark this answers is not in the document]
rwalczyna: When cpp plugin is being created, then example app is also cpp. I will change it to dotnet :)
Bundle API that I used for reading/writing binary data requires Tizen 5.5. But I found another way for doing the same, so I can go back to 4.0 version.

packages/messageport/lib/messageport_tizen.dart
```diff
@@ -0,0 +1,164 @@
...
+class LocalPort {
+  LocalPort._(this._id, this._portName, this._isTrusted);
+
+  /// Register messageport and sets listener
+  ///
+  /// Names of already registered ports has to be unique.
+  /// Exception will be thrown when using on already registered port
+  void register(OnMessageReceived onMessage) {
+    if (_registeredPorts.contains(portName)) {
+      throw Exception('Port $portName is already registered');
```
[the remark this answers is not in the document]
rwalczyna: I made a small change which enables creating independent LocalPort object with the same name. In case of registering both (or more) local ports, all ports will receive the same message.

packages/messageport/lib/messageport_tizen.dart
```diff
@@ -0,0 +1,164 @@
...
+    return _registeredPorts.contains(_portName);
+  }
+
+  final int _id;
+  final bool _isTrusted;
+  final String _portName;
+  StreamSubscription<dynamic>? _streamSubscription;
+  late EventChannel _eventChannel;
+
+  static final Set<String> _registeredPorts = <String>{};
```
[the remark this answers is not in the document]
rwalczyna: I have changed the implementation and deleted `_registeredPorts` set. All values sent from dart are validated by native side, so nothing wrong should happen in case of invalid parameters - exception will be thrown.
Also there is nothing wrong with sharing the same port name by two or more applications - ports are defined by its name and Application Id.

packages/messageport/tizen/src/messageport.cc
```diff
@@ -175,43 +164,66 @@ MessagePortResult MessagePortManager::CheckRemotePort(
...
 MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
                                            std::string& port_name,
-                                           flutter::EncodableValue& message) {
+                                           flutter::EncodableValue& message,
+                                           bool is_trusted) {
   LOG_DEBUG("Send (%s, %s)", remote_app_id.c_str(), port_name.c_str());
-  bundle* b = bundle_create();
-
-  bool result = ConvertEncodableValueToBundle(message, b);
+  bundle* b;
```
pkosko, line 170:
```suggestion
  bundle* b = nullptr;
```

packages/messageport/tizen/src/messageport.cc
```diff
@@ -175,43 +164,66 @@ MessagePortResult MessagePortManager::CheckRemotePort(
 
 MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
                                            std::string& port_name,
-                                           flutter::EncodableValue& message) {
+                                           flutter::EncodableValue& message,
+                                           bool is_trusted) {
```
pkosko, line 168: maybe log is_trusted also in below log?

packages/messageport/tizen/src/messageport.cc
```diff
@@ -175,43 +164,66 @@ MessagePortResult MessagePortManager::CheckRemotePort(
...
   bundle_free(b);
   return CreateResult(ret);
 }
 
 MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
                                            std::string& port_name,
                                            flutter::EncodableValue& message,
-                                           int local_port) {
+                                           bool is_trusted, int local_port) {
   LOG_DEBUG("Send (%s, %s), port: %d", remote_app_id.c_str(), port_name.c_str(),
```
pkosko, line 192: maybe log is_trusted?

packages/messageport/tizen/src/messageport.cc
```diff
@@ -175,43 +164,66 @@ MessagePortResult MessagePortManager::CheckRemotePort(
...
             local_port);
-  bundle* b = bundle_create();
+  bundle* b;
```
pkosko, line 194:
```suggestion
  bundle* b = nullptr;
```

packages/messageport/example/tizen/tizen-manifest.xml
```diff
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="com.example.messageport_tizen_example" version="1.0.0" api-version="5.5" xmlns="http://tizen.org/ns/packages">
+<manifest package="com.example.dotnet" version="1.0.0" api-version="4.0" xmlns="http://tizen.org/ns/packages">
```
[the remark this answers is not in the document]
rwalczyna: The name will be changed to org.tizen.messageport_tizen_example (as flutter-tizen require tizen.org domain [here](https://github.sec.samsung.net/f-project/plugins/wiki/Checklist)).

swift-kim (approved):
It worked as expected. Thank you!

pkosko (approved):
LGTM

PR #113, opened 2021-05-26
[WebView] Added implementation of rotation of webview

pkosko:
Replaced a "// NOTE: Not supported by LWE on Tizen." with a working implementation

Before: rotation of application was causing a problem with rescalling
web view content.

After: webview is being recreated properly and looks good after rotation
event.

Observed issue: because of recretation of WebView object, browsing
history managed by webView is erased after rotation of application.
Probably it is not crucial.

Signed-off-by: Piotr Kosko/Tizen API (PLT) /SRPOL/Engineer/Samsung Electronics <p.kosko@samsung.com>

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -168,7 +168,81 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
       context_(nullptr) {
   SetTextureId(FlutterRegisterExternalTexture(texture_registrar_));
   InitWebView();
+  InitFlutterChannels();
 
+  auto initial_url = params[flutter::EncodableValue("initialUrl")];
+  if (std::holds_alternative<std::string>(initial_url)) {
+    webview_url_ = std::get<std::string>(initial_url);
+  } else {
+    webview_url_ = "about:blank";
+  }
+
+  auto settings = params[flutter::EncodableValue("settings")];
+  if (std::holds_alternative<flutter::EncodableMap>(settings)) {
+    webview_settings_ = std::get<flutter::EncodableMap>(settings);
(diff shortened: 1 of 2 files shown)


pkosko:
@seungsoo47 I've added an extension of plugin you created. Could you please take a look to modifications? Especially please take a look at [line 768](https://github.com/flutter-tizen/plugins/pull/113/files#diff-224116c6c9f92768b00c294ffe0a45214f07822a0ae11052b85c989306c7fa63L768) if my change is ok or such design was on purpose.

seungsoo47:
@pkosko Thank you for your effort. As you know, the rotation function is related to resize function, but LWE's resize function is not support now. 
If it is implemented, webview plugin may be also run without re-initialization. 
And I think that the history has to be remained as long as user does not close the app. 
@bwikbs @bbrto21 How do you think of it?

bwikbs:
I'm not sure because I didn't look closely, but... If what we're discussing now is that the user context needs to be maintained after rotation... I totally agree with @seungsoo47 's opinion.

bwikbs:
Now, I understand what's wrong with the rotation. @pkosko is this urgent? (I think we should do this after lwe's resize is implemented properly) If so, I'll implement it first.

bwikbs:
@pkosko why did you close this PR? 👀

pkosko:
@bwikbs in context of your comments, this implementation is invalid, thus I closed PR. Would you like me to reopen this PR and just leave it waiting for support on LWE layer and then to fix my implementation to use proper LWE interface?

bwikbs:
@pkosko The reason I asked was because I was curious about the reason. You can do it the way you said, or you can do it the other way anything. This is your PR, so it's up to you!  (If you need anything, just let me know. 😄

PR #119, opened 2021-06-01
[camera] Apply new external texture APIs

bbrto21:

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -253,12 +253,10 @@ CameraDevice::CameraDevice() {
 }
 
 CameraDevice::CameraDevice(flutter::PluginRegistrar *registrar,
-                           FlutterTextureRegistrar *texture_registrar,
                            CameraDeviceType type,
                            ResolutionPreset resolution_preset,
                            bool enable_audio)
     : registrar_(registrar),
-      texture_registrar_(texture_registrar),
       type_(type),
       resolution_preset_(resolution_preset),
       enable_audio_(enable_audio) {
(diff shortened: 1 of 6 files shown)


packages/camera/tizen/src/camera_device.cc
```diff
@@ -326,7 +324,40 @@ CameraDevice::CameraDevice(flutter::PluginRegistrar *registrar,
   SetResolutionPreset(resolution_preset_);
 
   // Init channels
-  texture_id_ = FlutterRegisterExternalTexture(texture_registrar_);
+  // texture_id_ = FlutterRegisterExternalTexture(texture_registrar_);
```
bwikbs, line 327: :scissors:!

packages/camera/tizen/src/camera_device.cc
```diff
@@ -852,6 +882,27 @@ void CameraDevice::UpdateStates() {
   GetRecorderState(recorder_state_);
 }
 
+bool CameraDevice::RegistAvailableMediaPacket(media_packet_h packet) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (packet_) {
```
bwikbs, line 887: It's not much of a difference, but how about replacing it with a new packet?

packages/camera/tizen/src/camera_device.h
```diff
@@ -323,9 +321,16 @@ class CameraDevice {
   bool UnsetRecorderRecordingLimitReachedCb();
   void UpdateStates();
 
+  bool RegistAvailableMediaPacket(media_packet_h packet);
+  media_packet_h UnregistAvailableMediaPacket();
```
swift-kim, line 325: Regist → Register?

packages/camera/tizen/src/camera_device.cc
bbrto21: Thanks!

packages/camera/tizen/src/camera_device.cc
bbrto21: please review again, I updated this PR

packages/camera/tizen/src/camera_device.h
bbrto21: it is no longer a member, please review again

swift-kim (approved):

PR #121, opened 2021-06-02
Add github action for continuous builds

bbrto21:
* Build examples of changed packages when create pull request

### DIFF
--- a/.github/workflows/build_changed_packages.yml
+++ b/.github/workflows/build_changed_packages.yml
@@ -1,0 +1,93 @@
+name: Build changed packages
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: jitterbit/get-changed-files@v1
+        id: files
+        with:
+          format: space-delimited
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Find changed packages
+        run: |
+          function list_include_item {
+            local list="$1"
+            local item="$2"
+            if [[ $list =~ (^|[[:space:]])"$item"($|[[:space:]]) ]] ; then
+              result=0
+            else
+              result=1
+            fi
+            return $result
+          }
+          CHANGED_PACKAGES=()
+          for changed_file in ${{ steps.files.outputs.all }}; do
+            if [[ "$changed_file" == "packages"*".cc" ]] || [[ "$changed_file" == "packages"*".h" ]]; then
+                package=$(echo ${changed_file} | sed 's|\/tizen.*||;s|.*packages\/||')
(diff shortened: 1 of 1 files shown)


bwikbs:
OMG ! :scream: 
Now, I know what you're doing! 
I'll test it and let you know! Thank you! I love it! :heart:

.github/workflows/build_changed_packages.yml
```diff
@@ -0,0 +1,97 @@
...
+            fi
+          done
+          if [[ ${#CHANGED_PACKAGES[@]} -eq 0 ]]; then
+              echo "ENABLE=false" >> "$GITHUB_ENV"
+          else
+              echo "ENABLE=true" >> "$GITHUB_ENV"
+              echo "CHANGED_PACKAGES=${CHANGED_PACKAGES[@]}" >> "$GITHUB_ENV"
+          fi
+      - name: Install prerequisite packages
+        if: contains(env.ENABLE, true)
```
swift-kim, line 44: How about
```suggestion
        if: env.ENABLE == 'true'
```
or just
```suggestion
        if: env.CHANGED_PACKAGES != ''
```

swift-kim (commented):
It is quite unfortunate that we have to split the workflows (jobs) into two only because GitHub Actions doesn't support force exiting a job earlier.

How about merging the workflows into one (`build.yml`) and splitting into two jobs like:

```yaml
name: Build

on:
  push:
  pull_request:

jobs:
  push:
    if: github.event_name == 'push'
    ...
  pr:
    if: github.event_name == 'pull_request'
    ...
```

.github/workflows/build_all_packages.yml
```diff
@@ -0,0 +1,62 @@
...
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install prerequisite packages
+        run: |
+          sudo apt-get install \
```
swift-kim, line 15: How about

```suggestion
          sudo apt install -y \
```

just for consistency with the line

```
sudo apt install -y pciutils zip libncurses5 python libpython2.7
```

.github/workflows/build_changed_packages.yml
bbrto21: Thanks~ I choose 2nd

.github/workflows/build_changed_packages.yml
```diff
@@ -0,0 +1,94 @@
+name: Build changed packages
+
+on:
+  pull_request:
```
swift-kim, line 4: Not building on `push`?
```suggestion
  push:
  pull_request:
```

bbrto21: Thanks! That's the part I missed.

swift-kim (approved):

bwikbs (approved):

PR #122, opened 2021-06-02
[path_provider] Update README with correct information

swift-kim:


packages/path_provider/README.md
```diff
@@ -20,12 +20,18 @@ import 'package:path_provider/path_provider.dart';
 
 For detailed usage, see https://github.com/flutter/plugins/tree/master/packages/path_provider/path_provider#usage.
 
+## Notes
+
+- On Tizen, `getExternalStorageDirectories` will return internal storage paths (such as `/home/owner/media/Music`) unlike on Android where the function returns external storage (separate partition or SD card) paths.
+- To access paths returned by `getExternalStorageDirectory` and `getExternalCacheDirectories`, you will need an SD card inserted to your Tizen device.
+- This plugin will work correctly only if your code is running in the **main isolate**.
```
bbrto21, line 27: Does it work only in main isolation? Is there a way to use this plugin in other isolation?

swift-kim: In secondary isolates you have to manually register the plugin by calling either

```dart
PathProviderPlugin.register();
```
or
```dart
PathProviderPlatform.instance = PathProviderPlugin();
```

with some platform check like https://github.com/flutter-tizen/plugins/pull/54#discussion_r605455288 if targeting cross platforms.

We eventually have to remove this limitation but for now I'm not sure how. I need to look into the current status of FFI plugins on other platforms.

bbrto21: How about including those ways in ```README``` until we find a way to remove limitation? 
The last time I looked at similar things with interest, the other major plugins didn't seem to care too much.

swift-kim: I don't have time to investigate this thing right now. I'll just remove this line.

bbrto21 (approved):

PR #120, opened 2021-06-02
[battery_plus] Migrate battery code to battery_plus

HakkyuKim:
The only significant change is adding some dart code to set the correct platform interface for Tizen.
Please see `lib/battery_plus_tizen.dart` and `lib/method_channel_battery_plus_tizen.dart` for details.

Issue #116, opened 2021-05-30
Migrate to "plus" plugins


swift-kim:
Flutter community plus plugins: https://plus.fluttercommunity.dev

Google is officially deprecating their first-party plugins in favor of these plus packages.

https://medium.com/flutter/whats-new-in-flutter-2-2-fd00c65e2039#ca45

> You should migrate your code to the “plus” packages soon and, in the coming weeks, the corresponding packages from Google will be updated to recommend that you do so.

https://github.com/flutter/plugins/pull/3971

HakkyuKim:
- [ ] battery_plus
- [ ] connectivity_plus
- [ ] device_info_plus
- [ ] package_info_plus
- [ ] sensors_plus
- [ ] share_plus
- [ ] network_info_plus (corresponds to wifi_info_flutter)

bbrto21:
I'm working on sensors_plus

bbrto21:
All migrations are complete.

PR #124, opened 2021-06-03
[package_info_plus] Migrate package_info code to package_info_plus

HakkyuKim:
part of #116 

Same problem exists in setting the Tizen platform interface, but needs to handle it differently than #120.

Pleas See: https://github.com/fluttercommunity/plus_plugins/blob/c7c942d1d19595a536c6ae18168d3781ddef7785/packages/package_info_plus/lib/package_info_plus.dart#L43-L55


packages/package_info_plus/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
+name: package_info_plus_tizen
+description: Tizen implementation of the package_info_plus plugin
+version: 1.0.0
+homepage: https://github.com/fluttercommunity/plus_plugins/tree/main/packages/package_info_plus
```
swift-kim, line 4:
```suggestion
homepage: https://github.com/flutter-tizen/plugins
```

swift-kim (approved):
Otherwise looks good to me!

bwikbs (approved):
Thanks for your hard working! :+1:

Issue #99, opened 2021-05-13
[connectivity] Crash when terminate app


bbrto21:
**How to reproduce** 
1. Run example : flutter-tizen run
2. Terminate example immediately  after launching app via press 'Q' key

**Callstack**
```
* thread #33, name = 'DartWorker', stop reason = signal SIGABRT
  * frame #0: 0xb6be24fc libc.so.6`__GI_raise + 156
    frame #1: 0xb6bcc788 libc.so.6`__GI_abort + 276
    frame #2: 0xb6bdb3d0 libc.so.6`__assert_fail_base + 368
    frame #3: 0xb6bdb480 libc.so.6`__GI___assert_fail + 68
    frame #4: 0xb66b3f40 libpthread.so.0`__pthread_mutex_lock + 596
    frame #5: 0xa79b2040 libflutter_engine.so`std::__1::mutex::lock() + 8
    frame #6: 0xa7e17dd4 libflutter_engine.so`flutter::ServiceProtocol::HandleListViewsMethod(rapidjson::GenericDocument<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator>, rapidjson::CrtAllocator>*) const + 100
    frame #7: 0xa7e17c6a libflutter_engine.so`flutter::ServiceProtocol::HandleMessage(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::map<std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::less<std::__1::basic_string_view<char, std::__1::char_traits<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string_view<char, std::__1::char_traits<char> > const, std::__1::basic_string_view<char, std::__1::char_traits<char> > > > > const&, rapidjson::GenericDocument<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator>, rapidjson::CrtAllocator>*) const + 158
    frame #8: 0xa7e178ba libflutter_engine.so`flutter::ServiceProtocol::HandleMessage(char const*, char const**, char const**, int, void*, char const**) + 210
    frame #9: 0xa7f166d6 libflutter_engine.so`dart::Service::EmbedderHandleMessage(handler=<unavailable>, js=<unavailable>) at service.cc:1240
[...]

PR #125, opened 2021-06-04
[connectivity_plus] Migrate connectivity code to connectivity_plus

HakkyuKim:
part of #116 

Setting the Tizen platform interface is done similarly with #120.


HakkyuKim:
Trying to figure out #99.

HakkyuKim:
> Trying to figure out #99.

I'm unable to solve #99, I have no clue why this error occurs. You have to press 'q' **instantly** when the first app screen appears. I was able to reproduce the issue only on mobile devices. The chances of users encountering this error may be slim, but I do hope to know what might be causing it. Does anyone have any ideas?

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+        std::make_unique<flutter::MethodChannel<flutter::EncodableValue>>(
+            registrar->messenger(), "dev.fluttercommunity.plus/connectivity",
+            &flutter::StandardMethodCodec::GetInstance());
+    m_event_channel =
+        std::make_unique<flutter::EventChannel<flutter::EncodableValue>>(
+            registrar->messenger(),
+            "dev.fluttercommunity.plus/connectivity_status",
+            &flutter::StandardMethodCodec::GetInstance());
+    method_channel->SetMethodCallHandler([this](const auto &call, auto result) {
+      this->HandleMethodCall(call, std::move(result));
```
bbrto21, line 138:
```suggestion
      HandleMethodCall(call, std::move(result));
```

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    auto event_channel_handler =
+        std::make_unique<flutter::StreamHandlerFunctions<>>(
+            [this](const flutter::EncodableValue *arguments,
+                   std::unique_ptr<flutter::EventSink<>> &&events)
+                -> std::unique_ptr<flutter::StreamHandlerError<>> {
+              LOG_INFO("OnListen");
+              this->registerObsever(std::move(events));
```
bbrto21, line 147:
```suggestion
              registerObsever(std::move(events));
```

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+              return nullptr;
+            },
+            [this](const flutter::EncodableValue *arguments)
+                -> std::unique_ptr<flutter::StreamHandlerError<>> {
+              LOG_INFO("OnCancel");
+              this->clearObserver();
```
bbrto21, line 153:
```suggestion
              clearObserver();
```

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+        result = "none";
+    }
+    return result;
+  }
+
+  void HandleMethodCall(
+      const flutter::MethodCall<flutter::EncodableValue> &method_call,
+      std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
+    ensureConnectionHandle();
+    LOG_INFO("method : %s", method_call.method_name().data());
```
bbrto21, line 104: Should we still care about [this](https://github.com/flutter-tizen/flutter-tizen/issues/71)?

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  virtual ~ConnectivityPlusTizenPlugin() {
+    if (m_connection != nullptr) {
+      connection_destroy(m_connection);
+      m_connection = nullptr;
+    }
+  }
+
+  void registerObsever(
```
bbrto21, line 43: Could you please change method names to start with a capital letter?

packages/connectivity_plus/example/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
...
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
+  integration_test_tizen:
+    path: ../../integration_test/
+
+dependency_overrides:
+  args: ^2.0.0
```
bbrto21, line 23: I don't think this is needed anymore
```suggestion
```

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+#include <memory>
+#include <sstream>
+#include <string>
+
+#include "log.h"
+
+class ConnectivityPlusTizenPlugin : public flutter::Plugin {
+ public:
+  static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar) {
+    LOG_INFO("RegisterWithRegistrar");
```
bbrto21, line 26: Seems like an unnecessary log.

packages/connectivity_plus/example/pubspec.yaml
HakkyuKim: Recent migration to Flutter 2.2.1(args 1.6.0 -> args 2.1.0) by flutter-tizen/flutter-tizen/pull/138 solved this. Thanks for noticing!

packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
HakkyuKim: I believe statically linking higher version libstdc++ by flutter-tizen/flutter-tizen/pull/106 solved the issue.

bbrto21: Oh you're right! I forgot this :smiley:

HakkyuKim:
@bbrto21 
Thanks for the review, I applied all your suggestions.

bbrto21 (approved):
LGTM, latest original plugin version is 1.0.4. Are you sure it's 1.0.1?

swift-kim:
@bbrto21 The plugins are versioned separately, so their versions have nothing to do with each other.

HakkyuKim:
> LGTM, latest original plugin version is 1.0.4. Are you sure it's 1.0.1?

You are talking about `connectivity_plus`? 
I will also update `connectivity_plus_platform_interface`.

BTW, I should update pubspec.yaml environment right?

> Recent migration to Flutter 2.2.1(args 1.6.0 -> args 2.1.0) by [flutter-tizen/flutter-tizen/pull/138](https://github.com/flutter-tizen/flutter-tizen/pull/138) solved this. Thanks for noticing!

```diff
environment:
    sdk: ">=2.12.0 <3.0.0"
-    flutter: ">=2.0.0"
+   flutter: ">=2.2.1"
```

bbrto21:
yes, I mean [connectivity_plus](https://pub.dev/packages/connectivity_plus) version   
```yaml
dependencies:
  connectivity_plus: ^1.0.1
```


> BTW, I should update pubspec.yaml environment right?

I think we can leave it

HakkyuKim:
@bbrto21 
Actually, I just found out that `args` got bumped up to 2.1 in flutter sdk 2.2.0 release.

If we remove:
```yaml
dependency_overrides:
  args: ^2.0.0
```
then minimum sdk version must be 2.2.0 because older version can't build it.

We can either leave `dependency_overrides`, or remove it and update min flutter version requirement. 
I'm slightly leaning towards just leaving it as done in `connectivity_plus`, but not quite sure which is a better decision. I'll think about it.

bbrto21:
> We can either leave `dependency_overrides`, or remove it and update min flutter version requirement.
> I'm slightly leaning towards just leaving it as done in `connectivity_plus`, but not quite sure which is a better decision. Any thoughts?

I think it's better to remove the dependency override.

packages/connectivity_plus/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+flutter:
+  plugin:
+    platforms:
+      tizen:
+        dartPluginClass: ConnectivityPlugin
+        pluginClass: ConnectivityPlusTizenPlugin
+        fileName: connectivity_plus_tizen_plugin.h
+
+dependencies:
+  connectivity_plus_platform_interface: ^1.0.0
```
swift-kim, line 15: Are you going to update this to 1.0.2?

swift-kim (approved):

HakkyuKim:
OK, so I came with the following conclusion. Because `connectivity_plus_tizen_example` has the following dependencies:

- connectivity_plus_tizen_example -> connectivity_plus -> connectivity_plus_linux
- connectivity_plus_tizen_example -> integration_test -> flutter sdk

And because the error message is:

```
pub get failed (1; So, because connectivity_plus_example depends on both connectivity_plus ^1.0.1 and integration_test any from sdk, version solving failed.)
```

To run integration test with `connectivity_plus`, you need to set flutter sdk to at least 2.2.0. This doesn't have to do with `connectivity_plus_tizen`, so you can ignore my original comment.

After changing `pubspec.yaml` of example package as follows:

```diff
# connectivity_plus_tizen_example
environment:
  sdk: ">=2.12.0 <3.0.0"
-  flutter: ">=2.0.0"
+  flutter: ">=2.2.0"
```

Now the error message is more clear on how to build the app:

```
The current Flutter SDK version is 2.0.1.                               
                                                                        
Because connectivity_plus_example requires Flutter SDK version >=2.2.0, version solving failed.
Running "flutter pub get" in example...                                 
pub get failed (1; Because connectivity_plus_example requires Flutter SDK version >=2.2.0, version solving failed.)
```

packages/connectivity_plus/pubspec.yaml
HakkyuKim: Yes, I'll update it together with the sdk version issue!

HakkyuKim:
@bbrto21, @swift-kim 
I'll merge this if the last version change makes sense.

I had to rebase to run test, only the last commit has actually changed:
"Update dependency versions 3d4d218"

swift-kim:
@HakkyuKim Okay, thanks!

bbrto21:
@HakkyuKim good!

PR #130, opened 2021-06-16
[device_info_plus] Migrate device_info code to device_info_plus

HakkyuKim:
part of #116


packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -0,0 +1,295 @@
...
+    int ret;
+    bool hasResult = false;
+    std::string resultStr;
+    std::string errorStr;
+
+    // http://tizen.org/system/model_name
+    ret = system_info_get_platform_string("http://tizen.org/system/model_name",
+                                          &value);
+    if (ret != SYSTEM_INFO_ERROR_NONE) {
+      resultStr = get_error_message(ret);
```
WonyoungChoi, line 61: `hasResult` will be true if any one of the items to get information succeeds. If it fails to obtain information from this item, this error message can be displayed as a normal value. It is difficult for users to distinguish between normal value and error message. In error situations, it seems good to exclude this item or provide an empty string.

bwikbs: >  If it fails to obtain information for this item,

Can you tell me what is that case? (error message sample?)

WonyoungChoi: @bwikbs for example, if it fails to get modelName, but all other values succeeded, the result would be:
```json
 modelName: "some error message", //  (in case of failure getting info from CAPI)
 cpuArch: "armv7l",
 platformVersion: "6.5.0",
 ....
```

bwikbs: Ah.. Now I got point! 
At first, We handled an error, but when we did that, the test failed, so it came to this.
We also considered empty, but then there is no way to show the error message..

WonyoungChoi: @bwikbs Well.. IMO, this function is not a critical function to send an error message directly to the user. We just need to return as much information as we can return. Rather, it is more useful to let the user check that the value is returned correctly. In the current impl, it is absolutely necessary to check whether the returned value is an error message or not.

WonyoungChoi: How about just logging error message instead of sending it?

bwikbs: @WonyoungChoi I also feel that your intellectual part is right. But I don't have a opinion because I'm not sure how to do it right.  :smile:

packages/device_info_plus/lib/device_info_plus_tizen.dart
```diff
@@ -0,0 +1,155 @@
...
+  /// The value is an empty String if it is not available.
+  final String deviceType;
+
+  /// http://tizen.org/system/manufacturer
+  ///
+  /// The value is an empty String if it is not available.
+  final String manufacturer;
+
+  /// Deserializes from the message received from [_kChannel].
+  static TizenDeviceInfo fromMap(Map<String, dynamic> map) {
```
WonyoungChoi, line 104: @HakkyuKim I'm just curious and asking. I am a dart newbie. :) why do you use static method here instead of factory constructor?

swift-kim: It's not always the case but generally speaking, the factory constructor should be used only when it's necessary. (e.g. The constructor might not create a new instance and instead return an instance from a cache, or it might return an instance of a subtype.)

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
swift-kim: It's always the best practice to follow implementations of other platforms.

packages/device_info_plus/lib/device_info_plus_tizen.dart
HakkyuKim: @WonyoungChoi
I'm glad to ponder about this intersting question.
Honestly, I haven't given much thought into the current code style, since the focus of this PR was to change only the necessary code to work with the `device_info_plus` package. 

`fromMap` reminded me of a similar method named `fromJson` from [Json serializing documentation](https://flutter.dev/docs/development/data-and-backend/json#serializing-json-inside-model-classes).
At the time, I didn't notice that `fromJson` was a named constructor because a static method that returns a new instance and a named constructor both seemed appropriate in simple data type transformation(`Map` to `DeviceInfoTizen`).

However, if the method was a factory constructor, to me at least, would have given me the impression that some untrivial actions may be performed additionally before returning an instance.
Nevertheless, a named constructor seems more appropriate, as done with `fromJson`, for consistency sake and becuase it always guarentees returning an instance.

@swift-kim also made a good point about factory constructors canonically being used to return an instance of a subtype.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
HakkyuKim: @WonyoungChoi, @bwikbs
Thanks for the suggestion and the history of the current implementation.

@swift-kim
I quickly skimmed the implementations of other platforms, they seem to return null for fields that may not be available in some systems, but assumes that it's always possible to return data for non-nullable fields. I will provide more precise information after looking into it.

packages/device_info_plus/lib/device_info_plus_tizen.dart
HakkyuKim: All other platform implementations of `XXXDeviceInfo` use static method `fromMap`, `TizenDeviceInfo` should follow this for consistency. See https://github.com/fluttercommunity/plus_plugins/blob/a50f5ada496155d71c98040043b642bfdaa48b49/packages/device_info_plus_platform_interface/lib/model/android_device_info.dart#L114-L139 for Android.

WonyoungChoi: Using the factory keyword and implementing the factory pattern with a static method are functionally the same. In the end it doesn't matter. 😅 I found the related article about this: https://dash-overflow.net/articles/factory/. 
@HakkyuKim  Thanks for your kind replies to my questions.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
HakkyuKim: Based on the implementations of other platforms---specifically Android and Linux---I think it's best to send an empty string for unavailable values like @WonyoungChoi suggessted, and log error messages to Console.

As @bwikbs said, then there is no way to send the error message to dart code, but I think platform interface was designed so that it doesn't care why the values were unavailable, it just understands that some values may be missing and handles them appropriately. For example, by setting default values for non-nullable fields.

Since we implement our own `DeviceInfoPlatform`, we can make dart code receive error messages if we use `BasicMethodChannel` instead of `MethodChannel`, but I don't see the need to make the code different from `device_info_plus` just for this reason.

## Suggestion

As-is

```cpp
ret = system_info_get_platform_string("http://tizen.org/system/model_name",
                                        &value);
if (ret != SYSTEM_INFO_ERROR_NONE) {
    resultStr = get_error_message(ret);
    errorStr.append(" Failed to get model name: ");
    errorStr.append(resultStr);
} else {
    hasResult = true;
    resultStr = std::string(value);
    LOG_INFO("model name : %s\n", value);
    free(value);
}
msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
    "modelName", resultStr));
```

To-be

```cpp
ret = system_info_get_platform_string("http://tizen.org/system/model_name",
                                        &value);
if (ret == SYSTEM_INFO_ERROR_NONE) {
    resultStr = std::string(value);
    free(value);
} else {
    resultStr = "";
    LOG_ERROR("Failed to get modelName from the system: %s", get_error_message(ret));
}

msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
    "modelName", resultStr));
```

On a different note, since we're already making code changes, how about this to make the code shorter?

```cpp
std::vector<std::pair<std::string, std::string>> requests = {
      {"modelName", "http://tizen.org/system/model_name"},
      {"cpuArch", "http://tizen.org/feature/platform.core.cpu.arch"},
      {"nativeApiVersion",
       "http://tizen.org/feature/platform.native.api.version"},
      {"platformVersion", "http://tizen.org/feature/platform.version"},
      {"webApiVersion", "http://tizen.org/feature/platform.web.api.version"},
      {"buildDate", "http://tizen.org/system/build.date"},
      {"buildId", "http://tizen.org/system/build.id"},
      {"buildString", "http://tizen.org/system/build.string"},
      {"buildTime", "http://tizen.org/system/build.time"},
      {"buildType", "http://tizen.org/system/build.type"},
      {"buildVariant", "http://tizen.org/system/build.variant"},
      {"buildRelease", "http://tizen.org/system/build.release"},
      {"deviceType", "http://tizen.org/system/device_type"},
      {"manufacturer", "http://tizen.org/system/manufacturer"},
  };

for (auto &request : requests) {
    std::string &map_key = request.first;
    std::string &tizen_key = request.second;
    std::string resultStr = "";
    int ret = system_info_get_platform_string(tizen_key.c_str(), &value);
    if (ret == SYSTEM_INFO_ERROR_NONE) {
      resultStr = std::string(value);
      free(value);
    } else {
      LOG_ERROR("Failed to get %s from the system: %s", tizen_key.c_str(),
              get_error_message(ret));
    }
    msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
        map_key, resultStr));
}

result->Success(flutter::EncodableValue(msg));
```

bwikbs: Wow It's really impressive! :scream: 
Thanks for the detailed explanation! I love it :heart:
It seems to be much better than before.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -49,242 +49,43 @@ class DeviceInfoPlusTizenPlugin : public flutter::Plugin {
...
-      errorStr.append(resultStr);
-    } else {
-      hasResult = true;
-      resultStr = std::string(value);
-      LOG_INFO("model name : %s\n", value);
-      free(value);
+    for (auto &request : requests_) {
+      const std::string &map_key = request.first;
+      const std::string &tizen_key = request.second;
+      std::string resultStr = "";
```
swift-kim, line 56: Violates the naming convention.

Suggestion:
```suggestion
      std::string result("");
```

packages/device_info_plus/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+flutter:
+  plugin:
+    platforms:
+      tizen:
+        pluginClass: DeviceInfoPlusTizenPlugin
+        fileName: device_info_plus_tizen_plugin.h
+
+dependencies:
+  device_info_plus_platform_interface: ^1.0.0
```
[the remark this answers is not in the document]
HakkyuKim: Upgraded version dependency. The changes only affect web platforms, so no code changes were made.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
HakkyuKim: Done!

swift-kim (approved):

PR #134, opened 2021-06-21
Keep EventChannel in memory for the lifetime.

### LINKED ISSUE #131: Invalid use of EventChannel in some plugins
I analyzed some crash on TM1 recently and it turns out that there is a problem with use of `EventChannel`.

There are some places, where developers create `EventChannel` as local object which is deleted at the exit of the function. Unfortunately we can't do this, because `EventChannel` not only passes handlers to engine, but also keeps `is_listening_` variable. This variable is used in [binary_handler](https://github.com/flutter-tizen/engine/blob/603031f27b70f1dc943e958fff103ef526ef7418/shell/platform/common/client_wrapper/include/flutter/event_channel.h#L64) and also in [SetStreamHandler](https://github.com/flutter-tizen/engine/blob/603031f27b70f1dc943e958fff103ef526ef7418/shell/platform/common/client_wrapper/include/flutter/event_channel.h#L53) to cancel listening - without it there might be some calls to handler even after removing the handler.
But as far as I know, everything on platform channel is being called on one thread, so there should be no issue if we would just remove `is_listening_` from `EventChannel` and make it local to `binary_helper` lambda.
What do you think about this solution? One drawback is that we would need to make this change in upstream.

The second solution is to fix the code of plugins - **wearable_rotary_plugin**, **video_player**, **sensors**, **connectivity** and **battery** are affected.

I think that the second approach is better.

rwalczyna:
EventChannels have to be preserved in memory in order to work properly.
I tested modified plugins, no issues occurred.

It fixes #131 

Signed-off-by: Rafal Walczyna <r.walczyna@samsung.com>

### DIFF
--- a/packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
+++ b/packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
@@ -22,43 +22,8 @@ class BatteryPlusTizenPlugin : public flutter::Plugin {
  public:
   static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar) {
     LOG_DEBUG("RegisterWithRegistrar for BatteryPlusTizenPlugin");
-    auto method_channel =
-        std::make_unique<flutter::MethodChannel<flutter::EncodableValue>>(
-            registrar->messenger(), "dev.fluttercommunity.plus/battery",
-            &flutter::StandardMethodCodec::GetInstance());
-    auto event_channel =
-        std::make_unique<flutter::EventChannel<flutter::EncodableValue>>(
-            registrar->messenger(), "dev.fluttercommunity.plus/charging",
-            &flutter::StandardMethodCodec::GetInstance());
-
     auto plugin = std::make_unique<BatteryPlusTizenPlugin>();
-
-    auto method_channel_handler = [plugin_pointer = plugin.get()](
-                                      const auto &call, auto result) {
-      LOG_DEBUG("HandleMethodCall call");
-      plugin_pointer->HandleMethodCall(call, std::move(result));
-    };
-    auto event_channel_handler =
-        std::make_unique<flutter::StreamHandlerFunctions<>>(
-            [plugin_pointer = plugin.get()](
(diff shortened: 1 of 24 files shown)


swift-kim (approved):
Looks good to me.

It's not mandatory, but also consider increasing the plugins versions (in pubspec.yaml and README.md) and updating CHANGELOGs if you want to publish these changes.

HakkyuKim (commented):
Updated #125

bwikbs (approved):
It must be hard to find, thank you! :+1:

pkosko (approved):
LGTM

packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
```diff
@@ -193,6 +158,43 @@ class BatteryPlusTizenPlugin : public flutter::Plugin {
     result->Success(flutter::EncodableValue(percentage));
   }
 
+ private:
+  void SetupChannels(flutter::PluginRegistrar *registrar) {
+    auto method_channel =
+        std::make_unique<flutter::MethodChannel<flutter::EncodableValue>>(
+            registrar->messenger(), "dev.fluttercommunity.plus/battery",
+            &flutter::StandardMethodCodec::GetInstance());
+    m_event_channel =
```
bbrto21, line 167: Could you change prefix ```m_``` to postfix ```_``` to represent member of class?
I think this suggestion is out of scope, but I'd appreciate it if you could

swift-kim: IMO, renaming the variables with a new style is quite a big change which should be done in a separate commit. I still hope someone to do the work for all our packages.

rwalczyna: @bbrto21, I didn't want to mix `m_` and `_` in one file, so I decided to use `m_` in part of the plugins. I don't like it either :)

bbrto21:
plz, rebase

PR #146, opened 2021-07-09
[webview_flutter] Apply PlatformView's Intreface

bwikbs:
Now, This PR only contains PlatformView's Intreface change.

### DIFF
--- a/packages/webview_flutter/pubspec.yaml
+++ b/packages/webview_flutter/pubspec.yaml
@@ -1,7 +1,7 @@
 name: webview_flutter_tizen
 description: Tizen implementation of the webview plugin
 homepage: https://github.com/flutter-tizen/plugins
-version: 0.2.2
+version: 0.3.0
 
 environment:
   sdk: ">=2.12.0 <3.0.0"
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -157,8 +157,9 @@ double ExtractDoubleFromMap(const flutter::EncodableValue& arguments,
 
 WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
                  flutter::TextureRegistrar* texture_registrar, double width,
-                 double height, flutter::EncodableMap& params, void* winHandle)
-    : PlatformView(registrar, viewId, winHandle),
+                 double height, flutter::EncodableMap& params,
+                 void* platform_window)
+    : PlatformView(registrar, viewId, platform_window),
       texture_registrar_(texture_registrar),
       webview_instance_(nullptr),
       width_(width),
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -145,14 +145,20 @@ int ExtractIntFromMap(const flutter::EncodableValue& arguments,
   }
   return -1;
 }
-double ExtractDoubleFromMap(const flutter::EncodableValue& arguments,
-                            const char* key) {
-  if (std::holds_alternative<flutter::EncodableMap>(arguments)) {
-    flutter::EncodableMap values = std::get<flutter::EncodableMap>(arguments);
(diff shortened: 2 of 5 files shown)


swift-kim:
This package is an immediate successor of webview_flutter_tizen, right? Shouldn't we just change the directory name (`webview_flutter` to `webview_flutter_lwe`) and discontinue the original package instead of copying all the files and maintaining two sets of source code?

pub.dev admin option:

> image

Also you need to update README in the root of this repo.

Also since there has been a breaking change in the PlatformView API (https://github.com/flutter-tizen/engine/pull/136), please let me know when this PR is ready for merge so that I can create an engine release.

bwikbs:
As we discussed, It's decided to keep the existing packages and only add ewk webview.
Now this PR contains only interface changes.

packages/webview_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: webview_flutter_tizen
 description: Tizen implementation of the webview plugin
 homepage: https://github.com/flutter-tizen/plugins
-version: 0.2.2
+version: 0.2.3
```
swift-kim, line 4: Since this is a breaking change (the engine update is required), you might want to increase the version to 0.3.0.

bwikbs: Ok thanks :smile:

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -157,8 +157,8 @@ double ExtractDoubleFromMap(const flutter::EncodableValue& arguments,
 
 WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
                  flutter::TextureRegistrar* texture_registrar, double width,
-                 double height, flutter::EncodableMap& params)
-    : PlatformView(registrar, viewId),
+                 double height, flutter::EncodableMap& params, void* winHandle)
```
bbrto21, line 160:
```suggestion
                 double height, flutter::EncodableMap& params, void* win_handle)
```

packages/webview_flutter/tizen/src/webview.h
```diff
@@ -27,7 +27,7 @@ class WebView : public PlatformView {
  public:
   WebView(flutter::PluginRegistrar* registrar, int viewId,
           flutter::TextureRegistrar* textureRegistrar, double width,
-          double height, flutter::EncodableMap& params);
+          double height, flutter::EncodableMap& params, void* winHandle);
```
bbrto21, line 30:
```suggestion
          double height, flutter::EncodableMap& params, void* win_handle);
```

packages/webview_flutter/tizen/src/webview.cc
bwikbs: Changed same as the name of the platform view.

packages/webview_flutter/tizen/src/webview.h
bwikbs: Changed same as the name of the platform view.

swift-kim (approved):

bbrto21 (approved):

bwikbs:
> please let me know when this PR is ready for merge so that I can create an engine release.

@swift-kim Now, It's ready I guess

swift-kim:
@bwikbs I'll not be at the office until tomorrow, so could you do some ad hoc testing (like running simple apps on various devices including watch/watch emulator/TV) of the current latest engine before publishing a release?

bwikbs:
@swift-kim Sure, no problem!
BTW, https://github.com/flutter-tizen/engine/pull/142 include breaking change also. 
So based on that I will update and test this PR

bwikbs:
I've tested webview example & hello world. ( include flutter-tizen/engine#142)
Here is result:

- [x] TM1
- [x] FHUB
- [ ] TW3
- [ ] TV

swift-kim:
@bwikbs Thanks. You don't have to take too much time for testing. It's just for finding major bugs like app not launching properly.

PR #147, opened 2021-07-09
fix broken tts integration test and fix file names

HakkyuKim:


packages/flutter_tts/example/integration_test/flutter_tts_test.dart
```diff
@@ -11,6 +11,6 @@ void main() {
   testWidgets('Can speak', (WidgetTester tester) async {
     final FlutterTts flutterTts = FlutterTts();
     var result = await flutterTts.speak("Hello World");
-    expect(result, 1);
+    expect(result, null);
```
swift-kim, line 14: The flutter_tts plugin has a design defect, and almost all its functions are expected to return `Success(1)` on success and `Success(0)` on failure (not `Error`). It's even not documented very well so I referred to the Android and macOS implementations. So according to the fact, this test case seems to be correct. But I'm not sure whether we have to change our current implementation.

HakkyuKim: Hmm didn't know about it until just now, and I see your point.
I don't think we should use the API incorrectly when there's no documentation that its intentional.

Still, since the current integration test with our implementation fails, we can remove 
the test code and leave a issue about it to revisit later.

swift-kim: Okay, you might add `skip: true` to `testWidgets()` for now.

HakkyuKim: That's better. Thanks.

swift-kim (approved):

bbrto21 (approved):
LGTM

Issue #148, opened 2021-07-09
Running integration tests with CI service


HakkyuKim:
I tried several ways to run emulators in CI server for integration tests but concluded that it just cannot be done without digging into how Tizen emulators work under the hood. I did manage to run the emulator, but sdb can't find it: https://cirrus-ci.com/task/5968653813284864 (test branch: https://github.com/HakkyuKim/plugins/tree/cirrus).

Look for:

```sh
em-cli launch --name flutter_emulator
Launch command : "/home/tizen/tizen-studio/platforms/tizen-4.0/common/emulator/bin/emulator.sh" --conf "/home/tizen/tizen-studio-data/emulator/vms/flutter_emulator/vm_launch.conf" -j "/home/tizen/tizen-studio/jdk/bin/java"
The VM is launched
...

sdb devices
List of devices attached 

```

So instead, I'm considering running integration tests on our machine that's connected to physical targets using Azure Pipelines [self-hosted agent](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#install).

There are a few considerations:

- Corporate policy regarding network security (I'll start asking around next week)
- Pricing
    According to [pricing for Azure DevOps](https://azure.microsoft.com/en-us/pricing/details/devops/azure-devops-services/), we need at minimum "User License: Basic Plan", which is free for the first 5 users.
- Cost of maintenance 
    The host machine and the Tizen devices attached to it.

Personally I think it's the best option to do CI for plugin integration tests and possibly future engine regression tests. Please share any ideas or suggestions.

TODOs:

- [x] write test tool script (reference: [flutter-plugin-tools](https://github.com/flutter/plugins/tree/master/script/tool)) #150 
- [x] fix broken integration tests or exclude them in CI task: video_players, integration_test, etc #169 #182 
- [x] write ~~azure pipelines~~github workflow script #169 
- [x] think about how to let the tool know about unsupported device profiles of each plugin https://github.com/flutter-tizen/plugins/pull/191

swift-kim:
Regarding the pricing, we don't need to buy user licenses. We just need to add a self-hosted agent which costs $15 per month.

![image](https://user-images.githubusercontent.com/40190588/125051322-cd330800-e0dd-11eb-916d-e15ade4638b4.png)

HakkyuKim:
Using two self hosts is because one is being used by the engine repo right? Then yes $15 per month. 

[...]

PR #150, opened 2021-07-14
Running multiple integration tests script

HakkyuKim:
Script that runs multiple integration tests locally and aggregates the result. Implementation is referenced from [flutter-plugin-tools](https://github.com/flutter/plugins/tree/master/script/tool). Please see the help message(`./tools/run_integration_test.py -h`) for details oh how to use the script.

Contributes to #148 

Example Command:

```sh
# exclude plugins that don't have tests
# exclude failing tests
./tools/run_integration_test.py
```

Output:

```
FAILED: wearable_rotary
Missing integration tests (use --exclude if this is intentional).
FAILED: audioplayers
Missing integration tests (use --exclude if this is intentional).
FAILED: webview_flutter
flutter-tizen test integration_test failed, see the output above for details.
SUCCEEDED: network_info_plus
FAILED: package_info_plus
Timeout expired. The test may need more time to finish.
If you expect the test to finish before timeout, check if the tests 
require device screen to be awake or if they require manually 
clicking the UI button for permissions.
SUCCEEDED: battery
FAILED: image_picker
No tests ran.
SUCCEEDED: wifi_info_flutter
FAILED: camera
Timeout expired. The test may need more time to finish.
If you expect the test to finish before timeout, check if the tests 
require device screen to be awake or if they require manually 
clicking the UI button for permissions.
SUCCEEDED: sensors_plus
SUCCEEDED: flutter_tts
SUCCEEDED: path_provider
SUCCEEDED: device_info
SUCCEEDED: shared_preferences
SUCCEEDED: sensors
FAILED: integration_test
Timeout expired. The test may need more time to finish.
If you expect the test to finish before timeout, check if the tests 
require device screen to be awake or if they require manually 
clicking the UI button for permissions.
SUCCEEDED: url_launcher
SUCCEEDED: device_info_plus
SUCCEEDED: share
FAILED: video_player
Timeout expired. The test may need more time to finish.
If you expect the test to finish before timeout, check if the tests 
require device screen to be awake or if they require manually 
clicking the UI button for permissions.
SUCCEEDED: messageport
SUCCEEDED: battery_plus
SUCCEEDED: connectivity
SUCCEEDED: connectivity_plus
FAILED: package_info
Timeout expired. The test may need more time to finish.
If you expect the test to finish before timeout, check if the tests 
require device screen to be awake or if they require manually 
clicking the UI button for permissions.
SUCCEEDED: wakelock
FAILED: permission_handler
flutter-tizen test integration_test failed, see the output above for details.
SUCCEEDED: share_plus
```

Currently each plugin test will all run on the same connected device, and it cannot skip tests for plugins that aren't supported with that device's profile.


swift-kim:
How is this related to the flutter_plugin_tools package? What's common between your implementation and flutter_plugin_tools?

HakkyuKim:
@swift-kim 
The flutter_plugin_tools package is a command line app and it has several features besides running integration tests. This python script referenced implementations of some commands related to running integration tests:

- command line interface design: argument naming, descriptions, and constraints, ...
- aggregation strategy: Defining test result states into success, fail, and skip. Implementations on what and how log messages are displayed.
- test policy: Which condition is considered a test failure? For example, plugins without integration tests is a test failure unless explicitly specifying it with the `--exclude` option.
- CI consideration: Adopting `--run-on-changed-packages` option for PR triggers.

bbrto21:
Thanks for making such a great tool!
But while trying this out, I found something weird!

**flutter tool logs**
```bash
flutter-tizen drive --driver=test_driver/integration_test.dart --target=integration_test/path_provider_test.dart
Running "flutter pub get" in example...                            546ms
The platform profile is used for signing.                               
Building a Tizen application in debug mode...                      28.8s
✓ Built build/tizen/tpk/org.tizen.path_provider_tizen_example-1.0.0.tpk (25.8MB).
Installing build/tizen/tpk/org.tizen.path_provider_tizen_example-1.0.0.tpk...        24.4s
VMServiceFlutterDriver: Connecting to Flutter application at http://127.0.0.1:45413/sCDkqKWjWBg=/
VMServiceFlutterDriver: Isolate found with number: 1779345267457467
VMServiceFlutterDriver: Isolate is paused at start.
VMServiceFlutterDriver: Attempting to resume isolate
VMServiceFlutterDriver: Connected to Flutter application.
All tests passed.
```

**dlogs**
```bash
...
I/ConsoleMessage(18728): flutter: 00:03 +6 -7: (tearDownAll)
I/ConsoleMessage(18728): flutter: 00:03 +7 -7: Some tests failed.
```

HakkyuKim:
@bbrto21 
I've checked the test code in `path_provider` and the results in the dlog is correct, the integration test shouldn't have passed. It seems that the drive command doesn't return the correct results. I'm currently not sure what's causing this. I'll look into it.

HakkyuKim:
Although the problem mentioned by @bbrto21 is not caused by this script, the problem should be fixed for the script to work correctly. I'll work on that first.

HakkyuKim:
With https://github.com/flutter-tizen/flutter-tizen/pull/177, it's now possible to display correct results of integration_test using the command `flutter-tizen test integration_test`. The script is updated to use this command instead of `flutter-tizen drive`. Thanks @bbrto21 for noticing the problem, and @swift-kim for providing solutions to make the test command work.

Some plugins fail integration tests because they either require the screen to be awake or require manually clicking the UI button for permission requests.

bbrto21 (approved):
LGTM

tools/run_integration_test.py
```diff
@@ -0,0 +1,315 @@
...
+        if not completed_process.stderr:
+            errors.append('pub get failed. Make sure the pubspec file \
+                    in your project is valid.')
+        else:
+            errors.append(completed_process.stderr)
+        return PluginResult.fail(errors)
+
+    # This prevents ci test failures when depending packages accidentally
+    # publishes breaking changes as minor or patch versions.
+    # (e.g. package_info_plus 1.0.4 breaks PackageInfo's public constructor.)
```
swift-kim, line 139: Shouldn't this error be detected by the test tool?

tools/run_integration_test.py
```diff
@@ -0,0 +1,315 @@
...
+        help='Specifies which plugins to test. If it is not specified and \
+            --run-on-changed-packages is also not specified, \
+                then it will include every plugin under packages.')
+    parser.add_argument('--exclude',
+                        type=str,
+                        nargs='*',
+                        default=[],
+                        help='Exclude plugins from test.')
+    parser.add_argument(
+        '--run-on-changed-plugins',
```
swift-kim, line 70:
```suggestion
        '--run-on-changed-packages',
```
Use consistent terms (line 62 and 80).

swift-kim (commented):
Some nitpicks. Overall looks good to me!

tools/run_integration_test.py
HakkyuKim: Good point, yes I think it should be detected and we should fix the problem when such error occurs. According to https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version, `package_info_plus` should release a new patch version that restores backward compatibility. I've reported to the maintainers: https://github.com/fluttercommunity/plus_plugins/issues/392.

Maybe I could add an option like "--run-on-downgraded-packages", we might encounter similar problems in the future once the CI has been setup.

swift-kim: To be more specific, the following changes to platform interface classes will break existing Tizen implementations (FFI-based packages only).

- Adding a property without providing a default implementation
- Changing an existing methods signature (such as adding an optional parameter)

What we have to do in such situation is to update our implementations as soon as possible. The errors are not false alarm, as the users will encounter exactly the same error when depending on our packages. Expecting the users to run `flutter pub downgrade` by themselves doesn't make sense because they will not even understand the situation.

HakkyuKim: Let me see if I've got it right.

If the package we depend on(let's say A) make those mentioned changes, normally it will be released after increasing the major version, for example, 1.0.0 to 2.0.0. At this point if our package(let's say B)'s pubspec was:

```
dependency:
    A: ^1.0.0
```

B will still build normally because pub tool doesn't automatically get the next higher major version. 

But if A was accidentally released as the next minor version, for example, 1.0.0 to 1.1.0, the pub tool will get A-1.1.0 and the build will fail due to breaking changes.

My initial thought was to wait for A to fix the problem by releasing a new minor version that removes all breaking changes and then release a new major version (what the maintainers probably wanted to do in the first place). And In the meantime allow the CI to pass with the minimum stable(?) version until A is fixed.

Instead in such cases, we should let the CI fail and resolve the breaking changes immediately because:

> as the users will encounter exactly the same error when depending on our packages. Expecting the users to run flutter pub downgrade by themselves doesn't make sense because they will not even understand the situation.

Is this correct?

swift-kim: > My initial thought was to wait for A to fix the problem by releasing a new minor version that removes all breaking changes and then release a new major version

Changing a method signature or adding a property is a reasonably "minor" change (although it's against backwards compatibility), and the maintainers of platform interface packages do not increase the major version in such situation. It's not mandatory for the maintainers to follow the versioning policy strictly, and we can't force them to follow it.

So yes, if any breaking change has (unfortunately) been made to a platform interface class, we need to resolve it immediately.

HakkyuKim: Ah okay, I got it now, thanks for the detailed explanation. I'll remove the "downgrading" code and resolve `package_info_plus` in a separate PR.

Issue #149, opened 2021-07-12
[flutter_tts] Unskip integration test


HakkyuKim:
Integration test for tts is skipped due to reasons discussed in this [PR](https://github.com/flutter-tizen/plugins/pull/147).

PR #152, opened 2021-07-14
[flutter_tts] Always send valid values instead of Error

bbrto21:
* The implementation of this plugin expects to return a valid value such as 1,0.
So I mimicked other platform implementations.

### DIFF
--- a/packages/flutter_tts/CHANGELOG.md
+++ b/packages/flutter_tts/CHANGELOG.md
@@ -5,3 +5,7 @@
 ## 1.0.1
 
 * Stabilize a speaking states management.
+
+## 1.1.0
+
+* All APIs now return valid values.(0 or 1, empty list, etc.)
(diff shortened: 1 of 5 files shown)


bbrto21:
Fix #149

HakkyuKim:
Was somewhat unsure about doing this fix, but if it's already done I'm OK with it! How about leaving a comment about the intentional API misuse in the source file? So that other engineers won't be confused about using `Success(flutter::EncodableValue(0))` to indicate errors.

bbrto21:
@HakkyuKim I added some comments in source file, please review again~

packages/flutter_tts/CHANGELOG.md
```diff
@@ -4,4 +4,9 @@
...
 ## 1.0.1
 
-* Stabilize a speaking states management.
\ No newline at end of file
+* Stabilize a speaking states management.
+
+## 1.1.0
+
+* All APIs now return valid values.(0 or 1, empty list, etc.)
+    * Now the native implementation doesn't throw an error to dart
```
swift-kim, line 12: Please refrain from using nested lists in the CHANGELOG file when you don't need to.

swift-kim (approved):

HakkyuKim (approved):
LGTM

PR #154, opened 2021-07-16
[google_maps_flutter] Add APIs related to camera property

seungsoo47:
* Handle CameraUpdate class for AnimateCamera and MoveCamera API
* Note that AnimateCamera API is not supported in Web. Hence, as in
google_maps_flutter_web, it will run the same as MoveCamera API.

### DIFF
--- a/packages/google_maps_flutter/README.md
+++ b/packages/google_maps_flutter/README.md
@@ -22,10 +22,8 @@ dependencies:
   google_maps_flutter_tizen: ^0.0.1
 ```
 
-In addition, you need a Maps JavaScript API Key to use this plugin. Please refer to the site below to get the API key.
-```
-https://developers.google.com/maps/documentation/javascript/get-api-key
-```
+In addition, you need a Maps JavaScript API Key to use this plugin. Please refer to the site below to get the API key.  
+<https://developers.google.com/maps/documentation/javascript/get-api-key>
 
 To use 'newCameraPosition' API, please add 'v=beta' option in map.html as shown below:
 ```
--- a/packages/google_maps_flutter/README.md
+++ b/packages/google_maps_flutter/README.md
@@ -22,3 +22,14 @@ dependencies:
   google_maps_flutter_tizen: ^0.0.1
 ```
 
+In addition, you need a Maps JavaScript API Key to use this plugin. Please refer to the site below to get the API key.
+```
+https://developers.google.com/maps/documentation/javascript/get-api-key
+```
+
+To use 'newCameraPosition' API, please add 'v=beta' option in map.html as shown below:
+```
+<script async
+    src="https://maps.googleapis.com/maps/api/js?v=beta&key=YOUR_API_KEY&callback=initMap">
+</script>
+```
--- a/packages/google_maps_flutter/lib/src/google_maps_controller.dart
+++ b/packages/google_maps_flutter/lib/src/google_maps_controller.dart
@@ -511,7 +511,8 @@ class GoogleMapController {
           zoomDelta = (json[1] as num) + 0.0;
         }
         // Web only supports integer changes...
-        int newZoomDelta = zoomDelta < 0 ? zoomDelta.floor() : zoomDelta.ceil();
+        final int newZoomDelta =
+            zoomDelta < 0 ? zoomDelta.floor() : zoomDelta.ceil();
         if (json.length == 3) {
           // With focus
           try {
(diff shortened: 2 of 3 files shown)


packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -459,11 +476,101 @@ class GoogleMapController {
...
+        _setPanBy('${json[1]}, ${json[2]}');
+        break;
+      case 'zoomBy':
+        String? focusLatLng;
+        double zoomDelta = 0.0;
+        if (json[1] != null) {
+          zoomDelta = (json[1] as num) + 0.0;
+        }
+        // Web only supports integer changes...
+        int newZoomDelta = zoomDelta < 0 ? zoomDelta.floor() : zoomDelta.ceil();
```
bbrto21, line 514:
```suggestion
        final int newZoomDelta = zoomDelta < 0 ? zoomDelta.floor() : zoomDelta.ceil();
```

packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -459,11 +476,101 @@ class GoogleMapController {
...
-    // assert(_googleMap != null, 'Cannot update the camera of a null map.');
-    // return _applyCameraUpdate(_googleMap!, cameraUpdate);
-
     assert(_widget != null, 'Cannot update the camera of a null map.');
+    _applyCameraUpdate(cameraUpdate);
+  }
+
+  // Translates a [CameraUpdate] into operations on a [Javascript].
+  Future<void> _applyCameraUpdate(CameraUpdate update) async {
+    final List<dynamic> json = update.toJson() as List<dynamic>;
```
bbrto21, line 485: Don't you need a type check or length like the one below?

```dart 
assert(latlng['lat'] is num && latlng['lng'] is num);
```

seungsoo47: Do you mean json[0]'s type?
I am not sure but if it is other type or other value, I think it may jump to default statement.
Should I add an assert statement?

packages/google_maps_flutter/lib/src/google_maps_controller.dart
seungsoo47: Done

packages/google_maps_flutter/lib/src/google_maps_controller.dart
bbrto21: Whatever it is, I don't care.  I've seen you usually put assertions in the parts you're currently implementing.
So I just wanted these to be consistent with other implementations.

seungsoo47: I think it will be able to handled by default statement. Hence, I think it is not necessary here.

bbrto21 (approved):

packages/google_maps_flutter/README.md
```diff
@@ -22,3 +22,14 @@ dependencies:
   google_maps_flutter_tizen: ^0.0.1
 ```
 
+In addition, you need a Maps JavaScript API Key to use this plugin. Please refer to the site below to get the API key.
+```
+https://developers.google.com/maps/documentation/javascript/get-api-key
```
rwalczyna, line 27: Consider putting this as a link. Currently developer has to select text, copy link and manually paste it into address bar. Clicking into link would be easier.

seungsoo47: Thank you~

packages/google_maps_flutter/lib/src/google_maps_controller.dart
```diff
@@ -351,11 +350,28 @@ class GoogleMapController {
...
+    await _callMethod(await controller, 'panBy', [options]);
+  }
+
+  Future<void> _setFitBounds(String options) async {
+    await _callMethod(await controller, 'fitBounds', [options]);
   }
 
   Future<String> _callMethod(
       WebViewController mapView, String method, List<Object?> args) async {
+    print('callMethod: $method($args)');
```
bwikbs, line 374: How about removing these logs? (I mean ``print`` ) 
Even if it is necessary during development, I would like to remove it in the end. I couldn't find such a log in other plugins.

seungsoo47: You are right. It will be removed before it is merged with master branch.

bwikbs (approved):
Thanks for your hard working!

PR #156, opened 2021-07-20
[webview_flutter] Clean up debug code & Organize source

bwikbs:
This is a simple code cleanup to better adapt upstream.


packages/webview_flutter/lib/webview_flutter_tizen.dart
```diff
@@ -36,6 +39,219 @@ enum _PlatformViewState {
...
+
+  @override
+  Widget build({
+    required BuildContext context,
+    required CreationParams creationParams,
+    WebViewPlatformCreatedCallback? onWebViewPlatformCreated,
+    Set<Factory<OneSequenceGestureRecognizer>>? gestureRecognizers,
+    required WebViewPlatformCallbacksHandler webViewPlatformCallbacksHandler,
+  }) {
+    assert(webViewPlatformCallbacksHandler != null);
```
bbrto21, line 73: vscode says this is always true because it can't be null.

packages/webview_flutter/lib/webview_flutter_tizen.dart
```diff
@@ -36,6 +39,219 @@ enum _PlatformViewState {
...
+    Key? key,
+    required this.viewType,
+    this.onPlatformViewCreated,
+    this.hitTestBehavior = PlatformViewHitTestBehavior.opaque,
+    this.layoutDirection,
+    this.gestureRecognizers,
+    this.creationParams,
+    this.creationParamsCodec,
+  })  : assert(viewType != null),
+        assert(hitTestBehavior != null),
```
bbrto21, line 110: dart(unnecessary_null_comparison)

packages/webview_flutter/lib/webview_flutter_tizen.dart
```diff
@@ -792,18 +822,14 @@ mixin _PlatformViewGestureMixin on RenderBox implements MouseTrackerAnnotation {
   set hitTestBehavior(PlatformViewHitTestBehavior value) {
     if (value != _hitTestBehavior) {
       _hitTestBehavior = value;
-      if (owner != null) {
-        markNeedsPaint();
-      }
+      if (owner != null) markNeedsPaint();
```
bbrto21, line 825:
```suggestion
      if (owner != null) { 
        markNeedsPaint();
      }
```

bbrto21:
I'd appreciate it if you could share the upstream commit you referenced.

swift-kim:
Did you just change the code order without actual changes?

I think you can split the code into a couple of files if they are from different files.

FYI) Here is the class hierarchy I analyzed before: https://github.com/flutter-tizen/plugins/pull/54#pullrequestreview-625891709

packages/webview_flutter/lib/webview_flutter_tizen.dart
bwikbs: Most codes come from the original code.(include this part!)

packages/webview_flutter/lib/webview_flutter_tizen.dart
bwikbs: Umm.. Is it a same context like above comment?

packages/webview_flutter/lib/webview_flutter_tizen.dart
bwikbs: Well ... This is also from original code.

bwikbs:
> I'd appreciate if you could share the upstream commit you referenced.

Here is Information what you requested

- 'packages/flutter/lib/src/rendering/platform_view.dart' github.com:flutter/flutter.git@02c026b03cd31dd3f867e5faeb7e104cce174c5f
- github.com:flutter/plugins.git@036fb33ae18bd6a1d2af29302f2124132ca10768

bwikbs:
> Did you just change the code order without actual changes?
> 
> I think you can split the code into a couple of files if they are from different files.
> 
> FYI) Here is the class hierarchy I analyzed before: [#54 (review)](https://github.com/flutter-tizen/plugins/pull/54#pullrequestreview-625891709)

I guess It's a good way. I will separate it with its original file.

packages/webview_flutter/lib/webview_flutter_tizen.dart
bbrto21: This is not just a copy. You're defining a new class with inheritance like below, so you don't have to keep the exact same thing.

Does this make it difficult to sync with upstream code future?

```dart
class TizenWebView implements WebViewPlatform
```

packages/webview_flutter/lib/webview_flutter_tizen.dart
bbrto21: yes, it is the same, I'll close this and keep only one discussion.

packages/webview_flutter/lib/webview_flutter_tizen.dart
swift-kim: @bwikbs Are you sure? :P

https://github.com/flutter/flutter/blob/02c026b03cd31dd3f867e5faeb7e104cce174c5f/packages/flutter/lib/src/rendering/platform_view.dart#L696-L698

swift-kim: Oh, that curly brace is not of the if statement. My mistake.

packages/webview_flutter/lib/webview_flutter_tizen.dart
bwikbs: In fact, All I need is a very small difference branch and I want most of the logic and code to be identical to that of original.
Of course, the reason for doing this is to apply upstream easily.

I understand your intentions, but  I don't make any contribution to platform view logic and I want to make this clear & reveal it in code for now. (This is my intention..if this is strange, I am willing to change it at any time!)

packages/webview_flutter/lib/webview_flutter_tizen.dart
bwikbs: It's a slightly different story, but the change in the number of line in a sentence is the result of  ``vscode``. 
I'm wondering what settings of ``vscode`` will not change original part.

bwikbs:
I tried to divide by file.. but I thought it was too small, so I just divided the ``modified`` parts and the ``non-modified`` parts.

swift-kim (approved):
Thank you!

packages/webview_flutter/lib/platform_view.dart
```diff
@@ -0,0 +1,243 @@
...
+// found in the LICENSE file.
+
+// github.com:flutter/flutter.git@02c026b03cd31dd3f867e5faeb7e104cce174c5f
+// packages/flutter/lib/src/rendering/platform_view.dart
+// Imported from above file, the content has not been modified.
+
+part of 'webview_flutter_tizen.dart';
+
+/// How an embedded platform view behave during hit tests.
+enum PlatformViewHitTestBehavior {
```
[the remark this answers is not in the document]
bwikbs: Good point! Thanks!! :+1:

packages/webview_flutter/lib/platform_view_tizen.dart
```diff
@@ -0,0 +1,615 @@
...
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// github.com:flutter/flutter.git@02c026b03cd31dd3f867e5faeb7e104cce174c5f
+// packages/flutter/lib/src/rendering/platform_view.dart
+// packages/flutter/lib/src/widgets/platform_view.dart
+// packages/flutter/lib/src/services/platform_views.dart
+// Imported from above files, the content has been minimally modified.
+// (e.g. Rename keyword 'Android' -> 'Tizen' )
```
bbrto21, line 11: The context for this implementation is now better understood.

bbrto21 (approved):

bbrto21:
one more things, how about moving implementation code into src dir
see : https://dart.dev/tools/pub/package-layout#implementation-files

bwikbs:
@bbrto21 You're right! thanks!!! :clap:

PR #158, opened 2021-07-20
[video_player] Update integration_test

XuWanchaoS:
- Update integration_test
- Update CHANGELOG.md and README.md
- Set paused after video completed.

PR #159, opened 2021-07-22
[path_provider] Fix integration test errors and update dependencies 

HakkyuKim:
Changes:
- ~~add missing `permission_handler_tizen` dependency in example~~
- ~~request `mediaLibrary` permission and not `accessMediaLocation` to access Tizen media directories~~
- remove `permission_handler` dependency
- remove test code that depend on `permission_handler`
- migrate test code to null safety
- bump up versions

### DIFF
--- a/packages/path_provider/README.md
+++ b/packages/path_provider/README.md
@@ -8,8 +8,8 @@ This package is not an _endorsed_ implementation of `path_provider`. Therefore, 
 
 ```yaml
 dependencies:
-  path_provider: ^2.0.1
-  path_provider_tizen: ^2.0.0
+  path_provider: ^2.0.2
+  path_provider_tizen: ^2.0.1
 ```
 
 Then you can import `path_provider` in your Dart code:
--- a/packages/path_provider/example/integration_test/path_provider_test.dart
+++ b/packages/path_provider/example/integration_test/path_provider_test.dart
@@ -2,10 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
-
-import 'dart:async';
-
 import 'dart:io';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:path_provider/path_provider.dart';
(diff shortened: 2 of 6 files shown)


packages/path_provider/example/pubspec.yaml
```diff
@@ -18,7 +18,9 @@ dev_dependencies:
     sdk: flutter
   integration_test_tizen:
     path: ../../integration_test/
-  permission_handler: ^6.1.1
+  permission_handler: ^8.1.3
+  permission_handler_tizen:
+    path: ../../permission_handler
```
swift-kim, line 23: The example app will fail to run on TV devices if this dependency is added by default.

```
E/ConsoleMessage( 5302): FlutterApplication.cs: Run(83) > Unhandled exception: System.DllNotFoundException: Unable to load shared library 'flutter_plugins.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libflutter_plugins.so: cannot open shared object file: No such file or directory
E/ConsoleMessage( 5302):    at Runner.GeneratedPluginRegistrant.IntegrationTestPluginRegisterWithRegistrar(FlutterDesktopPluginRegistrar registrar)
E/ConsoleMessage( 5302):    at Runner.GeneratedPluginRegistrant.RegisterPlugins(IPluginRegistry registry) in /home/swift/Git/plugins/packages/path_provider/example/tizen/flutter/GeneratedPluginRegistrant.cs:line 22
E/ConsoleMessage( 5302):    at Runner.App.OnCreate() in /home/swift/Git/plugins/packages/path_provider/example/tizen/App.cs:line 11
E/ConsoleMessage( 5302):    at Tizen.Applications.CoreBackend.UICoreBackend.OnCreateNative(IntPtr data)
E/ConsoleMessage( 5302):    at Tizen.Applications.CoreBackend.UICoreBackend.Run(String[] args)
E/ConsoleMessage( 5302):    at Tizen.Applications.CoreApplication.Run(String[] args)
```

packages/path_provider/example/integration_test/path_provider_test.dart
```diff
@@ -2,10 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
-
```
swift-kim, line 6: Hmm. As of Flutter 2.2.1, integration_test has not been migrated to null safety yet. You are supposed to do this migration after the next Flutter stable release, although the code will just run okay with null-safety and only warnings will be generated in the IDE.

HakkyuKim: I've switched to tag 2.2.1 in flutter branch and looked into the integration_test package, is it perhaps not the right way to check?

May also apply to #158?

packages/path_provider/example/pubspec.yaml
HakkyuKim: Looks very familiar...I'll will look into this.

packages/path_provider/example/integration_test/path_provider_test.dart
swift-kim: Ah, you're right. It's already been migrated. I was thinking of Flutter 2.0.1 but not 2.2.1.

https://github.com/flutter/flutter/blob/2.0.1/packages/integration_test/pubspec.yaml
https://github.com/flutter/flutter/blob/2.2.1/packages/integration_test/pubspec.yaml

packages/path_provider/example/pubspec.yaml
swift-kim: Wasn't it unclear for you? TV apps shouldn't depend on permission_handler_tizen because libprivacy_privilege_manager.so is not available on them.

HakkyuKim: Oh, I got it. Overlooked the limitation part.
Since the example tests `path_provider`, I'll skip the test cases that require `permission_handler` and remove its dependencies.

packages/path_provider/README.md
```diff
@@ -8,8 +8,8 @@ This package is not an _endorsed_ implementation of `path_provider`. Therefore,
 
 ```yaml
 dependencies:
-  path_provider: ^2.0.1
-  path_provider_tizen: ^2.0.0
+  path_provider: ^2.0.2
+  path_provider_tizen: ^2.0.1
```
bbrto21, line 12: This patch only fixes test-related code. nevertheless, do we have to up version?

HakkyuKim: As far as I understand, a version specifies a fixed state of every content of a package. I remember reading it somewhere in one of the dart or flutter docs, I'll try to link the doc if I find it.

But if you think about it, from a test runner's perspective, the code has changed how it runs and therefore the output will be different to the runner. It may not be worth releasing it, but we haven't used build number versioning in any of our packages so I didn't want to create an exception.

bbrto21 (approved):
LGTM

PR #162, opened 2021-07-23
[url_launcher] Use get_error_message() for error message conversion

swift-kim:
- Use `get_error_message()` for error message conversion.
- Also minor changes as indicated in the CHANGELOG file. The minimum Dart SDK version has been increased to 2.13.0 to use the new type alias feature.

### DIFF
--- a/packages/url_launcher/lib/src/app_control.dart
+++ b/packages/url_launcher/lib/src/app_control.dart
@@ -76,13 +76,19 @@ class AppControl {
   /// Corresponds to `app_control_set_operation()`.
   void setOperation(String operation) {
     assert(_handle != nullptr);
-    _throwOnError(_appControlSetOperation(_handle, operation.toNativeUtf8()));
+    final Pointer<Utf8> operationNative = operation.toNativeUtf8();
+    final int ret = _appControlSetOperation(_handle, operationNative);
+    malloc.free(operationNative);
+    _throwOnError(ret);
   }
 
   /// Corresponds to `app_control_set_uri()`.
   void setUri(String uri) {
     assert(_handle != nullptr);
-    _throwOnError(_appControlSetUri(_handle, uri.toNativeUtf8()));
+    final Pointer<Utf8> uriNative = uri.toNativeUtf8();
+    final int ret = _appControlSetUri(_handle, uriNative);
+    malloc.free(uriNative);
+    _throwOnError(ret);
   }
 
   /// Corresponds to `app_control_send_launch_request()`.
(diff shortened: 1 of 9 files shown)


packages/url_launcher/lib/src/app_control.dart
```diff
@@ -2,158 +2,100 @@
...
     assert(_handle != nullptr);
-
-    final int ret = _setOperation(_handle, operation.toNativeUtf8());
-    if (ret != 0) {
-      throw PlatformException(
-        code: _getErrorCode(ret),
-        message: 'Failed to execute app_control_set_operation.',
-      );
-    }
+    _throwOnError(_appControlSetOperation(_handle, operation.toNativeUtf8()));
```
WonyoungChoi, line 79: `toNativeUtf8()` method allocates new memory space, it should be freed. 
see: 
- https://github.com/dart-lang/ffi/blob/4dd32429880a57b64edaf54c9d5af8a9fa9a4ffb/lib/src/utf8.dart#L83
- https://pub.dev/packages/ffi/example

WonyoungChoi, line 79: The `Arena` of `package:ffi` can be help.
```suggestion
   using((Arena arena) {
      _throwOnError(_appControlSetOperation(_handle, operation.toNativeUtf8(arena)));
   });
```

swift-kim: Good point. Thanks!

swift-kim: I think explicitly deallocating the memory looks more straightforward in this case because there's only one allocation.

swift-kim: Ah, using `Arena` makes the code much cleaner. I'll apply this suggestion. Thanks!

packages/url_launcher/lib/src/app_control.dart
```diff
@@ -2,158 +2,104 @@
...
-
-    final int ret = _setOperation(_handle, operation.toNativeUtf8());
-    if (ret != 0) {
-      throw PlatformException(
-        code: _getErrorCode(ret),
-        message: 'Failed to execute app_control_set_operation.',
-      );
-    }
+    using((Arena arena) {
+      _throwOnError(_appControlSetOperation(_handle, operation.toNativeUtf8()));
```
WonyoungChoi, line 80:
```suggestion
      _throwOnError(_appControlSetOperation(_handle, operation.toNativeUtf8(arena)));
```
Maybe you should pass `arena` as the allocator to `toNativeUtf8()` method. I am not sure. please check one more.

swift-kim:
The CI is still dumb.

bbrto21 (approved):

Issue #161, opened 2021-07-23
Update analysis_options.yaml


swift-kim:
- 1st-party plugin implementations should use linter rules from https://github.com/flutter/plugins/blob/master/analysis_options.yaml.
- Some 1st-party plugins still depend on legacy options: https://github.com/flutter/plugins/blob/master/analysis_options_legacy.yaml
- 3rd-party plugins use their own linter rules. Copy analysis_options files from their repos.

bbrto21:
[analyze.txt](https://github.com/flutter-tizen/plugins/files/7089491/analyze.txt)

PR #166, opened 2021-08-03
[shared_preferences] Fix memory leaks

WonyoungChoi:
Fix memory leaks
- Use Arena to release memory allocated in toNativeUtf8() method.
- Use Arena instead of `malloc()` to release allocated memory automatically. 
- Free `pString` after use it. 
- Fix some styles.

### DIFF
--- a/packages/shared_preferences/example/integration_test/shared_preferences_test.dart
+++ b/packages/shared_preferences/example/integration_test/shared_preferences_test.dart
@@ -101,13 +101,13 @@ void main() {
 
     testWidgets('simultaneous writes', (WidgetTester _) async {
       final List<Future<bool>> writes = <Future<bool>>[];
-      final int writeCount = 100;
+      const int writeCount = 100;
       for (int i = 1; i <= writeCount; i++) {
         writes.add(preferences.setInt('int', i));
       }
-      List<bool> result = await Future.wait(writes, eagerError: true);
+      final List<bool> result = await Future.wait(writes, eagerError: true);
       // All writes should succeed.
-      expect(result.where((element) => !element), isEmpty);
+      expect(result.where((bool element) => !element), isEmpty);
       // The last write should win.
       expect(preferences.getInt('int'), writeCount);
     });
--- a/packages/shared_preferences/lib/shared_preferences_tizen.dart
+++ b/packages/shared_preferences/lib/shared_preferences_tizen.dart
@@ -102,40 +102,36 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
 
   @override
   Future<bool> setValue(String valueType, String key, Object value) async {
-    _preferences[key] = value;
-
-    return using((Arena arena) {
-      int ret;
+    final int ret = using((Arena arena) {
       final Pointer<Utf8> pKey = key.toNativeUtf8(allocator: arena);
       switch (valueType) {
         case 'Bool':
-          ret = bindings.setBoolean(pKey, (value as bool) ? 1 : 0);
-          break;
+          return bindings.setBoolean(pKey, (value as bool) ? 1 : 0);
         case 'Double':
-          ret = bindings.setDouble(pKey, value as double);
-          break;
+          return bindings.setDouble(pKey, value as double);
         case 'Int':
-          ret = bindings.setInt(pKey, value as int);
-          break;
+          return bindings.setInt(pKey, value as int);
         case 'String':
-          ret = bindings.setString(
+          return bindings.setString(
             pKey,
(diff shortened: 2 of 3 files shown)


packages/shared_preferences/lib/shared_preferences_tizen.dart
```diff
@@ -26,51 +26,49 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
...
-    }
+    using((Arena arena) {
+      // Try as boolean type
+      final Pointer<Int8> pBool = arena();
+      ret = bindings.getBoolean(pKey, pBool);
+      final bool boolValue = pBool.value == 1;
+      if (ret == 0) {
+        _cachedPreferences![key] = boolValue;
+        return;
+      }
```
swift-kim, line 37: If there's no need to call `malloc.free()`, the code can be simplified as

```suggestion
      if (bindings.getBoolean(pKey, pBool) == 0) {
        _cachedPreferences![key] = pBool.value == 1;
        return;
      }
```

packages/shared_preferences/lib/shared_preferences_tizen.dart
```diff
@@ -101,39 +99,56 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
...
-        joined = _separator;
-      } else {
-        joined = _separator + list.join(_separator) + _separator;
+    return using((Arena arena) {
+      int ret;
+      final Pointer<Utf8> pKey = key.toNativeUtf8(allocator: arena);
+      switch (valueType) {
+        case 'Bool':
+          ret = bindings.setBoolean(pKey, (value as bool) ? 1 : 0);
+          break;
```
swift-kim, line 121: The more code fitting in one screen, the better readability.
```suggestion
        case 'Bool':
          return bindings.setBoolean(pKey, (value as bool) ? 1 : 0);
```
(`== 0` should be added at line 146.)

packages/shared_preferences/lib/shared_preferences_tizen.dart
```diff
@@ -26,51 +26,49 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
...
-    final Pointer<Int8> pBool = malloc();
-    ret = bindings.getBoolean(pKey, pBool);
-    final bool boolValue = pBool.value == 1;
-    malloc.free(pBool);
-    if (ret == 0) {
-      _cachedPreferences![key] = boolValue;
-      return 1;
-    }
+    using((Arena arena) {
+      // Try as boolean type
```
swift-kim, line 30: Please either remove this or add `.` at the end.

WonyoungChoi: my mistake.. :(

packages/shared_preferences/lib/shared_preferences_tizen.dart
WonyoungChoi: good point! :)

swift-kim (approved):

packages/shared_preferences/example/integration_test/shared_preferences_test.dart
```diff
@@ -101,13 +101,13 @@ void main() {
 
     testWidgets('simultaneous writes', (WidgetTester _) async {
       final List<Future<bool>> writes = <Future<bool>>[];
-      final int writeCount = 100;
```
bwikbs, line 104: Shouldn't it be similar to the original if possible? Is this fragmentation worth the risk?
(I'm just curious what other people think regrading this topic)

swift-kim: I personally don't care because replacing `analysis_options.yaml` is going on in the flutter/plugins repo (https://github.com/flutter-tizen/plugins/issues/161) and this integration test file will be updated soon when the migration is done.

bwikbs (approved):
I still have a personal favorite wish that  we keep the original part as much as we can, but It's not important and everything else looks fine! thanks!! :+1:

bbrto21:
@WonyoungChoi please, share a guide to build and run this plugin after this PR

swift-kim:
Has it changed?

WonyoungChoi:
@bbrto21 There are no changes that affect how you use them.  Any problem to build and run do you have?

bbrto21:
It works fine after ```flutter-tizen clean```. please don't mind :smile:

bbrto21:
@WonyoungChoi Could you please update a version so that release this plugin to pub.dev?

WonyoungChoi:
@bbrto21 I made a PR for updating version. #172

PR #173, opened 2021-08-10
[geolocator] Add location and permission manager

bbrto21:
* Handle a checkPermission method
* Handle a isLocationServiceEnabled method
* Note: These are basic implementations of manager


bwikbs (approved):
Consider the introduction of a permission package.

PR #175, opened 2021-08-13
Change project types to staticLib

swift-kim:
- Change project types and remove linker-specific properties according to the tool change in https://github.com/flutter-tizen/flutter-tizen/pull/195. (A deprecation warning will be generated if the project type is not updated.)
- Change python scripts file modes to 755.


swift-kim:
The CI is broken :(

packages/messageport/tizen/project_def.prop
```diff
@@ -2,31 +2,21 @@
 # for details.
 
 APPNAME = messageport_tizen_plugin
-type = sharedLib
-profile = common-5.5
+type = staticLib
+profile = common-4.0
```
swift-kim, line 6: This change is intentional.

HakkyuKim:
The problem is caused by: https://github.com/flutter-tizen/plugins/blob/92364110fd5d0bb16ab3d1fb803b9c24f403d865/.github/workflows/build.yml#L16-L18

which should have been removed. I'll fix it.

WonyoungChoi:
it should be rebased for checks green.

HakkyuKim:
@swift-kim 
The #176 just merged to master, it will work after rebase :)

HakkyuKim (approved):

PR #179, opened 2021-08-17
[webview_flutter] Update supported devices

swift-kim:
The webview plugin is not supported on watch devices because

- The implementation doesn't support the Evas GL backend.
- Non-public APIs are used in `liblightweight-web-engine.flutter.so`.


bwikbs (approved):
Thanks!!! :+1:

bbrto21 (approved):

PR #182, opened 2021-08-18
[permission_handler] Allow emulators to pass integration test

HakkyuKim:
Opening app settings to view permission status is only allowed in real targets. Specifically, the dependent app `com.samsung.clocksetting.apps` in emulators don't have the required feature.


bwikbs (approved):
👍

bbrto21 (approved):

bbrto21:
I realized that there is a way to avoid asking permission on emulator
before run test do the below

```shell
touch /opt/askuser_disable
```

PR #183, opened 2021-08-18
[google_maps_flutter] Implement initial marker module

seungsoo47:
* Add markersController and markerController class
* Add properties and methods related to marker for Google Map
* Add properties and methods related to infoWindow for Google Map
* Fix the widget to be launched only once

### DIFF
--- a/packages/google_maps_flutter/lib/src/convert.dart
+++ b/packages/google_maps_flutter/lib/src/convert.dart
@@ -161,7 +161,7 @@ util.GInfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
   if (markerSnippet.isNotEmpty) {
     buffer.write('<div class="infowindow-snippet">');
     buffer.write(markerSnippet);
-    buffer.write('</dvi>');
+    buffer.write('</div>');
   }
   buffer.write('</div>\'');
 
--- a/packages/google_maps_flutter/lib/src/convert.dart
+++ b/packages/google_maps_flutter/lib/src/convert.dart
@@ -141,8 +141,8 @@ LatLng _convertToLatLng(String value) {
 }
 
 util.GInfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
-  final markerTitle = marker.infoWindow.title ?? '';
-  final markerSnippet = marker.infoWindow.snippet ?? '';
+  final String markerTitle = marker.infoWindow.title ?? '';
+  final String markerSnippet = marker.infoWindow.snippet ?? '';
 
   // If both the title and snippet of an infowindow are empty, we don't really
   // want an infowindow...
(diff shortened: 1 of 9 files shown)


bwikbs (approved):
Good work 👍 
I hope that this approach will be a breakthrough for our plugin development.

packages/google_maps_flutter/lib/src/convert.dart
```diff
@@ -112,3 +108,120 @@ bool _isTrafficLayerEnabled(Map<String, dynamic> rawOptions) {
...
+  buffer.write('\'<div id="marker-${marker.markerId.value}-infowindow">');
+  if (markerTitle.isNotEmpty) {
+    buffer.write('<h3 class="infowindow-title">');
+    buffer.write(markerTitle);
+    buffer.write('</h3>');
+  }
+  if (markerSnippet.isNotEmpty) {
+    buffer.write('<div class="infowindow-snippet">');
+    buffer.write(markerSnippet);
+    buffer.write('</dvi>');
```
swift-kim, line 164: Hmm?

bwikbs: 👀 Wow!

packages/google_maps_flutter/lib/src/convert.dart
```diff
@@ -4,10 +4,6 @@
 
 part of google_maps_flutter_tizen;
```
swift-kim, line 5: How do you think of this instruction?

> Note: You may have heard of the `part` directive, which allows you to split a library into multiple Dart files. We recommend that you avoid using `part` and create mini libraries instead.

https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package

The part directive was used in webview_flutter_tizen because we wanted to keep the exact copy of the original code which contained private types/functions.

bwikbs: IMO, It is currently under development, so this may be applied when it is merged into the master.
Of course, it will be convenient in the future if it is organized from now on.

seungsoo47: This plugin is being implemented similar to original plugin structure. Of course, It could be changed in the form of a mini-library, but I think it's probably after the first release.

packages/google_maps_flutter/lib/src/convert.dart
seungsoo47: I'm sorry, but what do you mean?

packages/google_maps_flutter/lib/src/convert.dart
swift-kim: Okay, I see.

packages/google_maps_flutter/lib/src/convert.dart
swift-kim: The typo: `</dvi>`

seungsoo47: Oh. Good. I will fix it. Thanks

PR #188, opened 2021-08-23
[geolocator] Setup channel for geolocator service update

bbrto21:

### DIFF
--- a/packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
+++ b/packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
@@ -4,6 +4,9 @@
 
 #include "geolocator_tizen_plugin.h"
 
+#include <flutter/event_channel.h>
+#include <flutter/event_sink.h>
+#include <flutter/event_stream_handler_functions.h>
 #include <flutter/method_channel.h>
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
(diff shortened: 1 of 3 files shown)


packages/geolocator/tizen/src/locaton_manager.h
```diff
@@ -12,8 +12,14 @@
 #include "location.h"
 #include "tizen_result.h"
 
+// Defined in:
+
+// https://github.com/Baseflow/flutter-geolocator/blob/master/geolocator_platform_interface/lib/src/enums/location_service.dart
```
swift-kim, line 17: Unexpected newline between the comments.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -44,6 +49,27 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
   virtual ~GeolocatorTizenPlugin() {}
 
  private:
+  void SetupGeolocatorServiceUpdatesChannel(
+      flutter::BinaryMessenger *messenger) {
```
bwikbs, line 53: Suddenly curious...
``flutter::BinaryMessenger *messenger``
``flutter::BinaryMessenger* messenger``
Which one are we? doesn't it matter?

bwikbs (approved):
:+1:

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
bbrto21: I think both are fine If you choose one consistently within a single file. and I always turn on auto formatting in vscode.
https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions

swift-kim: The former is preferred by the Google style and the latter is used by the Chromium style. I don't know why. This repo follows the Google style so the former is correct.

bwikbs: That's a tragedy :cry:

packages/geolocator/tizen/src/locaton_manager.h
bbrto21: Thanks!

PR #189, opened 2021-08-24
[google_maps_flutter] Implement getScreenCoordinate and getLatLng API

seungsoo47:
* Get coordinates (like LatLng or Point) information from the map


bwikbs (approved):
Is it almost finish?

seungsoo47:
Yes. I think the initial version is almost finished.

PR #192, opened 2021-08-31
[geolocator] Setup channel for geolocator updates

bbrto21:
* Implement Set/UnsetOnLocationUpdated

### DIFF
--- a/packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
+++ b/packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
@@ -33,6 +33,7 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
     auto plugin = std::make_unique<GeolocatorTizenPlugin>();
 
     plugin->SetupGeolocatorServiceUpdatesChannel(registrar->messenger());
+    plugin->SetupGeolocatorUpdatesChannel(registrar->messenger());
 
     channel->SetMethodCallHandler(
         [plugin_pointer = plugin.get()](const auto &call, auto result) {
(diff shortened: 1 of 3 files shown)


packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -186,7 +205,27 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
 
   void OnCancelGeolocatorServiceUpdates() {
     geolocator_service_updates_event_sink_ = nullptr;
-    TizenResult tizen_result = location_manager_->UnsetOnServiceStateChanged();
+    location_manager_->UnsetOnServiceStateChanged();
+  }
+
+  void OnListenGeolocatorUpdates(
```
bwikbs, line 211: Any special reason to make separately?  Can't we do it all at once?

bbrto21: Well, I'd appreciate it if you could elaborate on your proposal.

bwikbs: never mind. It's wrong position!

bwikbs (approved):
Thanks

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
bwikbs: My first intention is here: 
https://github.com/flutter-tizen/plugins/blob/171542aaed62f0e87df6ae43150b2844cbf2aad8/packages/geolocator/tizen/src/geolocator_tizen_plugin.cc#L74

HakkyuKim (approved):
LGTM

PR #193, opened 2021-09-01
[geolocator] Implement open App/location settings

bbrto21:

### DIFF
--- a/packages/geolocator/tizen/src/setting.cc
+++ b/packages/geolocator/tizen/src/setting.cc
@@ -1,4 +1,3 @@
-
 // Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
@@ -86,7 +85,7 @@ TizenResult Setting::LaunchAppSetting() {
   return app_ctrl.SendLauchRequest();
 }
 
-TizenResult Setting::LaunchLocatoinSetting() {
+TizenResult Setting::LaunchLocationSetting() {
   AppControl app_ctrl;
 
   app_ctrl.SetAppId("com.samsung.setting-location");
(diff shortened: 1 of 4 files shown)


packages/geolocator/tizen/src/setting.cc
```diff
@@ -0,0 +1,97 @@
+
```
bwikbs, line 1: :scissors:  plz~

bwikbs:
I think ``TizenResult``  can be moved to common part.  (for all tizen plugin)

bwikbs (approved):
:+1:

packages/geolocator/tizen/src/setting.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef SETTING_H_
+#define SETTING_H_
+
+#include "tizen_result.h"
+
+namespace Setting {
+
+TizenResult LaunchAppSetting();
+TizenResult LaunchLocatoinSetting();
```
HakkyuKim, line 13: Just one typo.

```suggestion
TizenResult LaunchLocationSetting();
```

packages/geolocator/tizen/src/setting.cc
```diff
@@ -0,0 +1,97 @@
...
+  std::string package_name_;
+};
+
+}  // namespace
+
+TizenResult Setting::LaunchAppSetting() {
+  PackageName name;
+  AppControl app_ctrl;
+
+  app_ctrl.SetAppId("com.samsung.clocksetting.apps");
```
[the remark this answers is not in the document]
bbrto21: Yes. this ID came from the setting app, which is installed by default on the Galaxy Watch series.
and the location APIs is not supported on Smart TV profile

packages/geolocator/tizen/src/setting.h
bbrto21: Thanks!

HakkyuKim (approved):
LGTM

Issue #204, opened 2021-09-07
[webview_flutter] Screen does not display properly when using multiple webviews.


bwikbs:
The screen does not display properly when using multiple webviews. 
You can check it with the following sample:
[Sample](https://gist.github.com/bwikbs/5459b215fdcb55e59f3dd01f88db0159)

Result on TM1
![dump_screen2](https://user-images.githubusercontent.com/4053790/132424563-ed9d7080-e6bc-4581-b41a-cf2def9fc7d2.jpg)

![dump_screen1](https://user-images.githubusercontent.com/4053790/132424576-b4e2d0f1-3bac-42f8-9e2b-2fe4be1ea7e8.jpg)

When only one is being updated, it seems normal, but when several are updated at the same time, the contents are mixed and displayed.

bbrto21:
Doesn't the video or camera plugin have this symptom?

bwikbs:
I haven't tried it, but I don't know if the MM player supports decoding such case.

bwikbs:
To narrow down where the problem is, I used a version that uses lwe's SW backend.
Here is Result.

bwikbs:
There is a misunderstanding on LWE behavior. It's not multi thread issue.
I guess EGL configuration issue.

PR #205, opened 2021-09-08
[webview_flutter] Update surface management logic

bwikbs:
Include 2 things:

- Add surface for synchronization
- Use the memory pool evenly

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -150,6 +150,7 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
       webview_instance_(nullptr),
       width_(width),
       height_(height),
+      working_surface_(nullptr),
       candidate_surface_(nullptr),
       rendered_surface_(nullptr),
       is_mouse_lbutton_down_(false),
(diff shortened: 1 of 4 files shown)


bbrto21:
Is it related to #204?

bwikbs:
> Is it related to #204?

No but It's result of #204 investigation :smile:

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -774,20 +775,27 @@ void WebView::InitWebView() {
...
         if (isRendered) {
-          texture_registrar_->MarkTextureFrameAvailable(GetTextureId());
+          std::lock_guard<std::mutex> lock(mutex_);
+          if (candidate_surface_) {
+            tbm_pool_->Release(candidate_surface_);
+            candidate_surface_ = nullptr;
+          } else {
+            texture_registrar_->MarkTextureFrameAvailable(GetTextureId());
+          }
+          candidate_surface_ = working_surface_;
```
bbrto21, line 797: Q. It looks like `working_surface_` could be `nullptr` because `tbm_pool_->GetAvailableBuffer() `can return `nullptr`. nevertheless, is its safety guaranteed?

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -935,14 +943,14 @@ void WebView::HandleCookieMethodCall(
...
+  std::lock_guard<std::mutex> lock(mutex_);
   if (!candidate_surface_) {
-    if (!rendered_surface_) {
-      return nullptr;
-    } else {
+    if (rendered_surface_) {
       return gpu_buffer_;
+    } else {
+      return nullptr;
     }
```
bbrto21, line 952:
```suggestion
    }
    return nullptr;
```

packages/webview_flutter/tizen/src/webview.cc
bwikbs: No. But I don't assume such a situation If we met that situation It should fail Immediately.
This is intended thing.

packages/webview_flutter/tizen/src/webview.cc
bwikbs: :+1:

bbrto21 (approved):

bbrto21 (approved):

bbrto21:
To release this, please update version

bwikbs:
A release is planned with lwe to fix https://github.com/flutter-tizen/plugins/issues/204 issue.

PR #216, opened 2021-09-13
[tizen_app_control] Add tizen_app_control package

swift-kim:
Contributes to https://github.com/flutter-tizen/flutter-tizen/issues/210.

Depends on https://github.com/flutter-tizen/engine/pull/183 and https://github.com/flutter-tizen/engine/pull/184 (`AZURE_BUILD_ID=565`).

tizen_app_control is now a separate package from the internal tizen_core package. The API is similar to tizen_core's, but the implementation, example, and integration_test have almost been re-written from scratch.

Notable API changes:
- `AppControl.createAppControl` and `AppControl.dispose` have been removed in favor of the new finalizable Dart objects (https://github.com/flutter-tizen/engine/pull/183).
- `Map<String, dynamic>` replaces the type `AppControlExtraData`.
- The `ReceivedAppControl` class has been added. (Reference: [TizenFX's ReceivedAppControl class](https://docs.tizen.org/application/dotnet/api/TizenFX/API8/api/Tizen.Applications.ReceivedAppControl.html))
- `AppManager.terminateBackgroundApplication` and `AppManager.isRunning` have been added to support terminating service applications. (Reference: [TizenFX's ApplicationManager class](https://docs.tizen.org/application/dotnet/api/TizenFX/API8/api/Tizen.Applications.ApplicationManager.html))

After publishing this package, we may:
- Remove the `AppControl` implementation from tizen_core.
- Update the multi app template.
- Re-implement plugins:
  - share_plus_tizen
  - url_launcher_tizen
  - (image_picker_tizen)

Although this package is technically a pure Dart package, I decided to place it in this repo because it heavily depends on the embedder's native implementation.


swift-kim:
@HakkyuKim The CI build failed due to missing certificate profile. Can we add a certificate profile to the CI server?

swift-kim:
It would be important to note that the `await AppControl.sendLaunchRequest` call will not ever return when `replyCallback` is non-null and the callee app exited without replying (e.g. by pressing the back button in the image picker view). I couldn't find a good solution for this. @WonyoungChoi Do you know if it's an expected behavior on Tizen?

HakkyuKim:
> @HakkyuKim The CI build failed due to missing certificate profile. Can we add a certificate profile to the CI server?

#217 should solve the problem.

swift-kim:
@HakkyuKim Oops. I've already done the same thing in e3387a937dd7a3a039c5fa130d214558335435cc. Should I revert?

HakkyuKim:
@swift-kim 
Oh just saw it. I can close mine, let's just merge from this PR.

pkosko (approved):
Code looks very good. Approve from me.

HakkyuKim:
LGTM after solving:

> It would be important to note that the await AppControl.sendLaunchRequest call will not ever return when replyCallback is non-null and the callee app exited without replying.

I also couldn't think of a right solution. If Tizen has no preferred way of handling this, maybe the API should document the fact so users can use `Future.timeout`?

swift-kim:
@HakkyuKim Thanks. I added some comments but didn't specify a preferred way to deal with the situation.

My personal recommendation is to use [`CancelableOperation`](https://pub.dev/documentation/async/latest/async/CancelableOperation-class.html) from the async package with [`WidgetsBindingObserver`](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver/didChangeAppLifecycleState.html) (to listen for app resume events).

HakkyuKim (approved):
@swift-kim
Cool, didn't know that you could wrap a `Future` and cancel it. LGTM.

PR #218, opened 2021-09-13
Add an action to analyze packages

bbrto21:
* Change file name to analyze.yml, more comprehensively

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,4 +1,4 @@
-name: Format
+name: Analyze
 
 on: [push, pull_request]
 
@@ -18,6 +18,10 @@ jobs:
           done
       - name: Verify formatting
         run: flutter format --set-exit-if-changed packages
+      - name: Analyze the plugin's Dart code
+        run: |
+          cd packages
+          flutter analyze
   clang:
     runs-on: ubuntu-latest
     steps:
--- a/tools/commands/build_example.py
+++ b/tools/commands/build_example.py
@@ -12,9 +12,8 @@ def parse_args(args):
     return parser.parse_args(args)
 
 
-def _build_examples(plugin):
-    name = os.path.basename(plugin)
-    example_dir = os.path.join(plugin, 'example')
+def _build_examples(plugin_dir):
+    example_dir = os.path.join(plugin_dir, 'example')
     subprocess.run('flutter-tizen pub get', shell=True, cwd=example_dir)
 
     completed_process = subprocess.run('flutter-tizen build tpk --device-profile wearable -v',


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,28 @@
...
+              echo "HAS_CHANGED_PACKAGES=true" >> "$GITHUB_ENV"
+          fi
+      - name: Install flutter-tizen
+        if: ${{ env.HAS_CHANGED_PACKAGES == 'true' }}
+        run: |
+          git clone https://github.com/flutter-tizen/flutter-tizen.git
+      - name: Analyze changed packages
+        if: ${{ env.HAS_CHANGED_PACKAGES == 'true' }}
+        run: |
+          ROOT_DIR=`pwd`
```
WonyoungChoi, line 26: Where is `ROOT_DIR` used?

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,28 @@
...
+          export PATH=$PATH:`pwd`/flutter-tizen/bin
```
WonyoungChoi, line 27:
```suggestion
          echo "$GITHUB_WORKSPACE/flutter-tizen/bin" >> $GITHUB_PATH
```
You can use environment files instead. 
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files

tools/run_command.py
```diff
@@ -6,9 +6,11 @@
 from commands import integration_test
 from commands import build_example
 from commands import print_plugins
-
+from commands import analyze_plugins
 
 # Check tidy
+
+
```
HakkyuKim, line 13: Blanks lines.

tools/run_command.py
```diff
@@ -27,12 +29,19 @@ def run_build_examples(argv):
 def run_print_plugins(arv):
     print_plugins.run_print_plugins(arv)
 
+# Analyze plugin packages
+
+
```
HakkyuKim, line 34: Same here.

tools/run_command.py
```diff
@@ -27,12 +29,19 @@ def run_build_examples(argv):
...
+def run_analyze_plugins(arv):
+    analyze_plugins.run_analyze_plugins(arv)
+
 
 commands = {}
 commands['tidy'] = {'func': run_check_tidy, 'info': 'Check and update format for C++ files'}
 commands['test'] = {'func': run_integration_test, 'info': 'Run integration test'}
 commands['build'] = {'func': run_build_examples, 'info': 'Build examples of plugin'}
 commands['plugins'] = {'func': run_print_plugins, 'info': 'Print plugins list'}
+commands['analyze'] = {'func': run_analyze_plugins, 'info': 'Analyze plugins for Dart files'}
```
HakkyuKim, line 44: How about "Analyze the plugin's Dart code" for consistency with `flutter-tizen analyze`?

```suggestion
commands['analyze'] = {'func': run_analyze_plugins, 'info': 'Analyze the plugin's Dart code'}
```

tools/commands/analyze_plugins.py
```diff
@@ -0,0 +1,38 @@
...
+import commands.command_utils as command_utils
+
+
+def parse_args(args):
+    parser = command_utils.get_options_parser(
+        plugins=True, exclude=True, run_on_changed_packages=True, base_sha=True, command='analyze')
+    return parser.parse_args(args)
+
+
+def _run_analyze_plugins(plugin):
```
HakkyuKim, line 15: How about:
```suggestion
def _run_analyze_plugins(plugin_dir):
```


`plugin` is generally used for the name of the plugin, for example line 31.
I would appreciate if you also change `_build_examples(plugin)` to `_build_examples(plugin_dir)` in `_build_example.py`.

swift-kim:
We can use `dart analyze` or `flutter analyze` for static analysis without having to download flutter-tizen.

e.g. https://github.com/marketplace/actions/setup-dart-sdk#check-static-analysis-formatting-and-test-example

There are some caveats though. Let me discuss with you offline.

bbrto21:
> We can use `dart analyze` or `flutter analyze` for static analysis without having to download flutter-tizen.
> 
> e.g. https://github.com/marketplace/actions/setup-dart-sdk#check-static-analysis-formatting-and-test-example
> 
> There are some caveats though. Let me discuss with you offline.

After this conversation, we decided to make some changes.

.github/workflows/analyze.yml
bbrto21: Thank you for your review. I pushed a new commit which is simpler than before

tools/run_command.py
bbrto21: It was my mistake, I cleaned them.

tools/run_command.py
bbrto21: I applied this, thank you.

tools/commands/analyze_plugins.py
bbrto21: please, check again, I applied it to all the places you mentioned.

.github/workflows/format.yml
```diff
@@ -7,6 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
swift-kim, line 11: Why 2 but not more?

.github/workflows/format.yml
```diff
@@ -7,6 +7,8 @@ jobs:
...
       - uses: subosito/flutter-action@v1
         with:
           flutter-version: "2.2.1"
```
swift-kim, line 14: We concluded to always use the latest stable version. If you don't need any feature from `flutter analyze`, you can just use  [`dart analyze`](https://github.com/dart-lang/setup-dart).

.github/workflows/format.yml
```diff
@@ -18,6 +20,8 @@ jobs:
           done
       - name: Verify formatting
         run: flutter format --set-exit-if-changed packages
+      - name: Analyze the plugin's Dart code
+        run: tools/run_command.py analyze --run-on-changed-packages --base-sha $(git rev-parse HEAD^)
```
swift-kim, line 24: The format step runs for all packages while the analyze step only runs for changed packages. Why different.

Analyzing packages doesn't take much time (unlike the build job). We may just run `dart analyze` for all packages for simplicity.

.github/workflows/format.yml
bbrto21: I need to use `HEAD^`, so I followed this guide https://github.com/actions/checkout#checkout-head

.github/workflows/format.yml
bbrto21: > The format step runs for all packages while the analyze step only runs for changed packages. Why different.

OK. I basically preferred to analyze only the packages that have changed, but your opinion is also good. I will change this.

>  We may just run dart analyze for all packages for simplicity.

With Dart, more documentation is needed for each dart file's public API. (ref [this](https://stackoverflow.com/questions/65308789/why-is-flutter-analyze-different-from-dart-analyze))

If you don't mind we can change to `dart` from `flutter` later. otherwise I think we have to do more documentation about API before this PR

swift-kim (approved):

HakkyuKim (approved):

PR #219, opened 2021-09-13
Refactor tools to use subparsers

HakkyuKim:
Changes:
- Refactor `run_command.py` and `commands/*.py` to use subparsers for better error handling.
- Make `test_target` an optional parameter to resolve "missing required positional argument" error in `integration_test.py`.
- Use consistent error message in `integration_test.py`.

The last two commits(d470c84f760f5304807f9d5f0658834a35a002c1, 6662780d72eb2d50a9c20360ae02fdc98d4e99eb) are code clean ups. The major changes are in 0c52a49583d7d4abdf627d7a019905f19e1d035a.

### DIFF
--- a/tools/run_command.py
+++ b/tools/run_command.py
@@ -47,3 +47,4 @@ if __name__ == "__main__":
         commands[sys.argv[1]]['func'](sys.argv[2:])
     except Exception as e:
         print_usage()
+        exit(1)
--- a/tools/run_command.py
+++ b/tools/run_command.py
@@ -7,7 +7,6 @@ from commands import check_tidy
 from commands import integration_test
 from commands import build_example
 from commands import print_plugins
-from commands import command_utils
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


tools/run_command.py
```diff
@@ -47,3 +47,4 @@ def print_usage():
         commands[sys.argv[1]]['func'](sys.argv[2:])
     except Exception as e:
         print_usage()
+        exit(1)
```
bbrto21, line 50: To provide more information and to classify the error situation, how about this?
```python
    except KeyError:
        print_usage()
    except Exception as e:
        print(e)
        exit(1)
```

HakkyuKim: The tool prints misuses of commands as programmatic error, such as "list index out of range" for running `./tools/run_command.py`. I believe it can be solved by using `add_subparsers` in `argparse`, I'll update the PR after finishing #220.

HakkyuKim: @bbrto21 
I've refactored the code to use subparsers, please review the changes when possible.

bbrto21 (approved):
LGTM

PR #229, opened 2021-09-28
Refactor, run clean command after each test, improve logic for setting default targets.

HakkyuKim:
Sorry for the unexpected huge diff, adding default targets was not as straightforward as I originally thought. I may have done some over refactoring but I already had planned the changes for #230. It was simpler to work with the refactored code so I decided to apply it a bit early.

This PR must be merged after adding the `--platforms` option in `integration_test.yml`.

Summary:
- Run clean command after each test to save disk space. ~~(Nothing has changed between line 136~237)~~
- Don't catch exceptions and print stack trace.
- Fix incorrectly used nested f-string.
- Introduce `Target` which encapsulates a Tizen device.
- Introduce `TargetManager` which finds and manages a collection of `Target`s.
- Update outdated docstrings.
- Define some words used in code, such as `target`, `platform`, etc.
- Allow tool to select all connected targets for test devices by default.
- Add `--platforms` option which takes inputs of the form `<device_profile>-<tizen_version>`, the tool will find all satisfying targets and serve them as test devices.

```bash
# When connected with 4 emulators with different platforms.
./tools/run_command.py test --plugins wakelock
SUCCEEDED: wakelock ('mobile-6.0', 'M-6.0-x86', 'emulator-26131')
SUCCEEDED: wakelock ('wearable-5.5', 'W-5.5-circle-x86', 'emulator-26121')
SUCCEEDED: wakelock ('wearable-4.0', 'w-0521-1', 'emulator-26111')
SUCCEEDED: wakelock ('tv-6.0', 'T-samsung-6.0-x86', 'emulator-26101')
All tests passed!

./tools/run_command.py test --plugins wakelock --platforms wearable-5.5
SUCCEEDED: wakelock ('wearable-5.5', 'W-5.5-circle-x86', 'emulator-26121')
All tests passed!
```

### DIFF
--- a/tools/commands/integration_test.py
+++ b/tools/commands/integration_test.py
@@ -1,16 +1,32 @@
-"""A script that helps running integration tests for multiple Tizen plugins.
-
-To run integrations tests for all plugins under packages/, 
-run tools/run_integration_test.py in the root of this repository.
+"""A subcommand that runs integration tests for multiple Tizen plugins 
+under packages.
 
 Here are some of the common use cases. Assume packages contain plugins 
 a, b, c, d, and e.
-- running specific packages:
-tools/run_integration_test.py --plugins a b c # runs a, b, c
+- testing specific packages:
+tools/run_command.py test --plugins a b c # runs a, b, c
 - excluding some packages:
-tools/run_integration_test.py --exclude a b c # runs d, e
+tools/run_command.py test --exclude a b c # runs d, e
 - exclude has precedence:
-tools/run_integration_test.py --plugins a b c --exclude b # runs a c
+tools/run_command.py test --plugins a b c --exclude b # runs a c
+
+By default, the tool will run integration tests on all connected targets.
+A target is a Tizen device (either a physical device or an emulator) that can
+run Flutter applications. Generally targets are connected when device is 
+connected to the host PC either with a cable or wirelessly for physical device,
+and when launching targets from Tizen SDK's Emulator Manager for emulators.
+
+Each target has three important pieces of information:
+- name: the name of the target.
+- platform: defines the device profile and Tizen version, usually expressed in
+<device_profile>-<tizen_version> (ex: wearable-5.5, tv-6.0).
+- id: the identifier assigned to a **connected** target.
+
+Here are some of the use cases where the information can be useful:
+- testing all targets that satisfies platform:
+tools/run_command.py test --platform wearable-5.5
+- testing target with id:
+tools/run_command.py test --id some_id
 """
 
 import os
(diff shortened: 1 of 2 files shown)


tools/run_command.py
```diff
@@ -21,15 +21,11 @@
...
-            check_tidy.run_check_tidy(args)
-        elif args.subcommand == 'test':
-            integration_test.run_integration_test(args)
-        elif args.subcommand == 'build':
-            build_example.run_build_examples(args)
-        elif args.subcommand == 'plugins':
-            print_plugins.run_print_plugins(args)
-    except Exception as e:
-        print(e)
-        exit(1)
```
bbrto21, line 35: I'm fine that you remove this,  but I want to know a reason to remove them.

bbrto21 (approved):
LGTM

tools/run_command.py
HakkyuKim: Now that illegal cli arguments are all checked between lines 19~22, the try-catch block will only catch exceptions that's meaningful for debugging purposes only. (The tool is not for end-user developers so I'm not differentiating programming errors and runtime exceptions.) Catching exceptions will only print the error message while letting the program crash will let the python interpreter to print stack trace so it's easy to debug and exit with code 1 automatically.

bbrto21:
I can't test with wearable device not emulator.
because, the Tizen version includes the patch number on device.
Can someone test below command on product? I only have a test device with unofficial platform image now.

```
sdb capability 
platform_version:5.5.0.2
```

HakkyuKim:
@bbrto21 
I did not realize there were more numbers behind the first dot, I'll fix this.

tools/commands/integration_test.py
```diff
@@ -132,105 +132,115 @@ def _integration_test(plugin_dir, test_targets, timeout):
...
+                    TestResult.success(plugin_name, test_target))
+            else:
+                test_results.append(
+                    TestResult.fail(plugin_name, test_target, errors=errors))
+    finally:
+        subprocess.run('flutter-tizen clean',
+                       shell=True,
+                       cwd=example_dir,
+                       stderr=subprocess.PIPE,
+                       stdout=subprocess.PIPE)
```
swift-kim, line 243: You don't need set `stderr` and `stdout` to `subprocess.PIPE` in this case.

HakkyuKim: Right, thanks!

HakkyuKim:
> @bbrto21 I did not realize there were more numbers behind the first dot, I'll fix this.

Actually, I've set the default target as wearable-5.5 for now and since wearable-5.5.0.2 is technically not the exact version, the error message I've seen offline on your machine is correct.

I thought about improving this, but left it as a TODO.
https://github.com/flutter-tizen/plugins/blob/0ca3a1073c8aed519c12bb0059de37cee1f7a731/tools/commands/integration_test.py#L100-L102

Since the issue came up, I'll find a better way to set default targets.

HakkyuKim:
@bbrto21 
You should be able to test with any connected targets now.

bbrto21:
@HakkyuKim 
Thank you, it works fine :)

Issue #230, opened 2021-09-28
Run multiple self-hosted runners on a single machine.


HakkyuKim:
As integration tests are slow, we need to set more runners on the server machine. This is not possible in the current implementation because emulators are, to each runner, shared resources.

This could be solved if tools can create and manipulate emulators during test. I wasn't sure if the CI system could create, launch, poweroff, and delete emulators as I met many failures on several cloud systems. Luckily, it seems that manipulating emulators is possible as long as the server machine has a graphic display attached to it.

I quickly tested it on my machine. These two jobs ran simultaneously on the same machine without interfering each other, and there were no signs of significant performance drops, at least for the tests I've done.
- https://github.com/HakkyuKim/plugins/runs/3729901007
- https://github.com/HakkyuKim/plugins/runs/3729903732

PR #232, opened 2021-09-30
Ephemeral targets(emulators)

HakkyuKim:
Work in progess.

contributes to #230. 

Depends on #229, will rebase after #229 is merged.


tools/commands/integration_test.py
```diff
@@ -128,32 +126,136 @@ def run_integration_test(self, plugin_name, directory, timeout):
...
+        if completed_process.returncode != 0:
+            raise Exception('sdb failure.')
+
+        lines = completed_process.stdout.rstrip().split('\n')
+        for line in lines[1:]:
+            tokens = re.split('[\t ]', line)
+            id = tokens[0]
+            name = tokens[-1]
+            if name == self.name:
+                return id
```
WonyoungChoi, line 161:
```suggestion
            tokens = re.split('\s+', line)
            if tokens[-1] == self.name:
                return tokens[0]
```

tools/commands/integration_test.py
```diff
@@ -128,32 +126,136 @@ def run_integration_test(self, plugin_name, directory, timeout):
...
+
+    def launch(self):
+        completed_process = subprocess.run(f'em-cli launch -n {self.name}',
+                                           shell=True,
+                                           cwd='.')
+        if completed_process.returncode != 0:
+            raise Exception(f'Target {self.name} launch failed.')
+        time.sleep(5)
+        self.id = self._find_id()
+        pass
```
WonyoungChoi, line 172:
```suggestion
```

tools/commands/integration_test.py
```diff
@@ -128,32 +126,136 @@ def run_integration_test(self, plugin_name, directory, timeout):
...
+
+    def power_off(self):
+        # The poweroff logic for tv is different because sdb root is not allowed.
+        if self.device_profile == 'tv':
+            completed_process = subprocess.run(
```
WonyoungChoi, line 177: why don't you save and use the pid of the launched emulator instead of finding the pid by using ps and grep?

tools/commands/integration_test.py
```diff
@@ -128,32 +126,136 @@ def run_integration_test(self, plugin_name, directory, timeout):
...
+                universal_newlines=True,
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE)
+            if completed_process.returncode != 0:
+                raise Exception(f'Target {self.id} power off failed.')
+            pid = completed_process.stdout.strip().split(' ')[0]
+            completed_process = subprocess.run(f'kill -9 {pid}',
+                                               shell=True,
+                                               cwd='.',
+                                               stdout=open(os.devnull, 'wb'))
```
WonyoungChoi, line 190: I think `subprocess.check_output` method is more useful when just using the stdout result of the process. It also raise an error when the exit code is non-zero.
```suggestion
            output = subprocess.check_output(f'kill -9 {pid}',
                                               shell=True).decode('utf-8').splitlines()
```

tools/commands/integration_test.py
HakkyuKim: Just so that we're on the same page, do you mean creating an instance variable `self.pid` in `EphemeralTarget.launch()` and using is later in `EphemeralTarget.power_off()`?

WonyoungChoi: yes I think it's more clear if possible.

tools/commands/integration_test.py
```diff
@@ -128,32 +126,136 @@ def run_integration_test(self, plugin_name, directory, timeout):
...
+                return id
+        raise Exception(f'Could not find connected target {self.name}')
+
+    def launch(self):
+        completed_process = subprocess.run(f'em-cli launch -n {self.name}',
+                                           shell=True,
+                                           cwd='.')
+        if completed_process.returncode != 0:
+            raise Exception(f'Target {self.name} launch failed.')
+        time.sleep(5)
```
WonyoungChoi, line 170: Is this `5 seconds` safe enough to guarantee that the launch is completed?

tools/commands/integration_test.py
HakkyuKim: OK, I'll change it.

tools/commands/integration_test.py
HakkyuKim: As far as I know, there's no way to set callbacks or wait the emulator process so the number is set reasonably based on some testings. I haven't had any problems yet but if the emulator doesn't launch in that time, the program will emit the exception in line 160:
```python
raise Exception(f'Could not find connected target {self.name}')
```
I'll leave a comment about why the waiting is required.

tools/commands/integration_test.py
HakkyuKim: Actually I decided to stick with `run` for now if it's OK with you, I believe you suggested it because of repeating `completed_process` checks in `power_off()` before the current changes.

HakkyuKim:
@WonyoungChoi 
Thanks for the review, I've made changes based on your suggestion.

tools/commands/integration_test.py
WonyoungChoi: okay I understand. :)

WonyoungChoi (approved):
LGTM 👍

bwikbs (approved):
It's going a lot faster than I imagined. AWESOME!!! :100:

Issue #223, opened 2021-09-15
Public members lack documentation


bbrto21:
Packages with public APIs are more appropriate to analyze using `dart` rather than `flutter`
However, it is currently reported that there is lack documentation for 59 public members.
after solving all of this, we have to use `dart` instead of `flutter`

- [x] messageport/example/lib/main.dart:6:14 • Document all public members. • public_member_api_docs
- [x] messageport/example/lib/main.dart:7:14 • Document all public members. • public_member_api_docs
- [x] messageport/example/lib/main.dart:29:7 • Document all public members. • public_member_api_docs
- [x] messageport/lib/messageport_tizen.dart:84:16 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:12:7 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:13:3 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:15:16 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:22:16 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:33:16 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:43:16 • Document all public members. • public_member_api_docs
- [x] messageport/lib/src/messageport_manager.dart:56:19 • Document all public members. • public_member_api_docs
- [x] package_info_plus/lib/package_info_plus_tizen.dart:18:15 • Document all public members. • public_member_api_docs
- [x] share_plus/lib/share_plus_tizen.dart:18:15 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/app_control.dart:20:3 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:9:7 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:11:1 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:14:1 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:15:1 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:17:1 • Document all public members. • public_member_api_docs
- [x] url_launcher/lib/src/types.dart:18:1 • Document all public members. • public_member_api_docs
[...]

PR #234, opened 2021-10-05
[multi-packages] Resolve some linter warnings

HakkyuKim:
Resolve warnings from `dart analyze` for `package_info_plus`, `share_plus`, `wearable_rotary`, and `message_port`.

Partially contributes to #223.

### DIFF
--- a/packages/package_info_plus/lib/package_info_plus_tizen.dart
+++ b/packages/package_info_plus/lib/package_info_plus_tizen.dart
@@ -15,6 +15,7 @@ import 'package:package_info_plus_platform_interface/package_info_platform_inter
 /// being set to Linux's.
 /// https://github.com/fluttercommunity/plus_plugins/blob/c7c942d1d19595a536c6ae18168d3781ddef7785/packages/package_info_plus/lib/package_info_plus.dart#L43-L55
 class PackageInfoPlugin extends PackageInfoPlatform {
+  /// Register dart plugin to inject Tizen plugin.
   static void register() {
     // ignore: invalid_use_of_visible_for_testing_member
     PackageInfo.disablePackageInfoPlatformOverride = true;
--- a/packages/messageport/example/integration_test/messageport_test.dart
+++ b/packages/messageport/example/integration_test/messageport_test.dart
@@ -5,8 +5,8 @@
 import 'dart:async';
 
 import 'package:flutter_test/flutter_test.dart';
-import 'package:messageport_tizen/messageport_tizen.dart';
 import 'package:integration_test/integration_test.dart';
+import 'package:messageport_tizen/messageport_tizen.dart';
 
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
--- a/packages/messageport/example/lib/main.dart
+++ b/packages/messageport/example/lib/main.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: public_member_api_docs
+
 import 'dart:async';
 import 'package:flutter/material.dart';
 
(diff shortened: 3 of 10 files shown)


HakkyuKim:
@bbrto21 
Do you know anything about this warning?
It complains when I run `dart analyze` in `share_plus` but not in `package_info_plus`.

```
warning • analysis_options.yaml:1:10 • The include file
          'package:pedantic/analysis_options.yaml' in
          '/home/hakkyu/mydir/project/flutter_tizen_dev/plugins/analysis_options_plus.yaml'
          can't be found when analyzing
          '/home/hakkyu/mydir/project/flutter_tizen_dev/plugins/packages/share_plus'.
          • include_file_not_found
```

packages/package_info_plus/lib/package_info_plus_tizen.dart
```diff
@@ -15,6 +15,7 @@ import 'package:package_info_plus_platform_interface/package_info_platform_inter
 /// being set to Linux's.
 /// https://github.com/fluttercommunity/plus_plugins/blob/c7c942d1d19595a536c6ae18168d3781ddef7785/packages/package_info_plus/lib/package_info_plus.dart#L43-L55
 class PackageInfoPlugin extends PackageInfoPlatform {
+  /// Register dart plugin to inject Tizen plugin.
```
swift-kim, line 18: The meaning is not clear. A Dart plugin refers to a plugin written in pure Dart.

HakkyuKim: How about "Inject Tizen's implementation of the plugin."?

HakkyuKim:
The integration test is failing for `package_info_plus` because the screen of the emulator target running on the internal CI server is off. It does that sometimes after a few days even when I set the screen display to be "always on". This problem will be resolved after applying `--use-ephemeral-targets` in #232.

bbrto21:
> @bbrto21 Do you know anything about this warning? It complains when I run `dart analyze` in `share_plus` but not in `package_info_plus`.
> 
> ```
> warning • analysis_options.yaml:1:10 • The include file
>           'package:pedantic/analysis_options.yaml' in
>           '/home/hakkyu/mydir/project/flutter_tizen_dev/plugins/analysis_options_plus.yaml'
>           can't be found when analyzing
>           '/home/hakkyu/mydir/project/flutter_tizen_dev/plugins/packages/share_plus'.
>           • include_file_not_found
> ```

@HakkyuKim 
Some analysis option have dependencies on other plugins, you need to run flutter-tizen pub get before analyze

HakkyuKim:
> @HakkyuKim Some analysis option have dependencies on other plugins, you need to run flutter-tizen pub get before analyze

@bbrto21 
Not sure why but I couldn't run `flutter-tizen pub get` in `share_plus` due to formatting error in `pubspec.lock`, `flutter-tizen pub get` works after erasing the file and no warnings are shown for `dart analyze`. Thanks for the help.

packages/package_info_plus/lib/package_info_plus_tizen.dart
swift-kim:
```dart
/// Tells [PackageInfo] to use [MethodChannelPackageInfo] as the default instance of [PackageInfoPlatform].
```
should be enough I think, or you can move some comments from the class description to this method. Also a method comment should be written with an implied subject of _This method_, so it should be like "Injects" but not "Inject".

HakkyuKim:
> Hmm...the same problem occurs even after restarting the emulator. This is unexpected, I would have to check when I'm back in office.

It was the screen problem which I knew about.

PR #235, opened 2021-10-07
[webview_flutter] Update dependencies for 2.5.1

bwikbs:


swift-kim:
webview_flutter has been migrated to the federated plugin architecture since 2.1.0. You need to update dependencies and source files.

bwikbs:
@swift-kim You're right. After clean, It's not working at all. 😢

PR #236, opened 2021-10-07
[webview_flutter] Apply upstream (webview_flutter-v2.1.1)

bwikbs:
Signed-off-by: MuHong Byun <mh.byun@samsung.com>

PR #240, opened 2021-10-08
Update flutter version in CI script

HakkyuKim:


HakkyuKim:
Closing since #236 will cover this.

Issue #233, opened 2021-10-05
[permission_handler] The plugin asks for permission multiple times


swift-kim:
The plugin asks for the same permission repeatedly when a permission is denied by user.

How to reproduce:

1. Run the example app and click either of Permission.(calendar, contacts, location, locationAlways, locationWhenInUse, sms).
2. Deny when a permission request dialog is prompted. The plugin will ask for the same permission again.

Why this occurs:

Some privileges are redundantly pushed back to `permissions_to_request`.

https://github.com/flutter-tizen/plugins/blob/0c5e36251f4d56f3971a203c7c25c32699fbece8/packages/permission_handler/tizen/src/permission_manager.cc#L88-L128

According to https://docs.tizen.org/application/native/guides/security/privacy-related-permissions,

> Since the privileges are grouped, the user’s decision regarding one privilege applies to the whole group of related privileges. For example, if the user has granted permission to use the `http://tizen.org/privilege/account.read` privilege, permission is automatically granted to the `http://tizen.org/privilege/account.write` privilege also.

cc @bbrto21

swift-kim:
`geolocator_tizen` should have the same problem.

https://github.com/flutter-tizen/plugins/blob/master/packages/geolocator/tizen/src/permission_manager.cc

bwikbs:
Wow, You already finished the analysis! Are you already fixing it? Or just telling @bbrto21 ?

swift-kim:
@bwikbs No, I haven't touched anything. Anyone can work on this issue.

PR #241, opened 2021-10-08
[geolocator] Request the location permission only once

swift-kim:
Contributes to #233. There's no need to request the `location.coarse` permission.

See b4d9e959853f31adcb61adf8c2f084d70ec7fe07 for the actual code diff.

Side note: `ecore_main_loop_iterate` is no longer necessary after this change and thus the logic can be refactored.

### DIFF
--- a/packages/geolocator/README.md
+++ b/packages/geolocator/README.md
@@ -9,7 +9,7 @@ The Tizen implementation of [`geolocator`](https://github.com/Baseflow/flutter-g
  ```yaml
 dependencies:
   geolocator: ^7.4.0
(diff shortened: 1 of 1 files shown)


swift-kim:
@HakkyuKim Oh, I just forgot. Done!

swift-kim:
Who has published the first version of geolocator_tizen? The original publisher must transfer their ownership to tizen.org.

bbrto21:
@swift-kim Oh, I'm so sorry, I didn't know that until you mention it. and I just transfer thees ownership to tizen.org.

PR #258, opened 2021-10-29
Fix video player plugin freeze issue

xiaowei-guan:
When flutter engine upgrade to 2.5.1, only call mark frame available
API can triggle previous frame destruction callback.
#250

PR #260, opened 2021-11-02
[camera] Implement pausePreview and resumePreview

swift-kim:
- [x] Increase the frontend package version to 0.9.4.
- [x] Implement `pausePreview` and `resumePreview`. (Added in camera_platform_interface 2.1.0.)
- [x] Update example and miscellaneous files.

@bbrto21 The example app doesn't quite work smoothly on TM1, even with https://github.com/flutter-tizen/plugins/pull/258 applied to video_player_tizen. Could you take a look?


bbrto21:
> @bbrto21 The example app doesn't quite work smoothly on TM1, even with #258 applied to video_player_tizen. Could you take a look?

I think it seems to be a similar issue with video plugin freezing. I'll open new issue

packages/camera/tizen/src/camera_device.cc
```diff
@@ -993,6 +993,11 @@ void CameraDevice::Open(
           self->prepared_packet_ = packet;
           return;
         }
+        if (self->is_preview_paused_) {
+          media_packet_destroy(packet);
+          self->prepared_packet_ = nullptr;
+          return;
+        }
```
swift-kim, line 1000: Simply discard the prepared packet when `is_preview_paused_` is true, without stopping the preview itself (`camera_stop_preview`).

How do you like this approach? @bbrto21

bbrto21: Great! but the relationship between `Is_preview_paused_` and `StopCameraPreview` seems ambiguous. how about adding some comments?

Remark: Preview affects the state of the camera device. also it always pause the preview before start recording using `StopCameraPreview`

swift-kim: If I can safely call `StopCameraPreview` and `StartCameraPreview` at any time and it makes more sense to you, I'll change the implementation.

bbrto21: I'm totally fine with current Implementation. but I'll check the safety when I get back to the office :)

bbrto21 (approved):

bwikbs (approved):
Thanks! :+1:

packages/camera/tizen/src/camera_device.cc
swift-kim: I tested [this change](https://github.com/swift-kim/plugins/commit/fd32be23507dddfbc7c80becefb4d7202bf95bc4) on a TM1 device today, but it didn't work as expected, because the capture API requires the camera state to be `CAMERA_STATE_PREVIEW` before capturing. Even if I call `StartCameraPreview` before calling `StartCameraCapture` and call `StopCameraPreview` again after the capture is finished, the call to `StopCameraPreview` results in an `Invalid state` error for unknown reason.

The problem of the current implementation is that the preview textures are not delivered to Flutter even when the user is recording a video. This is inconsistent with the Android side implementation. Let's revisit this issue when it's needed.

swift-kim:
Another issue: Switching between camera devices (front and back cameras) leads to app freezing on TM1 devices quite frequently. The issue is not related to this PR.

bbrto21:
@swift-kim in this PR the test driver was deleted. however, no new test drivers were added. should we add new one?

swift-kim:
@bbrto21 Basically we don't need test drivers for most of our plugins, because we now run tests with `integration_test` but not `flutter_driver` any more. If you need a special environmental setup for testing this plugin, you can write a new test driver and run the test with `flutter_driver` (using the `flutter-tizen drive` command).

Issue #261, opened 2021-11-02
[VideoPlayer]Null check operator used on a null value


xiaowei-guan:
E/ConsoleMessage(19508): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: Null check operator used on a null value
E/ConsoleMessage(19508): #0      VideoPlayerController.initialize.errorListener (package:video_player/video_player.dart:381:51)
E/ConsoleMessage(19508): #1      _rootRunUnary (dart:async/zone.dart:1436:47)
E/ConsoleMessage(19508): #2      _CustomZone.runUnary (dart:async/zone.dart:1335:19)
E/ConsoleMessage(19508): #3      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1244:7)
E/ConsoleMessage(19508): #4      _BufferingStreamSubscription._sendError.sendError (dart:async/stream_impl.dart:362:15)
E/ConsoleMessage(19508): #5      _BufferingStreamSubscription._sendError (dart:async/stream_impl.dart:378:7)
E/ConsoleMessage(19508): #6      _BufferingStreamSubscription._addError (dart:async/stream_impl.dart:280:7)
E/ConsoleMessage(19508): #7      _ForwardingStreamSubscription._addError (dart:async/stream_pipe.dart:128:11)
E/ConsoleMessage(19508): #8      _ForwardingStream._handleError (dart:async/stream_pipe.dart:95:10)
E/ConsoleMessage(19508): #9      _ForwardingStreamSubscription._handleError (dart:async/stream_pipe.dart:157:13)
E/ConsoleMessage(19508): #10     _rootRunBinary (dart:async/zone.dart:1452:47)
E/ConsoleMessage(19508): #11     _CustomZone.runBinary (dart:async/zone.dart:1342:19)
E/ConsoleMessage(19508): #12     _
E/ConsoleMessage(19508): CustomZone.runBinaryGuarded (dart:async/zone.dart:1252:7)
E/ConsoleMessage(19508): #13     _BufferingStreamSubscription._sendError.sendError (dart:async/stream_impl.dart:360:15)
E/ConsoleMessage(19508): #14     _BufferingStreamSubscription._sendError (dart:async/stream_impl.dart:378:7)
E/ConsoleMessage(19508): #15     _DelayedError.perform (dart:async/stream_impl.dart:602:14)
[...]

PR #262, opened 2021-11-02
[video_player] Never return empty error messages

swift-kim:
- Never return empty error messages to avoid null reference exceptions. Fixes https://github.com/flutter-tizen/plugins/issues/261.
- Update video_player to 2.2.6 and update the example app.
- Minor cleanups.


xiaowei-guan (approved):
Good！

PR #267, opened 2021-11-08
[multi-packages] Migrate plus plugins to use flutter_lints

HakkyuKim:
Closes #254.

References: 
- https://github.com/fluttercommunity/plus_plugins/pull/519
- https://github.com/fluttercommunity/plus_plugins/pull/537

PR #269, opened 2021-11-09
[webview_flutter] Update webview_flutter (2.3.0) 

bwikbs:
**webview_flutter**
- Update  base to ``2.3.0``

#267

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -845,7 +845,8 @@ void WebView::HandleMethodCall(
     result->Success();
   } else if (method_name.compare("currentUrl") == 0) {
     result->Success(flutter::EncodableValue(webview_instance_->GetURL()));
-  } else if (method_name.compare("evaluateJavascript") == 0) {
+  } else if (method_name.compare("evaluateJavascript") == 0 ||
+             method_name.compare("runJavascriptReturningResult") == 0) {
     if (std::holds_alternative<std::string>(arguments)) {
       std::string js_string = std::get<std::string>(arguments);
       webview_instance_->EvaluateJavaScript(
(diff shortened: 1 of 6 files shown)


bwikbs:
@HakkyuKim
Now, webview is ready. Google Maps should be after this.

bbrto21:
The problem below can be fixed after this PR.

```
   info - google_maps_flutter/lib/src/circle.dart:31:32 - 'evaluateJavascript' is deprecated and shouldn't be used. Use [runJavascript] or [runJavascriptReturningResult]. Try replacing the use of the deprecated member with the replacement. - deprecated_member_use
```

bbrto21 (approved):

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -859,6 +860,20 @@ void WebView::HandleMethodCall(
...
+          js_string, [res = result.release()](std::string value) {
+            LOG_DEBUG("value: %s\n", value.c_str());
+            if (res) {
+              res->Success();
+              delete res;
+            }
+          });
+    } else {
+      result->Error("InvalidArguments", "Please set javascript string");
+    }
```
HakkyuKim, line 876: How about:

```cpp
else if (method_name.compare("evaluateJavascript") == 0 ||
             method_name.compare("runJavascriptReturningResult") == 0 ||
             method_name.compare("runJavascript") == 0) {
    if (std::holds_alternative<std::string>(arguments)) {
      std::string js_string = std::get<std::string>(arguments);
      webview_instance_->EvaluateJavaScript(
          js_string, [res = result.release()](std::string value) {
            LOG_DEBUG("value: %s\n", value.c_str());
            if (res) {
              bool returns_value =
                  method_name.compare("runJavascript") == 0 ? false : true;
              if (returns_value) {
                res->Success(flutter::EncodableValue(value));
              } else {
                res->Success();
              }
              delete res;
            }
          });
    } else {
      result->Error("InvalidArguments", "Please set javascript string");
    }
  }
```

HakkyuKim:
> The problem below can be fixed after this PR. we can ignore it.
> 
> ```
>    info - google_maps_flutter/lib/src/circle.dart:31:32 - 'evaluateJavascript' is deprecated and shouldn't be used. Use [runJavascript] or [runJavascriptReturningResult]. Try replacing the use of the deprecated member with the replacement. - deprecated_member_use
> ```

OK, all analyze errors are related to `google_maps`, we can force merge this first, then #270 can rebase afterwards.

packages/webview_flutter/tizen/src/webview.cc
bwikbs: This one is much better, thank you! :+1:

HakkyuKim (approved):
LGTM

Issue #268, opened 2021-11-08
[camera] Preview freeze issue


bbrto21:
> @bbrto21 The example app doesn't quite work smoothly on TM1, even with #258 applied to video_player_tizen. Could you take a look?

> I think it seems to be a similar issue with video plugin freezing. I'll open new issue

_Originally posted by @swift-kim in https://github.com/flutter-tizen/plugins/pull/260#issue-1041908191

PR #275, opened 2021-11-15
[camera] Fix preview freeze issue

bbrto21:

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -49,8 +49,6 @@ bool IsValidMediaPacket(media_packet_h media_packet) {
   return true;
 }
 
-}  // namespace
-
 ExifTagOrientation ChooseExifTagOrientatoin(OrientationType device_orientation,
                                             bool is_front_lens_facing) {
   ExifTagOrientation orientation = ExifTagOrientation::kTopLeft;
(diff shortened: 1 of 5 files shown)


bbrto21:
fix #268

bwikbs (approved):
:+1:

packages/camera/tizen/src/camera_device.cc
```diff
@@ -38,8 +40,19 @@ static std::string CreateTempFileName(const std::string &prefix,
...
+  int ret = media_packet_get_tbm_surface(media_packet, &surface);
+  if (ret != MEDIA_PACKET_ERROR_NONE) {
+    return false;
+  }
+  return true;
+}
+
+}  // namespace
+
+ExifTagOrientation ChooseExifTagOrientatoin(OrientationType device_orientation,
```
HakkyuKim, line 54: Is there a reason for not including this in the anonymous namespace after removing `static`?

packages/camera/tizen/src/camera_device.cc
```diff
@@ -79,7 +92,7 @@ static ExifTagOrientation ChooseExifTagOrientatoin(
   return orientation;
 }
 
-static RecorderOrientationTag ChooseRecorderOrientationTag(
+RecorderOrientationTag ChooseRecorderOrientationTag(
```
HakkyuKim, line 95: Ditto.

packages/camera/tizen/src/camera_device.cc
bbrto21: Oh... I thought I'd need it somewhere else someday, now that I think about it, I don't think there's any need for that.
I wrapped the functions in anonymous namespaces :)

HakkyuKim (approved):
LGTM

bbrto21:
published

PR #279, opened 2021-11-25
[webview_flutter] Fixed memory releasing for non-heap value

pkosko:
string literal was used to initialize 'path' variable, but later in code
free() was called for it. It causes releasing memory for non-heap value.

[Verification] Plugin builds properly.

### DIFF
--- a/packages/webview_flutter/tizen/src/webview_factory.cc
+++ b/packages/webview_flutter/tizen/src/webview_factory.cc
@@ -24,21 +24,22 @@ WebViewFactory::WebViewFactory(flutter::PluginRegistrar* registrar,
                                flutter::TextureRegistrar* textureRegistrar)
     : PlatformViewFactory(registrar), texture_registrar_(textureRegistrar) {
   char* path = app_get_data_path();
+  std::string path_string;
   if (!path || strlen(path) == 0) {
-    path = "/tmp/";
-  }
-  LOG_DEBUG("application data path : %s\n", path);
-  std::string localstoragePath = path + std::string("StarFish_localStorage.db");
-  std::string cookiePath = path + std::string("StarFish_cookies.db");
-  std::string cachePath = path + std::string("Starfish_cache.db");
-
-  LWE::LWE::Initialize(localstoragePath.c_str(), cookiePath.c_str(),
-                       cachePath.c_str());
-
-  if (path) {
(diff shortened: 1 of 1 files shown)


packages/webview_flutter/tizen/src/webview_factory.cc
```diff
@@ -24,21 +24,22 @@ WebViewFactory::WebViewFactory(flutter::PluginRegistrar* registrar,
...
+    free(path);
+    path = nullptr;
   }
-  LOG_DEBUG("application data path : %s\n", path);
-  std::string localstoragePath = path + std::string("StarFish_localStorage.db");
-  std::string cookiePath = path + std::string("StarFish_cookies.db");
-  std::string cachePath = path + std::string("Starfish_cache.db");
+  LOG_DEBUG("application data path : %s\n", path_string.c_str());
+  std::string localstoragePath =
+      path_string + std::string("StarFish_localStorage.db");
```
swift-kim, line 37: By the way you can rename all the local variables in this file with snake casing.

swift-kim (approved):

seungsoo47:
Wow, can you tell me which variable's memory is released? How did you check this?

packages/webview_flutter/tizen/src/webview_factory.cc
```diff
@@ -21,36 +21,38 @@
...
 
 WebViewFactory::WebViewFactory(flutter::PluginRegistrar* registrar,
-                               flutter::TextureRegistrar* textureRegistrar)
-    : PlatformViewFactory(registrar), texture_registrar_(textureRegistrar) {
+                               flutter::TextureRegistrar* texture_registrar)
+    : PlatformViewFactory(registrar), texture_registrar_(texture_registrar) {
   char* path = app_get_data_path();
+  std::string path_string;
   if (!path || strlen(path) == 0) {
-    path = "/tmp/";
```
pkosko, line 28: @seungsoo47 here is the line which was causing a releasing a non-heap memory with free() - refer to line 39.

I've removed this condition and replaced conditional memory releasing with a std::string-based stack memory management.

seungsoo47: 👍

seungsoo47 (approved):

PR #283, opened 2021-12-01
Show first frame after player prepared

xiaowei-guan:
image


swift-kim:
The change looks good but what's the difference?

xiaowei-guan:
> The change looks good but what's the difference?

when player is prepared and  play() is not called, the first frame is displayed.

if no this PR, only show app background.

swift-kim (approved):
Thank you!

PR #285, opened 2021-12-02
[path_provider] Update path_provider to 2.0.7

swift-kim:
`getDownloadsDirectory` has been removed because it is only relevant on desktop platforms. Use `getExternalStorageDirectories` with `StorageDirectory.downloads` as argument instead.

### DIFF
--- a/packages/path_provider/example/integration_test/path_provider_test.dart
+++ b/packages/path_provider/example/integration_test/path_provider_test.dart
@@ -74,35 +74,15 @@ void main() {
         final Future<List<Directory>?> result =
             getExternalStorageDirectories(type: null);
         expect(result, throwsA(isInstanceOf<UnsupportedError>()));
-      } else if (Platform.isAndroid) {
+      } else {
         final List<Directory>? directories =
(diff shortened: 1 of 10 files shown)


packages/path_provider/example/integration_test/path_provider_test.dart
```diff
@@ -74,35 +74,15 @@ void main() {
...
-
-        // final List<Directory>? directories =
-        //     await getExternalStorageDirectories(type: type);
-        // expect(directories, isNotNull);
-        // for (final Directory result in directories!) {
-        //   _verifySampleFile(result, '$type');
-        // }
       }
-    }, skip: true);
+    });
```
swift-kim, line 85: Now that we can choose emulator instances where integrations tests are to be run, this skipped test can be unskipped.

bbrto21 (approved):

bwikbs (approved):
GOOOD :+1:

swift-kim:
This package is missing .gitignore file. I'll update soon and then publish.

PR #286, opened 2021-12-02
[permission_handler] Update permission_handler to 8.3.0

swift-kim:
Only Android- and iOS-specific permissions have been added to [permission_handler_platform_interface](https://pub.dev/packages/permission_handler_platform_interface/changelog) (between 3.1.1 and 3.7.0).


packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc
```diff
@@ -10,16 +10,13 @@
 #include "service_manager.h"
 
 namespace {
-constexpr char kPermissionHandlerChannelName[] =
```
bbrto21, line 13: Do you have a specific reason to remove it?

bbrto21 (approved):

packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc
swift-kim: Most of the plugins in this repo use string literals as channel names without declaring them as constants.

packages/permission_handler/example/tizen/tizen-manifest.xml
```diff
@@ -1,11 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest package="org.tizen.permission_handler_tizen_example" version="1.0.0" api-version="4.0" xmlns="http://tizen.org/ns/packages">
-    <profile name="common"/>
-    <ui-application appid="org.tizen.permission_handler_tizen_example" exec="Runner.dll" type="dotnet" multiple="false" taskmanage="true" nodisplay="false" api-version="4" launch_mode="single">
+    <profile name="wearable"/>
```
bwikbs, line 3: Is this intended?

packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc
bwikbs: I like this way!

bwikbs (approved):
Thanks :+1:

packages/permission_handler/example/tizen/tizen-manifest.xml
swift-kim: Yes :) https://github.com/flutter-tizen/plugins/pull/281#discussion_r760807695

bwikbs: OMG! Everyone thinks the same way! :)

swift-kim: FYI, there's also a written guideline for this. :)

https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline#tizen-manifestxml

PR #292, opened 2021-12-06
[share_plus] Reimplement in Dart

swift-kim:
Reimplemented based on the tizen_app_control package.


bbrto21 (approved):

packages/share_plus/lib/share_plus_tizen.dart
```diff
@@ -2,23 +2,61 @@
...
+        'http://tizen.org/appcontrol/data/text': text,
+        if (hasSubject) 'http://tizen.org/appcontrol/data/subject': subject,
+      },
+    ).sendLaunchRequest();
+  }
+
+  @override
+  Future<void> shareFiles(
+    List<String> paths, {
+    List<String>? mimeTypes,
```
bwikbs, line 44: Is it okay if we don't set ``mimeTypes`` ?

swift-kim: I referred to https://docs.tizen.org/application/native/guides/app-management/common-appcontrols and there was no such requirement.

bwikbs: Previous code has following part.
```c++
    // Set mime type & image paths
    if (path_array.size() > 0) {
      ret = app_control_set_mime(handle, mime_type.c_str());
      RET_IF_ERROR(ret);

      ret = app_control_add_extra_data_array(
          handle, APP_CONTROL_DATA_PATH, path_array.data(), path_array.size());
      RET_IF_ERROR(ret);
    }
```
Just in case, I'm wondering if there is any need for the part corresponding to call ``app_control_set_mime``.

bwikbs (approved):
Cool work!! 👍

bbrto21:
published

Issue #272, opened 2021-11-09
[tizen_app_control] Integration test is failing


seungsoo47:
@swift-kim Integration test for tizen_app_control is failing like following. Could you check the problems?

$ ./tools/run_command.py test --plugins tizen_app_control --platforms mobile-6.5
============= Testing for tizen_app_control (1/1) =============

00:00 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart                                                                                                                         Building a Tizen application in debug mode...
The project type of messageport_tizen plugin is "sharedLib", which is deprecated in favor of "staticLib".

00:21 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart
00:22 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart                                                                                                                         The public profile is used for signing.

00:34 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart
00:35 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart                                                                                                                         Building a Tizen application in debug mode...                      36.3s
✓ Built build/tizen/tpk/org.tizen.tizen_app_control_example-1.0.0.tpk (26.5MB).

00:37 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart                                                                                                                         Installing build/tizen/tpk/org.tizen.tizen_app_control_example-1.0.0.tpk...
01:03 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart
01:04 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart                                                                                                                            27.0s

01:09 +0: loading /tmp/flutter_tools.KGWRDJ/KLIKGX/tizen_app_control_test.dart
01:11 +1: Can receive request from platform
01:11 +2: Can find matching applications
01:11 +3: Can send and receive request
01:11 +4: Omit invalid extra data
01:22 +4: Can send and receive reply
01:22 +4 -1: Can send and receive reply [E]
TimeoutException after 0:00:10.000000: Test timed out after 10 seconds.
dart:isolate  _RawReceivePortImpl._handleMessage

01:22 +5 -1: Cannot find target applications
[...]

PR #293, opened 2021-12-07
Update the test recipe and workflow

swift-kim:
Only use the recipe file without having to specify excluded packages (`--exclude`) again, as suggested in https://github.com/flutter-tizen/plugins/issues/272#issuecomment-964024353.

### DIFF
--- a/tools/commands/integration_test.py
+++ b/tools/commands/integration_test.py
@@ -456,11 +456,7 @@ def _integration_test(plugin_dir, platforms, timeout, generate_emulators):
             if not platforms:
                 platforms.extend(target_manager.get_platforms())
                 if not platforms:
-                    return [
-                        TestResult.fail(
-                            plugin_name,
-                            errors=['Cannot find any testable targets.'])
-                    ]
+                    return [TestResult.success(plugin_name, 'None')]
             errors = []
             completed_process = subprocess.run('flutter-tizen pub get',
                                                shell=True,
@@ -541,7 +537,7 @@ def run_integration_test(args):
     for testing_plugin in testing_plugins:
         test_num += 1
         print(
-            f'============= Testing for {testing_plugin} ({test_num}/{total_plugin_num}) ============='
+            f'============= Testing {testing_plugin} ({test_num}/{total_plugin_num}) ============='
         )
         platforms = args.platforms
         if testing_plugin in platforms_per_plugin:
--- a/tools/commands/integration_test.py
+++ b/tools/commands/integration_test.py
@@ -35,7 +35,7 @@ _LOG_PATTERN = r'\d\d:\d\d\s+([(\+\d+\s+)|(~\d+\s+)|(\-\d+\s+)]+):\s+(.*)'
 
 class Target:
     """A Tizen device that can run Flutter applications.
-    
+
     A target is a Tizen device (either a physical device or an emulator) that can
     run Flutter applications. Generally targets are connected when physical devices
     are connected to the host PC either with a cable or wirelessly, and when 
@@ -57,7 +57,7 @@ class Target:
 
     def run_integration_test(self, plugin_name, directory, timeout):
         """Runs integration test in the given directory.
-        
+
         Args:
             plugin_name (str): The name of the testing plugin.
             directory (str): The path to the directory in which to perform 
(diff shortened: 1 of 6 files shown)


tools/commands/integration_test.py
```diff
@@ -456,11 +456,7 @@ def _integration_test(plugin_dir, platforms, timeout, generate_emulators):
             if not platforms:
                 platforms.extend(target_manager.get_platforms())
                 if not platforms:
-                    return [
-                        TestResult.fail(
-                            plugin_name,
-                            errors=['Cannot find any testable targets.'])
-                    ]
+                    return [TestResult.success(plugin_name, 'None')]
```
HakkyuKim, line 459: I'm not quite convinced that this test scenario is a success, how about https://github.com/HakkyuKim/plugins/commit/1eb75d8cb8405a67e62e0a7e0e480a7ce9e5a1bc ?

swift-kim: Your code in https://github.com/HakkyuKim/plugins/commit/1eb75d8cb8405a67e62e0a7e0e480a7ce9e5a1bc is not very readable.

Without your change:
- New package, not added to recipe, has no integration test: Fail
- New package, not added to recipe, has integration test: Success
- New package, added to recipe (empty list), has no integration test: Fail
- New package, added to recipe (empty list), has integration test: Success
- New package, added to recipe (non-empty list), has no integration test: Fail
- New package, added to recipe (non-empty list), has integration test: Run tests on the specified targets

Are you concerned with the second case? (Please correct me if I'm wrong.)

swift-kim: The third case is not tested and probably the code needs a fix.

HakkyuKim: I'm actually concerned with the 4th case, the test should not be a success in my opinion because the integration test did not run on any target. Also, if recipe file is not used, this will happen:

```
# Without any launched emulator:
./tools/run_command.py test --plugins audioplayers
============= Testing audioplayers (1/1) =============
SUCCEEDED: audioplayers None
All tests passed!

# With a launched emulator:
./tools/run_command.py test --plugins audioplayers
============= Testing audioplayers (1/1) =============
00:00 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart                                                                                                                              Building a Tizen application in debug mode...
00:18 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart
The MyProfile profile is used for signing.

00:28 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart
00:29 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart                                                                                                                              Building a Tizen application in debug mode...                      29.4s
✓ Built build/tizen/tpk/org.tizen.audioplayers_tizen_example-1.0.0.tpk (30.1MB).

00:30 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart                                                                                                                              Installing build/tizen/tpk/org.tizen.audioplayers_tizen_example-1.0.0.tpk...
00:32 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart
00:33 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart                                                                                                                                  3.9s

00:36 +0: loading /tmp/flutter_tools.AZFYFE/OMXNHK/audioplayers_test.dart
00:37 +1: asset audio can be initialized
00:39 +2: asset audio can be played
00:39 +3: asset audio can seek
00:41 +4: asset audio can be paused
00:43 +5: asset audio do not exceed duration after audio completed
00:43 +5: (tearDownAll)
00:43 +5: All tests passed!
SUCCEEDED: audioplayers ('wearable-5.5', 'hello', 'emulator-26101')
All tests passed!
```
As you can see, both test resulted in a success but only one was actually tested on a target.

With the changes I suggested, the tool will now only see packages explicitly listed in the recipe file. For example, if `/packages/` contain `a`, `b`, `c`, `d`, and the recipe file only listed `a`, `b`, `c`, then package `d` will not be visible to the tool. If the recipe file specified an empty list for package `c`, the tool will exclude that package from testing. So the result may be something like this:

```
SUCCEEDED: a ('wearable-5.5', 'hello', 'emulator-26101')
SUCCEEDED: b ('wearable-5.5', 'hello', 'emulator-26101')
All tests passed!
EXCLUDED: c
```

And the results of the cases you describe will be:
- New package, not added to recipe, has no integration test: Not recognized by the tool (because not in recipe)
- New package, not added to recipe, has integration test: Not recognized by the tool (because not in recipe)
- New package, added to recipe (empty list), has no integration test: Excluded (because empty list)
- New package, added to recipe (empty list), has integration test: Excluded (because empty list)
- New package, added to recipe (non-empty list), has no integration test: Fail
- New package, added to recipe (non-empty list), has integration test: Run tests on the specified targets

For first and second case, we can tell the developer to update the recipe file when there's a PR for a new plugin. The reason why I changed the tool to see only packages in the recipe file is because I wanted to ignore `--plugins` and `--exclude` options simultaneously when --recipe option is used. (Or if you think those cases should be handled differently, I'm open to suggestions.)

swift-kim: I don't fully understand your comment but the result makes sense. I applied your change and modified a bit to make it more pythonic.

HakkyuKim (approved):
Great, LGTM!

bwikbs (approved):
Thanks!!

PR #298, opened 2021-12-08
[connectivity_plus] Update connectivity_plus to 2.1.0

swift-kim:
- Remove `connectivity_plus_tizen.dart` which is not relevant any more (see https://github.com/fluttercommunity/plus_plugins/commit/7e898675fbd18cb6ac90f36c1e6bd126aea1c64d).
- Support `ethernet` as a connectivity result.


packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc
```diff
@@ -42,7 +40,7 @@ class ConnectivityPlusTizenPlugin : public flutter::Plugin {
   void RegisterObserver(
       std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> &&events) {
     EnsureConnectionHandle();
-    if (connection_set_type_changed_cb(connection_, ConnectionTypeChangedCB,
+    if (connection_set_type_changed_cb(connection_, OnConnectionTypeChanged,
```
bbrto21, line 43: Why don't you use a simple lambda function instead of this?

swift-kim: Oh, it makes sense.

bbrto21 (approved):

PR #296, opened 2021-12-08
Reduce test failures by polling device boot status

HakkyuKim:
Issue and solution suggested by @swift-kim, thanks!

~~The last commit will be reverted after test passes.~~ passed.

PR #299, opened 2021-12-09
[device_info_plus] Make values nullable

swift-kim:
- Remove dependency on `device_info_plus_platform_interface`.
- Make values of `TizenDeviceInfo` nullable.
- Add values `profile`, `platformName`, `platformProcessor`, and `tizenId`.
- Make `MethodChannelDeviceInfo` invisible from users.


swift-kim:
The test is failing because Flutter 2.8 is out today. We need to fix it in another PR.

HakkyuKim (approved):
LGTM

HakkyuKim:
The integration failure is caused by the same issue you mentioned last Thursday, #296 will fix it if my assumption about the problem is correct.

PR #303, opened 2021-12-13
Fix --run-on-changed-packages option

HakkyuKim:
Fixes `--run-on-changed-packages` option.

Summary:
- Rewrote `get_target_plugins` logic.
- Added tests, run by `python3 -m unittest`.
- Relocated some files.

Any recommendation on python formatter?


bbrto21 (approved):
LGTM
```
Ran 7 tests in 0.070s

OK

```

swift-kim:
> Any recommendation on python formatter?

By the way why do we use Python instead of Dart in the first place?

HakkyuKim:
@swift-kim 
Initially there was only a single python file `check_tidy` in `tools`,  so adding `integration_test.py` was simply creating another script file in the directory which made sense to me at the time. I didn't plan to form a command line tool package but it turned out that way as new features were added.  I later thought about rewriting it in dart to remove using python in this repo but it wasn't a priority issue.

PR #312, opened 2021-12-17
[video_player] Fix empty seekCompletedCallback issue

ynoo-nam:
- Assign `on_seek_completed_` of player before calling `player_set_play_position` to fix intermittent errors.

   If `onSeekCompleted()` is called before `on_seek_completed_` is assigned, uninitialized `on_seek_completed_` can cause serious crash or it is not called at all.

### DIFF
--- a/packages/video_player/README.md
+++ b/packages/video_player/README.md
@@ -29,7 +29,7 @@ This package is not an _endorsed_ implementation of `video_player`. Therefore, y
 ```yaml
 dependencies:
   video_player: ^2.2.6
-  video_player_tizen: ^2.4.0
+  video_player_tizen: ^2.3.1
 ```
 
 Then you can import `video_player` in your Dart code:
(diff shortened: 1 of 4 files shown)


packages/video_player/README.md
```diff
@@ -29,7 +29,7 @@ This package is not an _endorsed_ implementation of `video_player`. Therefore, y
 ```yaml
 dependencies:
   video_player: ^2.2.6
-  video_player_tizen: ^2.3.0
+  video_player_tizen: ^2.4.0
```
swift-kim, line 32: The version should be 2.3.1 according to semantic versioning as this patch doesn't add new functionality.

ynoo-nam: I'm increasing minor version because #283 is not published.

swift-kim: https://github.com/flutter-tizen/plugins/pull/283 doesn't still add any functionality.

ynoo-nam: I thought #283 as an improvement of code. (https://semver.org/#spec-item-7)
I'll fix it!

swift-kim (approved):

bbrto21 (approved):

Issue #305, opened 2021-12-15
Rewrite the tooling in Dart


swift-kim:
https://github.com/flutter-tizen/plugins/pull/303#issuecomment-994243133

This doesn't have to happen at once. Existing python scripts can be invoked from Dart code so we can start from migrating the entrypoint script to Dart first.

PR #317, opened 2021-12-28
Add dart entry point to python tool

HakkyuKim:
Part of #305.

~~The last commit will be reverted when CI test passes.~~ passed.

Summary:
- Add dart entry point to python tool.
- Move python `tools` to dart package `flutter_tizen_plugin_tools`.
- Add `flush=True` in `integration_test.py` to show progress logs.
- Remove `from __future__ import print_function` in `check_tidy.py` since the script is running in python3 by `run_command.py`. Also erase installing python2.7 in `analyze.yml`.


nojinjeong:
Question: are we rewriting python logic in dart?

HakkyuKim:
@nojinjeong 
Gradually, yes.

tool/pubspec.yaml
```diff
@@ -0,0 +1,12 @@
...
+repository: https://github.com/flutter-tizen/plugins/tree/master/tool
+version: 0.1.0
+publish_to: none
+
+environment:
+  sdk: '>=2.15.0 <3.0.0'
+
+dependencies:
+  file: ^6.1.0
+  flutter_plugin_tools: ^0.7.3
```
bbrto21, line 12: Are you considering expanding `flutter_plugin_tools`?

HakkyuKim: Not expanding but rather reusing some of the common classes already implemented in `flutter_plugin_tools` since many features are not required for our use case.

.github/workflows/analyze.yml
```diff
@@ -24,12 +24,15 @@ jobs:
...
       - name: Install clang-format
         run: |
           sudo apt-get update
           sudo apt-get install clang-format-11
+      - name: Set up tools
+        run: dart pub get
+        working-directory: ${{ github.workspace }}/tool
       - name: Check tidy
-        run: tools/run_command.py tidy --dir packages
+        run: dart run ./tool/bin/flutter_tizen_plugin_tools.dart tidy --dir packages
```
swift-kim, line 38: Why don't you create an entry point script (`.sh`) for the Dart file and run `pub get` inside it?

tool/pubspec.yaml
bbrto21: Seems like a good try. thank you for quick answer.

bbrto21 (approved):

tool/lib/src/main.dart
```diff
@@ -0,0 +1,54 @@
...
+      io.exit(1);
+    }
+  }
+
+  const ProcessRunner processRunner = ProcessRunner();
+  final Directory sourceTreeRoot = packagesDir.parent;
+  final File pythonTool = sourceTreeRoot
+      .childDirectory('tool')
+      .childDirectory('tools')
+      .childFile('run_command.py');
```
swift-kim, line 36: Hmm. I thought argument parsing would also be done in this Dart file but the `run_command.py` file is still there. Will it be the next step?

.github/labeler.yml
```diff
@@ -49,8 +49,8 @@
   - packages/webview_flutter/**/*
 
 # Tools
-'tools':
-  - tools/**/*
+'tool':
+  - tool/**/*
```
swift-kim, line 53: `tools` looks better for me :(

tool/lib/src/main.dart
HakkyuKim: Yes, it will be in the next step. At the time, I didn't have a good idea on how to migrate argument parsing to dart while using python commands, so I submitted the work I've done.

.github/labeler.yml
HakkyuKim: I've changed it because the directory holds a single package but it's fine either way, I'll change it back.

.github/workflows/analyze.yml
HakkyuKim: Sounds better, I'll do that!

tools/pubspec.yaml
```diff
@@ -0,0 +1,12 @@
+name: flutter_tizen_plugin_tools
+description: Productivity utils for flutter_tizen/plugins
+repository: https://github.com/flutter-tizen/plugins/tree/master/tool
+version: 0.1.0
```
swift-kim, line 4: You don't have to version this package so this line can be removed.

swift-kim (approved):

PR #321, opened 2022-01-05
Migrate most tool commands to dart

HakkyuKim:
Migrate all commands except integration test to dart. Some commands are replaced to similar ones in `flutter_plugin_tools` to reduce code maintenance, in turn there are some notable changes which are summarized below:
- command option changes:
    - option name change, `plugins` -> `packages`: Due to the implementation of `PluginCommand` in `flutter_plugin_tools`. This change doesn't affect `integration-test` at the moment because it's not migrated to dart yet.
    - `--run-on-changed-packages`: Due to the implementation of `PluginCommand` in `flutter_plugin_tools`, if no packages have changed, the option makes command run on all packages. This makes sense since most changes outside packages are CI scripts and tool changes which may affect all packages. This change doesn't affect `integration-test` at the moment because it's not migrated to dart yet.
    - ~~option `--run-on-changed-packages` can't be used together with option `packages`(previously `plugins`): This will make implementating `integration-test` command in dart with the current `--recipe` logic(https://github.com/flutter-tizen/plugins/pull/293#discussion_r763723894) difficult so the behavior of recipe may change.~~
- subcommand name changes:
    - `build` -> `build-examples` (to be consistent with `flutter_plugin_tools`)
    - `plugins` -> `list` (using command `ListCommand` from `flutter_plugin_tools`)
    - `test` -> `integration-test` (to differentiate from unit test which may be added in the future)
    - `tidy` -> `format` (using command `FormatCommand` from `flutter_plugin tools`)
- subcommand behavior changes:
    - `format` (previously `tidy`): formats dart, java, objective-c, and c++. Internally it uses `clang-format --style=Google`, @bbrto21 any reasons for specifically using `clang-format-11` with `--style=file`? `clang-format`(clang-format-10 in ubuntu 20.04) and `clang-format-11` generate the same rules for `--style=Google`. For dart, it runs `flutter format` which is a wrapper of `dart format`.


HakkyuKim:
Test will pass after #322.

swift-kim:
> any reasons for specifically using `clang-format-11` with `--style=file`? `clang-format`(clang-format-10 in ubuntu 20.04) and `clang-format-11` generate the same rules for `--style=Google`.

As far as I remember, clang-format-11 and clang-format-12 had different formatting rules when we tested before. I'm not sure about the clang-format-10 case. We just wanted to make sure every user uses the same clang-format version. (The version is also specified in https://github.com/flutter-tizen/flutter-tizen/wiki/Style-guide for this reason.) You can change the version if you need.

bbrto21 (approved):

.clang-format
```diff
@@ -1,183 +0,0 @@
-# This file is generated by the following command
```
swift-kim, line 1: Don't delete this file. It helps users format their code especially when they're using VS Code (otherwise they have to explicitly specify a "default" format style in the settings). You can make the file simpler like https://github.com/flutter-tizen/flutter-tizen/blob/master/.clang-format.

tools/lib/src/integration_test_command.dart
```diff
@@ -0,0 +1,91 @@
...
+    );
+    argParser.addOption(_recipeArg,
+        help:
+            'The recipe file path. A recipe refers to a yaml file that defines\n'
+            'a list of target platforms to test for each plugin.\n\n'
+            'Pass this file if you want to select specific target platforms\n'
+            'for different plugins. Every package listed in the recipe file\n'
+            'will be recognized by the tool (same as $_packagesArg option)\n'
+            'and those that specify an empty list will be explicitly excluded\n'
+            '(same as $_excludeArg option). If $_recipeArg is used,\n'
```
swift-kim, line 39: Not listing a package in the recipe file has the same effect as explicitly providing an empty list?

tools/lib/src/integration_test_command.dart
```diff
@@ -0,0 +1,91 @@
...
+            '$_packagesArg and $_excludeArg options will be ignored.\n\n'
+            'plugins:\n'
+            '  a: [wearable-5.5, tv-6.0]\n'
+            '  b: [mobile-6.0]\n'
+            '  c: [wearable-4.0]\n'
+            '  d: [] # explicitly excluded\n',
+        valueHelp: 'recipe.yaml');
```
swift-kim, line 46:
```suggestion
        valueHelp: 'recipe.yaml',);
```

tools/lib/src/integration_test_command.dart
```diff
@@ -0,0 +1,91 @@
...
+      .childFile('run_command.py');
+
+  @override
+  Future<void> run() async {
+    if (!_pythonTool.existsSync()) {
+      print('Error: Cannot find ${_pythonTool.path}.');
+      throw ToolExit(1);
+    }
+
+    // (TODO: HakkyuKim) Migrate python tool logic to dart.
```
swift-kim, line 84:
```dart
// TODO(username): Converting color to RGB because class Color doesn't support
//                 hex yet. See http://link/to/a/bug/123
```
is the standard Flutter TODO format.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-checking-in-comments-that-ask-questions

tools/lib/src/main.dart
```diff
@@ -28,19 +32,15 @@ void main(List<String> args) {
...
-      .childDirectory('tools')
-      .childDirectory('tools')
-      .childFile('run_command.py');
-
-  if (!pythonTool.existsSync()) {
-    print('Error: Cannot find ${pythonTool.path}.');
-    io.exit(1);
-  }
+  final CommandRunner<void> commandRunner = CommandRunner<void>(
+      'pub global run flutter_tizen_plugin tools',
```
swift-kim, line 36: This package is not published so this executable name is somewhat misleading.
```suggestion
      'dart run script/tool/bin/flutter_plugin_tools.dart',
```

tools/lib/src/build_examples_command.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: implementation_imports
```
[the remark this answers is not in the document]
HakkyuKim: Nice, I'll do that.

tools/lib/src/build_examples_command.dart
```diff
@@ -0,0 +1,63 @@
...
+    bool builtSomething = false;
+    for (final RepositoryPackage example in package.getExamples()) {
+      final String packageName =
+          getRelativePosixPath(example.directory, from: packagesDir);
+
+      builtSomething = true;
+      // (TODO: HakkyuKim) Support different profiles.
+      final int exitCode = await processRunner.runAndStream('flutter-tizen',
+          <String>['build', 'tpk', '--device-profile', 'wearable', '-v'],
+          workingDir: example.directory);
```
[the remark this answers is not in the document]
HakkyuKim: I've added trailing comma here and other places based on:

> Typically this means that when everything would fit on one line, put it all on one line, otherwise, split it one item to a line.

tools/lib/src/integration_test_command.dart
```diff
@@ -0,0 +1,91 @@
...
+// found in the LICENSE file.
+
+// ignore_for_file: implementation_imports
+
+import 'package:file/file.dart';
+import 'package:flutter_plugin_tools/src/common/core.dart';
+import 'package:flutter_plugin_tools/src/common/plugin_command.dart';
+
+/// A command that runs integration test of plugin examples.
+class IntegrationTestCommand extends PluginCommand {
```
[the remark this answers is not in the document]
HakkyuKim: Yes, that's the plan.

tools/lib/src/integration_test_command.dart
HakkyuKim: Oh, my mistake.

tools/lib/src/main.dart
HakkyuKim: Great advice, I agree.

tools/lib/src/integration_test_command.dart
HakkyuKim: In this context "an excluded package" means that it was recognized as a test candidate but was marked/logged as "excluded". On the other hand, packages not listed in the recipe file are not recognized as test candidates. I'll write a clearer description.

swift-kim: Ah, that's good. The current description seems enough.

tools/lib/src/integration_test_command.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// Creates an instance of the integration-test command, which runs the
+  /// integration test of each plugin example on physical targets or emulators.
+  IntegrationTestCommand(Directory packagesDir) : super(packagesDir) {
+    argParser.addFlag(
+      _generateEmulatorsArg,
+      help: 'Create and destroy emulators during test.\n\n'
+          'Must provide either $_platformsArg or $_recipeArg option to specify\n'
+          'which platforms to create.',
```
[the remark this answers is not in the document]
swift-kim: Not just here.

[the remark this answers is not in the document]
HakkyuKim: I've followed the style of [PluginCommand](https://github.com/flutter/plugins/blob/ec80d3439adb860d5f44c2d8427d459b10cda9d0/script/tool/lib/src/common/plugin_command.dart#L74-L93) since those messages will be shown together with ones in the IntegrationTestCommand. Also the output format is ugly when newlines are removed (Texts aren't wrapped well like the flutter tool).

flutter_tizen_plugin_tools, newlines removed:
flutter_tizen_plugins_tools

flutter:
flutter tools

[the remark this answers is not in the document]
swift-kim: It depends on the user's terminal size. I actually see no problem with the output in the first image.

One thing you could do is to change the help value of the platforms option (`<device_profile>-<platform_version>`) to something like just `profile-version` (without brackets).

I also tried setting the `usageLineLength` argument of `CommandRunner` but it didn't work.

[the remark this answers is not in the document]
HakkyuKim: >I also tried setting the usageLineLength argument of CommandRunner but it didn't work.

I also tried this but it didn't seem to work as I expected. 

I'll remove the newline characters.

swift-kim (approved):

PR #322, opened 2022-01-05
[tizen_app_control] Update messageport_tizen dependency in example to fix build error

HakkyuKim:
Upgrade `messageport_tizen` to 0.2.0 in `tizen_app_control` example. If major version starts with 0, the next minor version is not considered a compatiable release so `pub get` will not retrieve 0.2.0 for `^0.1.0`.

Fixes:
```
Exception: The project type of messageport_tizen is "sharedLib" which is not supported.
           Make sure the package is up to date by running "flutter-tizen pub upgrade".

           If you are the maintainer of the messageport_tizen package, consider migrating the project to "staticLib" type.
           Otherwise, you may modify the value of "type" in /home/hakkyu/.pub-cache/hosted/pub.dartlang.org/messageport_tizen-0.1.0/tizen/project_def.prop to temporarily fix the problem.
```

### DIFF
--- a/packages/tizen_app_control/CHANGELOG.md
+++ b/packages/tizen_app_control/CHANGELOG.md
@@ -5,3 +5,4 @@
 ## 0.1.1
 
 * Update messageport_tizen to 0.2.0 in example.
+* Replace deprecated members in example app.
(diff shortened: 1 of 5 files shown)


swift-kim (approved):
Thanks for the fix!

PR #324, opened 2022-01-25
[tizen_audio_manager] Add tizen_audio_manager plugin

m-bialek:
Migrate audio_manager plugin from internal repository. The plugin allows getting and setting the volume level for a given media type, getting current playback type and listening to volume changes.


swift-kim:
Our package naming convention is to add the `_tizen` suffix to Tizen implementation of already existing plugins (e.g. `audioplayers`→`audioplayers_tizen`) and add the `tizen_` prefix to the API name if the package is a Dart wrapper of Tizen APIs (e.g. `tizen_app_control`, `tizen_notification`). How about renaming your package to `tizen_sound_manager`?

Also do you have a reason for moving this package from the internal repo to this repo?

m-bialek:
@swift-kim I have renamed the package. We were asked by Jeong-Kyun Pu to migrate some plugins from internal repository.

HakkyuKim:
I have a question about the migration. I can think of three ways to move internal repos:
1. Migrating as a Tizen dedicated plugin(Dart Wrapper of Tizen APIs).
2. Finding another app-facing package where Tizen implementation can contribute to. (In the case of managing volumes, I've looked some up but haven't found a complete one, there seems to be some thrown-away or developing projects.)
3. Creating a new federated package. For example, "managing different type of volume" may also be required in other platforms. We can create a federated package similar to [device_info_plus](player_set_volume).

What is the migration strategy of internal repos?

swift-kim:
@m-bialek The underlying Tizen API implemented by this plugin is [Sound Manager](https://docs.tizen.org/application/native/guides/multimedia/sound-manager). Do you have a specific reason to use the term "audio manager"?

Also please rename the directory `audio_manager` to `tizen_audio_manager`, according to https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline#structure.

swift-kim:
@HakkyuKim Most apps do not need to control the device volume by themselves so there doesn't seem to be a standardized way of doing it in a Flutter app. None of existing plugins on pub.dev that support the feature have more than 100 likes.

@m-bialek @jkpu I doubt if 3rd-party apps will ever have to use this plugin. In which scenarios do you think this plugin will be useful?

m-bialek:
@swift-kim The plugin was created to match [Tizen .NET api](https://docs.tizen.org/application/dotnet/api/TizenFX/latest/api/Tizen.Multimedia.AudioManager.html). Therefore the name is "audio manager".

packages/tizen_audio_manager/lib/tizen_audio_manager.dart
```diff
@@ -0,0 +1,157 @@
+// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'package:flutter/services.dart';
```
swift-kim, line 6: nit
```suggestion
import 'dart:async';

import 'package:flutter/services.dart';
```

swift-kim (commented):
The overall API design (C# to Dart port) looks very cool. I'll look into the implementation details soon.

HakkyuKim:
Migrating internal plugins as their original form(Dart wrapper of Tizen APIs) seems best at the moment, I'll review upcoming migrations with this in mind.

jkpu:
> @m-bialek @jkpu I doubt if 3rd-party apps will ever have to use this plugin. In which scenarios do you think this plugin will be useful?

I requested this task after I requested by hasik jun and juwon ahn.
 We are going to discuss the necessity of this migration.

packages/tizen_audio_manager/example/lib/volume_control.dart
```diff
@@ -0,0 +1,123 @@
...
+                onChanged: _onAudioTypeChanged)
+          ]),
+          Text('Volume: $_currentVolume/$_maxVolume'),
+          SizedBox(
+              width: 250,
+              child: Slider(
+                value: _currentVolume.toDouble(),
+                min: 0,
+                max: _maxVolume.toDouble(),
+                onChanged: _onVolumeSliderChanged,
```
swift-kim, line 111: You can use the `divisions` argument of [`Slider`](https://api.flutter.dev/flutter/material/Slider-class.html) to make the animation smoother.

packages/tizen_audio_manager/example/lib/volume_control.dart
```diff
@@ -0,0 +1,123 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(title: const Text('Volume control')),
+      body: Column(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: <Widget>[
+          Text('Current playback type: ' +
+              _currentPlaybackType.toString().split('.').last),
```
swift-kim, line 95: Prefer adjacent string concatenation.
```suggestion
          Text('Current playback type: ${_currentPlaybackType.name}'),
```

packages/tizen_audio_manager/example/lib/volume_control.dart
```diff
@@ -0,0 +1,123 @@
...
+          Row(mainAxisAlignment: MainAxisAlignment.center, children: <Widget>[
+            const Text('Audio type:'),
+            const SizedBox(width: 10),
+            DropdownButton<AudioVolumeType>(
+                value: _selectedType,
+                items: _dropdownButtonItems,
+                onChanged: _onAudioTypeChanged)
+          ]),
```
swift-kim, line 103:
```suggestion
                onChanged: _onAudioTypeChanged,)
          ],),
```

packages/tizen_audio_manager/example/lib/volume_control.dart
```diff
@@ -0,0 +1,123 @@
...
+                min: 0,
+                max: _maxVolume.toDouble(),
+                onChanged: _onVolumeSliderChanged,
+              )),
+          Text(_volumeChangedEvent != null
+              ? 'Volume for ' +
+                  _volumeChangedEvent!.type.toString().split('.').last +
+                  ' has changed to ' +
+                  _volumeChangedEvent!.level.toString()
+              : '')
```
swift-kim, line 118: How about
```suggestion
          if (_volumeChangedEvent == null)
            const Text('')
          else
            Text(
              'Volume for ${_volumeChangedEvent!.type.name} '
              'has changed to ${_volumeChangedEvent!.level}.',
            ),
```

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    if (!ConvertStringToSoundType(type_str, type)) {
+      LOG_ERROR("Failed to convert string to sound type.");
+      result->Error("UnknownSoundType",
+                    "Failed to convert string to sound type.");
+      return;
+    }
+
+    const int volume = std::atoi(volume_str.c_str());
```
swift-kim, line 216: nit
```suggestion
    const int volume = std::stoi(volume_str);
```

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -0,0 +1,321 @@
...
+
+  static void VolumeChangedCb(sound_type_e type, unsigned int volume,
+                              void *user_data) {
+    auto plugin = static_cast<AudioManagerTizenPlugin *>(user_data);
+    std::string type_str;
+
+    if (ConvertSoundTypeToString(type, type_str)) {
+      flutter::EncodableMap msg;
+      msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+          "type", type_str));
```
swift-kim, line 301: This could be just

```suggestion
      msg[flutter::EncodableValue("type")] = flutter::EncodableValue(type_str);
```

packages/tizen_audio_manager/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
...
+  flutter: ">=2.5.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+
+dev_dependencies:
+  flutter_lints: ^1.0.0
+  flutter_test:
+    sdk: flutter
```
swift-kim, line 18: Please remove this unnecessary dependency.

packages/tizen_audio_manager/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
+name: tizen_audio_manager
+description: A Flutter plugin that allows setting and getting volume level for different audio types as well as getting current playback type.
+homepage: https://github.com/flutter-tizen/plugins
+repository: https://github.com/flutter-tizen/plugins/tree/master/packages/audio_manager
```
swift-kim, line 4:
```suggestion
repository: https://github.com/flutter-tizen/plugins/tree/master/packages/tizen_audio_manager
```

swift-kim (commented):
Also please take a look at https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline#miscellaneous and add this package to the repo's README and `labeler.yml`/`recipe.yaml`.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -0,0 +1,321 @@
...
+
+#include <memory>
+#include <string>
+
+#include "log.h"
+
+namespace {
+
+constexpr std::array<std::pair<std::string_view, sound_type_e>, 8>
+    sound_type_string_map{{{"system", SOUND_TYPE_SYSTEM},
```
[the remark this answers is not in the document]
m-bialek: You are right, a map would be better. However, I am not sure if using one here would be ok according to the style guide (https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables). What do you think?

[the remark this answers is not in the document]
swift-kim: Oh, I haven't noticed the rule. You could do as you prefer.

swift-kim (approved):

PR #326, opened 2022-02-02
[tizen_log] Add tizen_log plugin

m-bialek:
Migrate tizen_log plugin which allows to use Tizen dlog service. I think this plugin will make development easier as it is possible to set own tag. It will also make logging more consistent.


packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,160 @@
...
+  static final _DLog _dlogPrint =
+      _library.lookup<NativeFunction<_DLogNative>>('dlog_print').asFunction();
+  static final RegExp _stackTraceRegExp =
+      RegExp(r'^#(\d+)\s+(.+)\((.+):(\d+):(\d+)\)$', multiLine: true);
+
+  /// For displaying a file name, function name and line number in the logs,
+  /// use --dart-define=DEBUG_MODE=debug flag in build time.
+  /// flutter-tizen run --dart-define=DEBUG_MODE=debug
+  static const String debugMode =
+      String.fromEnvironment('DEBUG_MODE', defaultValue: 'no');
```
swift-kim, line 28: Why is the expected value "debug", but not something like "true" or "enabled"?

Also you could make this property [private](https://dart.dev/guides/language/effective-dart/design#prefer-making-declarations-private) or even merge with `isDebugEnabled`.

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,160 @@
...
+        final _StackFrame frame = _stackTrace().firstWhere(
+            (_StackFrame e) => e.index == frameIndex,
+            orElse: () => _StackFrame(0, '', '', 0, 0));
+        if (frame.index == frameIndex) {
+          file ??= frame.file;
+          func ??= frame.function;
+          line ??= frame.line;
+        }
+      }
+      message = '${file ?? "-"}: ${func ?? "-"}(${line ?? "-"}) > ' + message;
```
swift-kim, line 103: nit: Please always prefer [adjacent string concatenation](https://dart.dev/guides/language/effective-dart/usage#do-use-adjacent-strings-to-concatenate-string-literals).
```suggestion
      message = '${file ?? "-"}: ${func ?? "-"}(${line ?? "-"}) > $message';
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,160 @@
...
+    malloc.free(messagePtr);
+  }
+
+  static List<_StackFrame> _stackTrace() {
+    final List<_StackFrame> frames = <_StackFrame>[];
+    try {
+      final Iterable<RegExpMatch> matches =
+          _stackTraceRegExp.allMatches(StackTrace.current.toString());
+      for (final RegExpMatch m in matches) {
+        final List<String?> g = m.groups(<int>[1, 2, 3, 4, 5]);
```
swift-kim, line 120: nit: Please avoid abbreviations.
```suggestion
      for (final RegExpMatch match in matches) {
        final List<String?> groups = m.groups(<int>[1, 2, 3, 4, 5]);
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,160 @@
...
+    final Pointer<Utf8> tagPtr = tag.toNativeUtf8();
+    final Pointer<Utf8> formatPtr = '%s'.toNativeUtf8();
+    final Pointer<Utf8> messagePtr = message.toNativeUtf8();
+    _dlogPrint(priority.value, tagPtr, formatPtr, messagePtr);
+    malloc.free(tagPtr);
+    malloc.free(formatPtr);
+    malloc.free(messagePtr);
+  }
+
+  static List<_StackFrame> _stackTrace() {
```
swift-kim, line 114:
```suggestion
  static List<_StackFrame> _getStackTrace() {
```
Also the stack trace can be very large so the better option would be (in terms of performance)
```suggestion
  static _StackFrame? _getStackFrameAt(int index) {
```

README.md
```diff
@@ -65,6 +66,7 @@ The _"non-endorsed"_ status means that the plugin is not endorsed by the origina
 | [**shared_preferences_tizen**](packages/shared_preferences) | ✔️ | ✔️ | ✔️ | ✔️ |
 | [**sqflite_tizen**](packages/sqflite) | ✔️ | ✔️ | ✔️ | ✔️ |
 | [**tizen_app_control**](packages/tizen_app_control) | ✔️ | ✔️ | ✔️ | ✔️ |
+| [**tizen_log**](packages/tizen_log) | ✔️ | ✔️ | ✔️ | ✔️ |
```
swift-kim, line 69: I don't think this plugin is compatible with TV devices (including TV emulator).

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,160 @@
...
+import 'dart:ffi';
+import 'package:ffi/ffi.dart';
+import 'package:flutter/foundation.dart';
+
+typedef _DLogNative = Void Function(
+    Int32, Pointer<Utf8>, Pointer<Utf8>, Pointer<Utf8>);
+typedef _DLog = void Function(int, Pointer<Utf8>, Pointer<Utf8>, Pointer<Utf8>);
+
+/// Provides the ability to use Tizen dlog logging service.
+/// More info: https://docs.tizen.org/application/native/guides/error/system-logs/
```
HakkyuKim, line 14: The linked page only explains four log levels; debug, info, warn, and error. It's also a native API guide document where most content is irrelevant to Flutter developers. We should provide some guidance of each log level in the public doc comment for better developer experience. I have the following distinctions in mind but please share other opinions @flutter-tizen/maintainers:
- verbose: log messages of very detailed information for development.
- debug: log messages for development
- info: log messages for administration, typically used to report progress of the application.
- warn: to indicate problems that the program can tolerate but should be resolved whenever possible.
- error: to indicate problems that disturbs the normal workflow of the application, such as functional or performance limitations.
- fatal: to indicate problems that entirely blocks the normal workflow of the application.

The difference between verbose and debug logs seems to be in the "level of detail", since debug logs are always shown together with verbose logs but verbose logs can be filtered when priority filter is set to "debug".

references:
- https://developer.tizen.org/dev-guide/training/native-app/en/wearable/lesson_13/index.html
- https://stackoverflow.com/questions/25457858/android-log-levels

README.md
m-bialek: I found [this instruction](https://developer.samsung.com/smarttv/develop/extension-libraries/smart-view-sdk/receiver-apps/debugging.html) to enable debugging on a real TV. Unfortunately, I am not able to check if it gives an access to logs. On an emulator I cannot get logs indeed.

swift-kim: I'm pretty sure that there's no way to get dlogs from TV devices (unless you're using a development board). Web apps use their own logging mechanism. .NET apps have no logging support on TV.

packages/tizen_log/lib/tizen_log.dart
m-bialek: I added description of each priority in the README and as comments for functions. I also replaced the link. What do you think?

m-bialek:
It seems dart_analyze check fails because of sqflite plugin:
```console
info - sqflite/example/lib/open_test_page.dart:687:7 - Prefer const over final for declarations. - prefer_const_declarations
```

swift-kim:
@m-bialek The analysis is failing probably due the release of Dart 2.16. We'll fix it.

packages/tizen_log/lib/tizen_log.dart
swift-kim: The log levels are already quite straightforward for most of developers to understand. I don't think we need to state the info in README.

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+  static void debug(String tag, String message,
+      {String? file, String? func, int? line}) {
+    _log(_LogPriority.debug, tag, message, file: file, func: func, line: line);
+  }
+
+  /// Sends log with INFO priority and tag. Info log messages are
+  /// for administration, typically used to report progress of the application.
+  ///
+  /// [file], [func] and [line] parameters are set automatically if debug mode
+  /// is enabled, but they can be explicitly overridden.
```
HakkyuKim, line 58: Ditto.
```suggestion
  /// Sends log with INFO priority and tag.
  ///
  /// Info log messages are for administration, typically used to report
  /// progress of the application.
  ///
  /// [file], [func] and [line] parameters are set automatically if debug mode
  /// is enabled, but they can be explicitly overridden.
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+      {String? file, String? func, int? line}) {
+    _log(_LogPriority.info, tag, message, file: file, func: func, line: line);
+  }
+
+  /// Sends log with WARN priority and tag. Warn log messages are supposed to
+  /// indicate problems that the program can tolerate, but should be resolved
+  /// whenever possible.
+  ///
+  /// [file], [func] and [line] parameters are set automatically if debug mode
+  /// is enabled, but they can be explicitly overridden.
```
HakkyuKim, line 69: Ditto.
```suggestion
  /// Sends log with WARN priority and tag.
  ///
  /// Warn log messages indicate problems that the program can tolerate, but
  /// should be resolved whenever possible.
  ///
  /// [file], [func] and [line] parameters are set automatically if debug mode
  /// is enabled, but they can be explicitly overridden.
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+      {String? file, String? func, int? line}) {
+    _log(_LogPriority.warn, tag, message, file: file, func: func, line: line);
+  }
+
+  /// Sends log with ERROR priority and tag. Error log messages are supposed
+  /// to indicate problems that disturb the normal workflow of the application,
+  /// such as functional or performance limitations.
+  ///
+  /// [file], [func] and [line] parameters are set automatically if debug mode
+  /// is enabled, but they can be explicitly overridden.
```
HakkyuKim, line 80:
```suggestion
  /// Sends log with ERROR priority and tag.
  ///
  /// Error log messages indicate problems that disturb the normal workflow of
  /// the application, such as functional or performance limitations.
  ///
  /// [file], [func] and [line] parameters are set automatically if debug mode
  /// is enabled, but they can be explicitly overridden.
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+      {String? file, String? func, int? line}) {
+    _log(_LogPriority.error, tag, message, file: file, func: func, line: line);
+  }
+
+  /// Sends log with FATAL priority and tag. Fatal log messages are supposed
+  /// to indicate problems that entirely block the normal workflow
+  /// of the application.
+  ///
+  /// [file], [func] and [line] parameters are set automatically if debug mode
+  /// is enabled, but they can be explicitly overridden.
```
HakkyuKim, line 91:
```suggestion
  /// Sends log with FATAL priority and tag.
  ///
  /// Fatal log messages indicate problems that entirely block the normal
  /// workflow of the application.
  ///
  /// [file], [func] and [line] parameters are set automatically if debug mode
  /// is enabled, but they can be explicitly overridden.
```

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+typedef _DlogNative = Void Function(
+    Int32, Pointer<Utf8>, Pointer<Utf8>, Pointer<Utf8>);
+typedef _Dlog = void Function(int, Pointer<Utf8>, Pointer<Utf8>, Pointer<Utf8>);
+
+/// Provides the ability to use Tizen dlog logging service.
+/// More info: https://developer.tizen.org/dev-guide/training/native-app/en/wearable/lesson_13/index.html
+class Log {
+  Log._();
+
+  static final DynamicLibrary _library = DynamicLibrary.open('liblog.so');
```
swift-kim, line 19:
```
══╡ EXCEPTION CAUGHT BY GESTURE ╞═══════════════════════════════════════════════════════════════════
The following ArgumentError was thrown while handling a gesture:
Invalid argument(s): Failed to load dynamic library 'liblog.so': liblog.so: cannot open shared
object file: No such file or directory

When the exception was thrown, this was the stack:
#0      _open (dart:ffi-patch/ffi_dynamic_library_patch.dart:12:43)
#1      new DynamicLibrary.open (dart:ffi-patch/ffi_dynamic_library_patch.dart:23:12)
#2      Log._library (package:tizen_log/tizen_log.dart:19:57)
#3      Log._library (package:tizen_log/tizen_log.dart)
#4      Log._dlogPrint (package:tizen_log/tizen_log.dart:21:7)
#5      Log._dlogPrint (package:tizen_log/tizen_log.dart)
...
Handler: "onTap"
Recognizer:
  TapGestureRecognizer#dff4e
════════════════════════════════════════════════════════════════════════════════════════════════════
```

Did you mean
```suggestion
  static final DynamicLibrary _library = DynamicLibrary.open('libdlog.so.0');
```

packages/tizen_log/lib/tizen_log.dart
HakkyuKim: @m-bialek Thanks for the fix, I've added some following suggestions below. I don't mind removing it from README as long as it's stated in the doc comments.

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,166 @@
...
+        final List<String?> groups = match.groups(<int>[1, 2, 3, 4, 5]);
+        if (groups.any((String? group) => group == null) == false) {
+          final int frameIndex = int.parse(groups[0]!);
+          if (frameIndex == index) {
+            return _StackFrame(
+              frameIndex,
+              groups[1]!.trim(),
+              groups[2]!.trim(),
+              int.parse(groups[3]!),
+              int.parse(groups[4]!),
```
swift-kim, line 133: Can we print only file names and function names as `Tizen.Log` ([TizenFX](https://samsung.github.io/TizenFX/stable/api/Tizen.Log.html)) does?

```
I/ConsoleMessage( 6889): Utils.cs: ParseEngineArgs(30) > Enabled: --enable-dart-profiling
I/ConsoleMessage( 6889): Utils.cs: ParseEngineArgs(30) > Enabled: --enable-checked-mode

W/tizen_log_example( 7304): package:tizen_log_example/main.dart: MyApp.build.<anonymous closure>(48) > warn message
V/tizen_log_example( 7304): package:tizen_log_example/main.dart: MyApp.build.<anonymous closure>(36) > verbose message
```

packages/tizen_log/lib/tizen_log.dart
swift-kim: Also, providing a link to the official [native API documentation](https://docs.tizen.org/application/native/guides/error/system-logs) just makes sense because this plugin is a wrapper around the native API.

HakkyuKim: Isn't that implementation detail? I agree if the link is not written in doc comments. Users just need to know how to use the API.

swift-kim: Doc comments can provide any information regarding the API including its implementation details. Also the document is not just about the native API but provides general information about the dlog logging service and the dlogutil utility. Of course it would be better if there's an official dedicated document (end-to-end guide) for Dart users.

HakkyuKim: > Doc comments can provide any information regarding the API including its implementation details.

It can if it's necessary but I didn't find any reasons why users need or want to read about native API but I agree that dlog logging service and dlogutil utility are required information. 

If we can't avoid linking the document, at least it should say something like:
```
/// [Log] is a wrapper of Tizen's native API, more info: 
/// https://docs.tizen.org/application/native/guides/error/system-logs/.
```

However if this plugin is only intended for inside developers who already know about how Tizen APIs are structured then I'm good with the original comment.

swift-kim: A "plugin" (by its definition) implicitly indicates that it's wrapping some underlying native API. A user has no idea about what dlog is and what it's intended for may want to learn what they can do with the API, by reading any related documentation (the native API guide in this case) if there's no Dart counterpart of it. Can you even guess where you can see the dlog output if there's only a few lines of doc comments without the link?

HakkyuKim: My intention was to avoid the link if it's possible or guide why the link is there. Reading about native API(or any C code examples) is still unnecessary even if the user doesn't know about dlog. The reason why the link should be in the comment in this case is because of dlog logging service and dlogutil utility, so I agree adding the link but suggested that we write more than "More info: \<link\>" when we're providing much more information than the user want.

packages/tizen_log/lib/tizen_log.dart
m-bialek: For example only `main.dart` instead of `package:tizen_log_example/main.dart` and only `<anonymous closure>` instead of `MyApp.build.<anonymous closure>`?

packages/tizen_log/lib/tizen_log.dart
m-bialek: Thank you for noticing this, I wrongly assumed `liblog.so` is present on all devices.

packages/tizen_log/lib/tizen_log.dart
swift-kim: Yes exactly. That seems to be the standard logging format of all Tizen modules. But we can do it later if the work will take some time.

packages/tizen_log/lib/tizen_log.dart
swift-kim: Could you answer this question? Will there be any other mode than "debug"?

HakkyuKim (commented):
@m-bialek I'm OK with the latest changes, you can also remove log priority explanations in `README.md` as @swift-kim suggested. Thanks for the fix.

HakkyuKim:
> @m-bialek The analysis is failing probably due the release of Dart 2.16. We'll fix it.

@m-bialek  CI will pass if you apply #327.

HakkyuKim (approved):

packages/tizen_log/lib/tizen_log.dart
swift-kim: Reminding again. @m-bialek

m-bialek: As far as I can see, I merged this property with `isDebugEnabled` as you suggested but I forgot to reply to this comment. I am sorry for that.

m-bialek: I copied this solution from another project. I do not think there will be other modes than debug.

m-bialek: I changed String to bool. Now the user can write `--dart-define=DEBUG_MODE=true` to enable debug mode.

packages/tizen_log/lib/tizen_log.dart
```diff
@@ -0,0 +1,183 @@
...
+/// information see:
+/// https://docs.tizen.org/application/native/guides/error/system-logs/
+class Log {
+  Log._();
+
+  static final DynamicLibrary _library = DynamicLibrary.open('libdlog.so.0');
+  static final _Dlog _dlogPrint =
+      _library.lookup<NativeFunction<_DlogNative>>('dlog_print').asFunction();
+  static final RegExp _stackTraceRegExp =
+      RegExp(r'^#(\d+)\s+(.+)\((.+):(\d+):(\d+)\)$', multiLine: true);
```
swift-kim, line 29: I just found that `DEBUG_MODE` doesn't work when the app is running in release mode because the stack trace format is different (the "column number" part is missing in release mode). I think you don't need to parse that part.

Debug mode:
```
#0      Log._getStackFrameAt (package:tizen_log/tizen_log.dart:134:22)
#1      Log._log (package:tizen_log/tizen_log.dart:115:36)
#2      Log.warn (package:tizen_log/tizen_log.dart:84:5)
#3      MyApp.build.<anonymous closure> (package:tizen_log_example/main.dart:48:38)
#4      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:989:21)
```

Output:
```sh
I/tizen_log_example(  624): main.dart: <anonymous closure>(44) > info message
W/tizen_log_example(  624): main.dart: <anonymous closure>(48) > warn message
```

Release mode:
```
#0      Log._getStackFrameAt (package:tizen_log/tizen_log.dart:134)
#1      Log._log (package:tizen_log/tizen_log.dart:115)
#2      Log.warn (package:tizen_log/tizen_log.dart:84)
#3      MyApp.build.<anonymous closure> (package:tizen_log_example/main.dart:48)
#4      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:989)
```

Output:
```sh
I/tizen_log_example(32556): -: -(-) > info message
W/tizen_log_example(32556): -: -(-) > warn message
```

m-bialek: Thank you for catching that. I have modified the regex and removed `column` property from `_StackFrame` class. I checked and now it also works in release mode.

swift-kim: Thank you for the quick fix!

swift-kim (approved):

PR #336, opened 2022-03-08
[webview_flutter] Organize includes

swift-kim:
Add missing includes and remove unused includes.

The most important part is the added `Ecore_IMF.h` dependency of `webview.h`. The plugin is not compatible with the updated `flutter_platform_view.h` that I'm currently working on due to the missing include. All occurrences of 

```cpp
#include <flutter/standard_message_codec.h>
```

will be removed after the work is done because they are only required by `flutter_platform_view.h`.

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.11
+
+* Organize includes.
+
 ## 0.3.10
 
 * Apply `PlatformView` and `PlatformViewFactory` API changes.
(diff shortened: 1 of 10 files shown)


bwikbs (approved):
LGTM!
Quick question) Why are you removing ``standard_message_codec.h``?

seungsoo47:
I have a question. Why are you moving the includes declared in cpp files to header files?

swift-kim:
@bwikbs Because it's not used by this plugin (C++ implementation).

@seungsoo47 I'm not moving anything. I'm just adding them because they're missing. For example, since `flutter::PluginRegistrar` is referenced by `webview_factory.h`, the file must include the header `flutter/plugin_registrar.h`.

PR #337, opened 2022-03-14
[webview_flutter] Support emulator

bwikbs:
* Use the sw backend for emulator


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -137,6 +139,23 @@ bool GetValueFromEncodableMap(const flutter::EncodableValue& arguments,
   return false;
 }
 
+bool needsSWBackend(void) {
+  bool result = false;
+  char* value;
+  int ret;
```
bbrto21, line 145:
```suggestion
  char* value = nullptr;
  int ret = 0;
```

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -137,6 +139,23 @@ bool GetValueFromEncodableMap(const flutter::EncodableValue& arguments,
   return false;
 }
 
+bool needsSWBackend(void) {
```
bbrto21, line 142: recommend:
```suggestion
bool needsSwBackend(void) {
```

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -916,6 +936,47 @@ void WebView::HandleMethodCall(
...
+    }
+    result->Error("InvalidArguments", "Please set 'key' properly");
+  } else if (method_name.compare("loadHtmlString") == 0) {
+    std::string html;
+    std::string baseUrl;
+    if (!GetValueFromEncodableMap(arguments, "html", &html)) {
+      result->Error("InvalidArguments", "Please set 'html' properly");
+      return;
+    }
+    if (GetValueFromEncodableMap(arguments, "baseUrl", &baseUrl)) {
```
bbrto21, line 960: `baseUrl` is not used.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -137,6 +139,23 @@ bool GetValueFromEncodableMap(const flutter::EncodableValue& arguments,
...
+  bool result = false;
+  char* value;
+  int ret;
+  ret = system_info_get_platform_string("http://tizen.org/system/model_name",
+                                        &value);
+  if (ret == SYSTEM_INFO_ERROR_NONE) {
+    if (strcmp(value, "Emulator") == 0) {
+      result = true;
+    }
+  }
```
seungsoo47, line 151:
```suggestion
  if ((SYSTEM_INFO_ERROR_NONE == ret) && (0 == strcmp(value, "Emulator")) {
      result = true;
  }
```
How about reducing the number of nested if statements? And how about putting the value in front of the variable in the conditional statement?

packages/webview_flutter/tizen/src/webview.cc
bwikbs: You're right. lwe is currently unable to process it, so it only shows the log.

packages/webview_flutter/tizen/src/webview.cc
bwikbs: Good point!

packages/webview_flutter/tizen/src/webview.cc
bwikbs: Thanks! That looks good! 👍

seungsoo47 (approved):

bbrto21 (approved):

PR #338, opened 2022-03-16
Rename id to serial

HakkyuKim:
Use the term "serial" to be consistent with sdb.

### DIFF
--- a/tools/lib/src/device.dart
+++ b/tools/lib/src/device.dart
@@ -47,9 +47,9 @@ class Device {
       tizenSdk: tizenSdk,
       processRunner: processRunner,
     );
-    device._id = device._findId();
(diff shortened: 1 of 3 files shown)


swift-kim:
Is the CI expected to build all packages when only the tool code has changed?

HakkyuKim:
@swift-kim 
Yes, it's because of the [`--run-on-changed-packages` option in `PluginCommand`](https://github.com/flutter/plugins/blob/f25e370447f654a4480a8ee84635ab79cad53804/script/tool/lib/src/common/plugin_command.dart#L74-L80).

swift-kim (approved):

bwikbs (approved):
👍

Issue #339, opened 2022-03-16
[webview_flutter_lwe] Crash when navigating to specific URL on emulator


swift-kim:
Issue 1. App crash

1. Set the example app's `initialUrl` to either https://stackoverflow.com or https://map.kakao.com.
2. Run.

Issue 2. Visual artifacts (residual image)

1. Set the `initialUrl` to https://map.naver.com (or any other).
2. Run.
3. Once the webview finishes to load the site, terminate the app.
4. Set the `initialUrl` to https://www.google.com (or any other).
5. Run again.

![T-samsung-6 0-x86-2022-03-16-134414](https://user-images.githubusercontent.com/40190588/158519126-90fc0ed8-28ce-4659-a5b0-18e759b216e4.png)
![T-samsung-6 0-x86-2022-03-16-130658](https://user-images.githubusercontent.com/40190588/158519167-9354e44c-04c5-48df-9892-179012ea4398.png)

Issue 3. Images not displayed

1. Set the `initialUrl` to https://www.naver.com.
2. Run and observe that images on the site (ads and etc.) are not displayed properly.

![T-samsung-6 0-x86-2022-03-16-135038](https://user-images.githubusercontent.com/40190588/158519184-8cb158d1-ac05-4a82-b9f3-dde4a40b8d6a.png)

Issue 4. Output resolution

As you can see in the above screenshots, texts are too unclear to read.

@seungsoo47 @bwikbs @bbrto21

bwikbs:
@swift-kim 
Wow 😱  it's only been a few hours since it merge, but you've already tried it.

At a glance, some seems to be related plug-in, and the other seem to be lwe issues.
We will solve the plugin-related issues first and then solve the lwe later.

I'll let you know here whenever something gets resolved a little bit.

bbrto21:
> * Set the example app's `initialUrl` to either https://stackoverflow.com or https://map.kakao.com.

It doesn't seem to crash anymore. I just tested it with #456.

swift-kim:
@bbrto21 I tried the patch but the app still intermittently crashes on TV emulator when `initialUrl` is set to https://map.kakao.com.

bbrto21:
Oh, you're right... There seems to be another problem lurking :cry:

swift-kim:
The app also crashes during page loading when I click the "Visit dart.dev" button in the flutter.dev page. I tested on a mobile emulator.

PR #342, opened 2022-03-17
[webview_flutter] In the case of sw backend, use only one buffer for rendering

bwikbs:
This is a compromise of partial rendering for performance. A sync issue may occur in the sw backend.

This patch improves rendering which is related to issue 2 of https://github.com/flutter-tizen/plugins/issues/339

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -174,9 +174,9 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
       platform_window_(platform_window) {
   use_sw_backend_ = NeedsSwBackend();
   if (use_sw_backend_) {
-    tbm_pool_ = new SingleBufferPool(width, height);
+    tbm_pool_ = std::make_unique<SingleBufferPool>(width, height);
   } else {
-    tbm_pool_ = new BufferPool(width, height, BUFFER_POOL_SIZE);
+    tbm_pool_ = std::make_unique<BufferPool>(width, height, BUFFER_POOL_SIZE);
   }
 
   texture_variant_ = new flutter::TextureVariant(flutter::GpuBufferTexture(
@@ -405,11 +405,6 @@ void WebView::Dispose() {
   if (texture_variant_) {
     delete texture_variant_;
     texture_variant_ = nullptr;
-  }
-
-  if (tbm_pool_) {
-    delete tbm_pool_;
-    tbm_pool_ = nullptr;
   }
 }
 
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -19,6 +19,7 @@
 #include "lwe/PlatformIntegrationData.h"
 #include "webview_factory.h"
 
+#define BUFFER_POOL_SIZE 5
 #define LWE_EXPORT
 extern "C" size_t LWE_EXPORT createWebViewInstance(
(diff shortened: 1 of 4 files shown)


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -164,14 +165,20 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
...
-      rendered_surface_(nullptr),
       is_mouse_lbutton_down_(false),
+      rendered_surface_(nullptr),
       has_navigation_delegate_(false),
       has_progress_tracking_(false),
       context_(nullptr),
       texture_variant_(nullptr),
       platform_window_(platform_window) {
-  tbm_pool_ = std::make_unique<BufferPool>(width, height);
+  use_sw_backend_ = needsSwBackend();
```
seungsoo47, line 175:
```suggestion
  use_sw_backend_ = NeedsSwBackend();
```
Please change the 'n' to uppercase.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -19,6 +19,7 @@
 #include "lwe/PlatformIntegrationData.h"
 #include "webview_factory.h"
 
+#define BUFFER_POOL_SIZE 5
```
seungsoo47, line 22: Why did you declare it to 5?

bwikbs: It is not the value set this time, it is the value it already had.
There was no particular reason.
It was moderately selected for hands greater than 3. 
at multiful webviews, if frequent updates such as animations occur, buffer shortages may occur, so a moderately large value was chosen.

packages/webview_flutter/tizen/src/webview.cc
bwikbs: Good point! done!

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -164,14 +165,20 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int viewId,
...
       context_(nullptr),
       texture_variant_(nullptr),
       platform_window_(platform_window) {
-  tbm_pool_ = std::make_unique<BufferPool>(width, height);
+  use_sw_backend_ = NeedsSwBackend();
+  if (use_sw_backend_) {
+    tbm_pool_ = new SingleBufferPool(width, height);
+  } else {
+    tbm_pool_ = new BufferPool(width, height, BUFFER_POOL_SIZE);
+  }
```
bbrto21, line 180: Why don't you use [unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr) anymore?

``` c++
std::cout << "\n" "2) Runtime polymorphism demo\n";
{
    // Create a derived resource and point to it via base type
    std::unique_ptr<B> p = std::make_unique<D>();

    // Dynamic dispatch works as expected
    p->bar();
}
```

bwikbs: Well, there's no other reason. When I first did it, an error occurred and I just thought it wouldn't work.
I'll try it again.

bwikbs: done!

bbrto21 (approved):

packages/webview_flutter/tizen/src/webview.cc
seungsoo47: I understood. Thanks for the details. 👍

seungsoo47 (approved):

PR #343, opened 2022-03-24
[webview_flutter] Add extra dependent binary

bwikbs:
* This is for tizen 7.0


seungsoo47:
Could not we find the dependant libraries on the target platform? Should the all dependant libraries add to the webview plugin? As you can see, the webview plugin's footprint (memory size) becomes too large than before.

HakkyuKim:
Our self-hosted runners can't pick up integration test jobs for some reason, I will check this when I get back to office next week.

swift-kim:
1. Why is the current plugin not compatible with Tizen 7.0?
2. Are the new binaries compatible with Tizen 6.5 or older devices?
3. webview_flutter_tizen 0.4.0 needs publishing. It has not been published on pub.dev yet.

bwikbs:
> Why is the current plugin not compatible with Tizen 7.0?

``lwe`` requires additional library & currently webview plugin use the platform one.
These are not included in the lwe binary due to the license issue.
I think the change of toolchain has caused the this problem.
Because dependency on the platform used a new tool chain.

> Are the new binaries compatible with Tizen 6.5 or older devices?

It's tested on TM1(6.5), rb5(7.0), mobile emulator(6.5)
And I hope it will be caused serious issue because it has not changed toolchain only add extra dependecy.

> webview_flutter_tizen 0.4.0 needs publishing. It has not been published on pub.dev yet.

Thanks for letting me know. This is considered to release to 0.4.1 this time.

bwikbs:
> Could not we find the dependant libraries on the target platform? Should the all dependant libraries add to the webview plugin? As you can see, the webview plugin's footprint (memory size) becomes too large than before.

There are no resolutions in the current situation. (I guess..)
We may think about other lwe release soluation. But it means that we have to re-consider the problems we were troubled before.

swift-kim:
> ``lwe`` requires additional library & currently webview plugin use the platform one.
These are not included in the lwe binary due to the license issue.
I think the change of toolchain has caused the this problem.
Because dependency on the platform used a new tool chain.

I don't quite get what you mean. Why does LWE require additional dependencies on Tizen 7.0 only?

> Are the new binaries compatible with Tizen 6.5 or older devices?

Which toolchain did you use to build the new binaries? Tizen devices use different versions of libstdc++ depending on their platform versions and the untested binaries can possibly cause an ABI compatibility issue if you were not using the lowest possible toolchain version.

> This is considered to release to 0.4.1 this time.

What does this mean?

bwikbs:
> I don't quite get what you mean. Why does LWE require additional dependencies on Tizen 7.0 only?

> Which toolchain did you use to build the new binaries? Tizen devices use different versions of libstdc++ depending on their platform versions and the untested binaries can possibly cause an ABI compatibility issue if you were not using the lowest possible toolchain version.

I just realized that I was watching the wrong target. 😱  Forget what I talked above.
This issue is an issue which is caused by absense of lwe or some missing files in Tizen 7.0 device.
I still think PR is available. Because there is a device without lwe.

> What does this mean?

It just means webview will be published at this time.

bwikbs:
After internal discussion, we decided to add LWE to the platform. 
So this PR will be closed soon.

PR #344, opened 2022-03-25
[video_player] Apply new texture release callback

xiaowei-guan:

PR #346, opened 2022-03-26
[camera] Apply texture API change

bwikbs:
Related to https://github.com/flutter-tizen/engine/pull/259

### DIFF
--- a/packages/camera/tizen/src/camera_device.cc
+++ b/packages/camera/tizen/src/camera_device.cc
@@ -343,41 +343,26 @@ CameraDevice::CameraDevice(flutter::PluginRegistrar *registrar,
           [this](size_t width,
                  size_t height) -> const FlutterDesktopGpuBuffer * {
             std::lock_guard<std::mutex> lock(mutex_);
-            if (prepared_packet_ && !IsValidMediaPacket(prepared_packet_)) {
-              media_packet_destroy(prepared_packet_);
-              prepared_packet_ = nullptr;
-            }
-            if (current_packet_ && !IsValidMediaPacket(current_packet_)) {
-              media_packet_destroy(current_packet_);
-              current_packet_ = nullptr;
-            }
-            if (!prepared_packet_ && !current_packet_) {
+            if (!current_packet_) {
               return nullptr;
             }
-            if (prepared_packet_ && !current_packet_) {
-              current_packet_ = prepared_packet_;
-              prepared_packet_ = nullptr;
-            }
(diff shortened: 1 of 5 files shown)


HakkyuKim:
Integration test CI is fixed. I'll rerun the test.

bbrto21:
This PR should be rebased upon #344. because the example is dependent on the video plug-in.
but PR works fine.

packages/camera/tizen/src/camera_device.cc
```diff
@@ -343,41 +343,26 @@ CameraDevice::CameraDevice(flutter::PluginRegistrar *registrar,
...
-            if (prepared_packet_ && !current_packet_) {
-              current_packet_ = prepared_packet_;
-              prepared_packet_ = nullptr;
-            }
-
             tbm_surface_h surface = nullptr;
             int ret = media_packet_get_tbm_surface(current_packet_, &surface);
             if (ret != MEDIA_PACKET_ERROR_NONE) {
               LOG_ERROR("media_packet_get_tbm_surface failed, error: %d", ret);
-              media_packet_destroy(current_packet_);
```
bbrto21, line 366: Why do you remove this line? if getting a tbm surface from media packets is fails, shouldn't media packets be destroyed? The reference is lost on the very next line.

packages/camera/tizen/src/camera_device.cc
```diff
@@ -343,41 +343,26 @@ CameraDevice::CameraDevice(flutter::PluginRegistrar *registrar,
           [this](size_t width,
                  size_t height) -> const FlutterDesktopGpuBuffer * {
             std::lock_guard<std::mutex> lock(mutex_);
-            if (prepared_packet_ && !IsValidMediaPacket(prepared_packet_)) {
-              media_packet_destroy(prepared_packet_);
-              prepared_packet_ = nullptr;
-            }
-            if (current_packet_ && !IsValidMediaPacket(current_packet_)) {
```
bbrto21, line 350: Could you remove `IsValidMediaPacket` definition too.

bwikbs: @bbrto21 Oh. good point!

packages/camera/tizen/src/camera_device.cc
bwikbs: You're right. It seems to be a mistake! 👍

bbrto21 (approved):

PR #351, opened 2022-04-11
[battery_plus] Refactor the C++ code

swift-kim:

- Extract the `DeviceBattery` class from `BatteryPlusTizenPlugin` for better separation of the platform channel communication and the device API usage. (I referred to the [battery_plus_windows](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/battery_plus/battery_plus_windows/windows/include/battery_plus_windows/system_battery.h) implementation.)
  - Create an enum `BatteryStatus` to indicate the current status of the device battery or an error status.
  - `RegisterObserver`/`UnregisterObserver` → `StartListen`/`StopListen`
  - `StartListen` takes `BatteryStatusCallback` as an argument.
  - Create a new method `GetLevel`.
  - Let `GetStatus` (previously `GetBatteryStatus`) return a value of type `BatteryStatus`.
- Create typedefs of long commonly-used types (also taken from [battery_plus_windows](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/battery_plus/battery_plus_windows/windows/battery_plus_windows_plugin.cpp)) for improved readability. For example,<p>
  ```patch
  - std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> events_;
  + std::unique_ptr<FlEventSink> events_;
  ```
- Create an explicit `BatteryStatusStreamHandler` class. Now the lifecycle of `EventSink` is solely managed by this class.
  - `OnListenInternal` returns a `StreamHandlerError` on initialization failure.
- Minor cleanups.
  - Remove `SetupChannels` and combine with `RegisterWithRegistrar`.
  - Style log messages and error codes in a consistent manner.
  - Wrap the non-public implementation with an anonymous namespace.

Additional note: The `batteryState` getter API has been added the platform interface and will be implemented in a later PR.

### DIFF
--- a/packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
+++ b/packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
@@ -4,26 +4,102 @@
 
 #include "battery_plus_tizen_plugin.h"
 
-#include <device/battery.h>
-#include <device/callback.h>
 #include <flutter/event_channel.h>
 #include <flutter/event_sink.h>
-#include <flutter/event_stream_handler_functions.h>
+#include <flutter/event_stream_handler.h>
 #include <flutter/method_channel.h>
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
 
-#include <map>
+#include <memory>
 #include <string>
 
-#include "log.h"
+#include "device_battery.h"
+
+namespace {
+
+typedef flutter::EventChannel<flutter::EncodableValue> FlEventChannel;
+typedef flutter::EventSink<flutter::EncodableValue> FlEventSink;
+typedef flutter::MethodCall<flutter::EncodableValue> FlMethodCall;
+typedef flutter::MethodResult<flutter::EncodableValue> FlMethodResult;
+typedef flutter::MethodChannel<flutter::EncodableValue> FlMethodChannel;
+typedef flutter::StreamHandler<flutter::EncodableValue> FlStreamHandler;
+typedef flutter::StreamHandlerError<flutter::EncodableValue>
+    FlStreamHandlerError;
+
+std::string BatteryStatusToString(BatteryStatus status) {
+  switch (status) {
+    case BatteryStatus::kCharging:
+      return "charging";
+    case BatteryStatus::kFull:
+      return "full";
+    case BatteryStatus::kDischarging:
+      return "discharging";
+    case BatteryStatus::kUnknown:
+    default:
+      return "unknown";
+  }
+}
+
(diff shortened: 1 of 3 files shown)


packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
```diff
@@ -4,199 +4,131 @@
...
+            flutter::EncodableValue(BatteryStatusToString(status)));
+      } else {
+        events_->Error(std::to_string(battery_.GetLastError()),
+                       battery_.GetLastErrorString());
+      }
+    };
+    if (!battery_.StartListen(callback)) {
+      return std::make_unique<FlStreamHandlerError>(
+          std::to_string(battery_.GetLastError()),
+          battery_.GetLastErrorString(), nullptr);
```
swift-kim, line 63: This currently results in an app crash. There will be a fix in the engine code soon.

Issue: https://github.com/flutter/flutter/issues/101682

bbrto21 (approved):

packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc
bbrto21: @swift-kim Any progress on this?

swift-kim: @bbrto21 It isn't likely the fix will land very soon. How do you think of [this](https://github.com/flutter-tizen/plugins/pull/387#discussion_r883346794) or [this](https://github.com/flutter-tizen/plugins/pull/387/commits/50086e595cd1839ae000f05c46868a4eacc3faba) workaround?

bbrto21: @swift-kim Actually, I also made a patch to prevent crash. how about [this](https://github.com/bbrto21/plugins/commit/a351f1143a797ac3c5644e5b46d9c25d614df237)?
First one has intentional leak and the second solution you mentioned doesn't seem very good as it depends on instances of other classes.

swift-kim: The `holder_` suffix seems redundant. Otherwise looks good to me.

bbrto21: @swift-kim Can I make PR to apply all place where have same problem?

swift-kim: Sure. Just one thing I forgot to mention: the `private:` section header is missing in the class.

PR #352, opened 2022-04-11
[connectivity_plus] Refactor the C++ code

swift-kim:
Part of [the code refactoring project](https://github.com/flutter-tizen/plugins/issues/354). This change is an analogue of https://github.com/flutter-tizen/plugins/pull/351.

- Extract the `Connection` class from `ConnectivityPlusTizenPlugin` for better separation of the platform channel communication and the device API usage. (I referred to the [connectivity_plus_windows](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus_windows/windows/connectivity_plus_windows_plugin.cpp) implementation.)
  - Create an enum `ConnectionType` to indicate the current connectivity type or an error status.
  - `RegisterObserver`/`ClearObserver` → `StartListen`/`StopListen`
  - `StartListen` takes `ConnectionTypeCallback` as an argument.
  - Create a new method `GetType`.
  - Remove `EnsureConnectionHandle`. The connection handle is now created and managed by the `Connection` class.
- Create typedefs of long commonly-used types.
- Create an explicit `ConnectivityStreamHandler` class. Now the lifecycle of `EventSink` is solely managed by this class.
- Minor cleanups.
  - Remove `SetupChannels` and combine with `RegisterWithRegistrar`.
  - Style log messages and error codes in a consistent manner.
  - Wrap the non-public implementation with an anonymous namespace.

Additional note: The connectivity type "bluetooth" has been added to the platform interface and will be implemented in a later PR.

### DIFF
--- a/packages/connectivity_plus/tizen/src/connection.cc
+++ b/packages/connectivity_plus/tizen/src/connection.cc
@@ -1,0 +1,74 @@
+// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "connection.h"
+
+#include "log.h"
+
+static ConnectionType ToConnectionType(connection_type_e type) {
+  switch (type) {
+    case CONNECTION_TYPE_WIFI:
+      return ConnectionType::kWiFi;
+    case CONNECTION_TYPE_CELLULAR:
+      return ConnectionType::kMobile;
+    case CONNECTION_TYPE_ETHERNET:
+      return ConnectionType::kEthernet;
+    case CONNECTION_TYPE_DISCONNECTED:
+    default:
+      return ConnectionType::kNone;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


packages/connectivity_plus/tizen/src/connection.cc
```diff
@@ -0,0 +1,73 @@
...
+  int ret = connection_create(&connection_);
+  if (ret != CONNECTION_ERROR_NONE) {
+    LOG_ERROR("Failed to create handle: %s", get_error_message(ret));
+    last_error_ = ret;
+  }
+}
+
+Connection::~Connection() {
+  if (connection_) {
+    connection_destroy(connection_);
```
bbrto21, line 33: For stability, it is recommended to unset the callback before destroying the connection handle.

swift-kim: Fixed!

bbrto21 (approved):

Issue #354, opened 2022-04-12
Use consistent coding style across all packages


swift-kim:
This issue is for tracking the code refactoring status of this repo. The main goal is to improve

- Overall code quality
- Readability and maintainability
- Consistency

of all C++ code.

Here are some general rules for styling the code:

- Everything from the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).
- Structure:
  - If the plugin class (which defines the `RegisterWithRegistrar` method) is too large (over 150 lines), consider breaking it into multiple classes.
    - The classes should have a clean design so that their roles can be easily known from their signatures. Ideally, the plugin class should be the only one that interacts with the underlying platform channel (processes method arguments and generates results), and other classes should not touch any Flutter API (`flutter/*.h`).
  - Enclose the plugin class within an anonymous namespace.
  - If you're writing a Tizen C API wrapper, avoid exposing Tizen-defined types (including enums) in its public interface.
  - Minimize use of [output parameters](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) when writing a function that returns a value. Consider using the `GetLastError` pattern ([example](https://github.com/flutter-tizen/plugins/pull/351)) or defining your own result type such as `TizenResult`.
    - Consider making use of C++ exceptions if they make the code simpler and easier to understand (e.g. multiple platform API invocations per single method call). See [this comment](https://github.com/flutter-tizen/plugins/pull/357#issuecomment-1098934624) for other available options.
  - Consider creating an explicit stream handler class that extends `flutter::StreamHandler` if the plugin makes use of event channels. ([example](https://github.com/flutter-tizen/plugins/pull/351))
- Logging (`log.h`):
  - Avoid unnecessary logging. Minimize use of debug logs (`LOG_DEBUG`) in the final code.
  - Every log should provide useful information. Any information that is already known to the caller (through a `PlatformException` object) is redundant. Instead, provide some additional context about the error so that the developer can easily debug the underlying issue.
- Method call handling:
  - Consider using [`GetValueFromEncodableMap`](https://github.com/flutter-tizen/flutter-tizen/blob/master/doc/develop-plugin.md#2-add-tizen-platform-code-cc) to extract arguments from an `EncodableMap`.
  - Prefer `==` over `.compare()` when comparing strings.<p>
    ```cpp
    const auto &method_name = method_call.method_name();

    if (method_name == "wifiName") {
      ...
    }
    ```
[...]

PR #355, opened 2022-04-12
[flutter_blue_tizen] Initial implementation

JRazek:
Adds support for [flutter blue](https://pub.dev/packages/flutter_blue) on Tizen platform.


swift-kim:
https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline

Please the rename the package directory to `flutter_blue`, and write your own README file without including any Android/iOS-specific information.

swift-kim:
The example app failed to launch on my watch device (running Tizen 5.5). Is this expected?

```
E/LAUNCHPAD(12013): launchpad_loader.c: __loader_terminate_cb(353) > dlopen(/opt/usr/globalapps/com.example.flutter_blue_tizen_example/bin/runner) is failed. error(/lib/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /opt/usr/globalapps/com.example.flutter_blue_tizen_example/bin/../lib/libprotobuf-lite.so))
```

The `libprotobuf-lite.so` file included by this package requires `GLIBCXX_3.4.26`, but the `libstdc++.so` shipped with the device is not compatible with that version (supports only `GLIBCXX_3.4.22`). I think you should either lower your compiler version (if you were using GBS to build the library, switch to the Tizen 5.5 rootstrap), or compile the library into a static library (`.a`).

swift-kim:
@HakkyuKim Please review the CI error. It seems an exception is raised when a package is not found in `recipe.yaml`.

tools/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
 dependencies:
   args: ^2.1.0
   file: ^6.1.0
-  flutter_plugin_tools: ^0.7.3
+  flutter_plugin_tools: ^0.8.2
```
HakkyuKim, line 12: Bumping up minor version when major version is non-positive may introduce breaking changes. Also this change is not relevant to `flutter_blue_tizen`. Please revert it.

HakkyuKim:
> @HakkyuKim Please review the CI error. It seems an exception is raised when a package is not found in `recipe.yaml`.

Oh..made a mistake on that one, thanks. #356 fixes the missing package issue.
The current tool logic for recipe is quite different from our original discussion, I'll redefine the recipe semantics or fix the tool code.

tools/pubspec.yaml
swift-kim: I advised him to try the latest one because the new version allowed use of local `.clang-format` files but the old version didn't. (He should've mentioned about this in the PR comment.)

HakkyuKim: Oh okay, the tool logic doesn't seem to break at the moment. I'll see if there are any required changes in the tool code.

packages/flutter_blue/example/analysis_options.yaml
```diff
@@ -0,0 +1,29 @@
+# This file configures the analyzer, which statically analyzes Dart code to
```
swift-kim, line 1: This file is redundant. Please remove.

packages/flutter_blue/example/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: flutter_blue_tizen_example
+description: Demonstrates how to use the flutter_blue_tizen plugin.
+publish_to: 'none'
+
+environment:
+  sdk: ">=2.12.0 <3.0.0"
+dependencies:
```
swift-kim, line 7: Please add newlines between sections

```suggestion

dependencies:
```
and sort the dependencies alphabetically. (Is `cupertino_icons` used?)

packages/flutter_blue/example/tizen/tizen-manifest.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="com.example.flutter_blue_tizen_example" version="1.0.0" api-version="4.0" xmlns="http://tizen.org/ns/packages">
+    <profile name="mobile"/>
+    <ui-application appid="com.example.flutter_blue_tizen_example" exec="runner" type="capp" multiple="false" nodisplay="false" taskmanage="true" hw-acceleration="on">
```
swift-kim, line 4: What's the reason for using the capp type?

packages/flutter_blue/proto/.gitignore
```diff
@@ -0,0 +1,2 @@
+flutterblue.pb.cc
+flutterblue.pb.h
```
swift-kim, line 2: Every file should end with a newline.

packages/flutter_blue/proto/flutterblue.proto
```diff
@@ -0,0 +1,215 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+syntax = "proto3";
+
+option java_package = "com.pauldemarco.flutter_blue";
+option java_outer_classname = "Protos";
+option objc_class_prefix = "Protos";
+option optimize_for = LITE_RUNTIME;
+package proto.gen;
+// Wrapper message for `int32`.
```
swift-kim, line 12: Apparently this file is not an exact copy of https://github.com/pauldemarco/flutter_blue/blob/master/protos/flutterblue.proto. Any reason?

packages/flutter_blue/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
+name: flutter_blue_tizen
+description: A new flutter plugin project.
+version: 0.0.1
+homepage:
```
swift-kim, line 4: https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline#pubspecyaml

Please fill the information properly.

packages/flutter_blue/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+environment:
+  sdk: ">=2.12.0 <3.0.0"
+  flutter: ">=1.20.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+  flutter_blue: ^0.8.0
```
swift-kim, line 13: Unused dependency. Please remove.

packages/flutter_blue/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
...
+  flutter_test:
+    sdk: flutter
+  flutter_lints: ^1.0.0
+
+flutter:
+  plugin:
+    platforms:
+      tizen:
+        pluginClass: FlutterBlueTizenPlugin
+        fileName: flutter_blue_tizen_plugin.h
```
swift-kim, line 25: Missing newline.

packages/flutter_blue/tizen/external/.clang-format
```diff
@@ -0,0 +1,2 @@
+DisableFormat: true
+SortIncludes: false
```
swift-kim, line 2: Should you specify the `SortIncludes` value? (Isn't `DisableFormat` enough?)

packages/flutter_blue/proto/flutterblue.proto
JRazek: Generated file from the original did not contain any namespaces and it resulted in name collisions.

packages/flutter_blue/example/tizen/tizen-manifest.xml
JRazek: It just turned out to be easier to debug. No problem in changing it back to `C#` if that's preferred.

packages/flutter_blue/tizen/external/.clang-format
JRazek: For some reason after removing `SortIncludes` the analysis fails.

packages/flutter_blue/example/tizen/tizen-manifest.xml
swift-kim: Yes, native apps are easier to debug, but please use the `dotnet` type in the final code. That's the convention of this repository because native apps are not generally allowed to run on TV.

packages/flutter_blue/proto/flutterblue.proto
swift-kim: Okay. Anyway please add a line break before line 12.

packages/flutter_blue/pubspec.yaml
swift-kim: Could you add the following removed lines back to the pubspec?

```yaml
dependencies:
  flutter:
    sdk: flutter

dev_dependencies:
  flutter_lints: ^1.0.0
```

- flutter: Might be parsed by pub.dev.
- flutter_lints: Used by the `analysis_options.yaml` file.

swift-kim:
@JRazek I thought you intentionally added the file because you didn't prefer the [default analysys options](https://github.com/flutter-tizen/plugins/blob/master/analysis_options.yaml). You can remove the file, but the source code will need significant changes.

JRazek:
Why the build CLI keeps failing?

swift-kim:
> Why the build CLI keeps failing?

Apparently the example app has no dependency on the Flutter framework. You should add `flutter:` to `dependencies` and `flutter: ">=2.0.0"` to `environment` in the pubspec.

See: https://github.com/flutter-tizen/plugins/blob/master/packages/audioplayers/example/pubspec.yaml

JRazek:
@swift-kim 
Should I edit anything else for the review?

swift-kim:
@JRazek Still many style issues.

1. Did you read https://google.github.io/styleguide/cppguide.html#Naming?
2. What does `btu` stand for? Is the namespace necessary?
3. Please consider moving non-public headers (anything other than `flutter_blue_tizen_plugin.h`) from the `inc` directory to the implementation directory (`src`). That's the convention of this repo.
4. Use more blank lines, especially between functions and at the beginning and end of a namespace declaration.
5. Could you list all supported devices? Is there an x86 version of `libprotobuf-lite.so`?
6. I don't want to repeat myself any more. Please do what I already told you before asking me again.

@HakkyuKim Please do some full review when you're back.

packages/flutter_blue/CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+## 1.0.0
+* Initial release.
```
HakkyuKim, line 2: Add newline after headings.

```suggestion

* Initial release.
```

packages/flutter_blue/LICENSE
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2022 Samsung Electronics Co., Ltd. All rights reserved.
+// Copyright (c) 2017 The Chromium Authors. All rights reserved.
```
HakkyuKim, line 2: `flutter_blue` is not written by Chromium Authors.
Copy License file from https://github.com/pauldemarco/flutter_blue/blob/master/LICENSE, then add Samsung copyright notice at the top.

packages/flutter_blue/README.md
```diff
@@ -0,0 +1,157 @@
...
+</p>
+<br><br>
+
+## Introduction
+
+FlutterBlue is a bluetooth plugin for [Flutter](https://flutter.dev), a new app SDK to help developers build modern multi-platform apps.
+
+## Alpha version
+
+This library is actively developed alongside production apps, and the API will evolve as we continue our way to version 1.0.
```
HakkyuKim, line 13: README file need not be copied. The following content seems enough.

# flutter_blue_tizen

## Usage

This package is not an _endorsed_ implementation of `flutter_blue`. Therefore, you have to include `flutter_blue_tizen` alongside `flutter_blue` as dependencies in your `pubspec.yaml` file.

```yaml
dependencies:
  flutter_blue: ^0.8.0
  flutter_blue_tizen: ^1.0.0

```

Then you can import `flutter_blue` in your Dart code:

```dart
import 'package:flutter_blue/flutter_blue.dart';
```

For detailed usage, see https://pub.dev/packages/flutter_blue#usage.

## Required privileges

To use this plugin in a Tizen application, you need to declare the following privileges in your `tizen-manifest.xml` file.

```xml
<privileges>
    <privilege>http://tizen.org/privilege/bluetooth</privilege>
</privileges>
```

## Supported devices

Plugin is supported on Tizen 6.5 iot-headed armv7l and aarch64 (RPI4, RB5).

packages/flutter_blue/example/tizen/tizen-manifest.xml
```diff
@@ -0,0 +1,14 @@
...
+        <label>example</label>
+        <icon>ic_launcher.png</icon>
+        <background-category value="media"/>
+        <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true"/>
+    </service-application>
+    <tv-info api-version="8.8.0">
+        <infolink>T-INFOLINK2021-1000</infolink>
+    </tv-info>
+    <feature name="http://tizen.org/feature/screen.size.all"/>
+</manifest>
```
HakkyuKim, line 14: Missing privilege, is it not required?

```xml
<privileges>
    <privilege>http://tizen.org/privilege/bluetooth</privilege>
</privileges>
```

I haven't read the implementation code yet so I don't know how the app is supposed to work, but isn't this a service application? The original capp was ui application: https://github.com/flutter-tizen/plugins/pull/355/commits/da57e6565754c2d41cda11e868aac1bd7471a6e5#diff-5eb7a9e892b5e46abd1d4213e963e37b8c6b4774fa74f82617804e163289a4f2.

packages/flutter_blue/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: flutter_blue_tizen
+description: Tizen implementation of flutter_blue plugin
+version: 1.0.0
+homepage: https://github.com/flutter-tizen/plugins
+repository: https://github.com/flutter-tizen/plugins/tree/master/packages/flutter_blue
```
HakkyuKim, line 5: nit (reordering):
```suggestion
homepage: https://github.com/flutter-tizen/plugins
repository: https://github.com/flutter-tizen/plugins/tree/master/packages/flutter_blue
version: 1.0.0
```

packages/flutter_blue/example/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
+name: flutter_blue_tizen_example
+description: Demonstrates how to use the flutter_blue_tizen plugin.
+publish_to: 'none'
+
+environment:
+  flutter: ">=2.0.0"
+  sdk: ">=2.12.0 <3.0.0"
+  
+dependencies:
+  cupertino_icons: ^1.0.2
```
HakkyuKim, line 10: Not used in the `flutter_blue` example app: https://github.com/pauldemarco/flutter_blue/blob/master/example/pubspec.yaml
```suggestion
```

HakkyuKim (commented):
I still haven't read the implementation, current review is only about the project structure.

You also have to add `flutter_blue_tizen` entry in the top level `README.md` file. Specifically, in [List of packages](https://github.com/flutter-tizen/plugins#list-of-packages):

| [**flutter_blue_tizen**](packages/flutter_blue) | [flutter_blue](https://github.com/pauldemarco/flutter_blue) (3rd-party) | [pub package](https://pub.dev/packages/flutter_blue_tizen) | No |

in between `device_info_plus_tizen` and `flutter_tts_tizen` entry. And in [Device limitations](https://github.com/flutter-tizen/plugins#device-limitations):


| [**flutter_blue_tizen**](packages/flutter_blue) | \<check or cross> | \<check or cross> | \<check or cross> | \<check or cross> | \<Reason why not all boxes are checked>

HakkyuKim:
The development of [flutter_blue](https://github.com/pauldemarco/flutter_blue) is migrated to [flutter_blue_plus](https://pub.dev/packages/flutter_blue_plus) (It's not part of [plus_plugins](https://github.com/fluttercommunity/plus_plugins)), see: https://github.com/pauldemarco/flutter_blue/issues/929.

packages/flutter_blue/example/tizen/tizen-manifest.xml
JRazek: I made a mistake here. It should have privileges and be an `ui-application`

packages/flutter_blue/LICENSE
```diff
@@ -1,28 +1,28 @@
...
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+   * Neither the name of Buffalo PC Inc. nor the names of its
```
HakkyuKim, line 14: This line should be "Neither the names of the copyright holders nor the names of the".

README.md
```diff
@@ -19,6 +19,7 @@ The _"non-endorsed"_ status means that the plugin is not endorsed by the origina
 | [**camera_tizen**](packages/camera) | [camera](https://github.com/flutter/plugins/tree/master/packages/camera) (1st-party) | [![pub package](https://img.shields.io/pub/v/camera_tizen.svg)](https://pub.dev/packages/camera_tizen) | No |
 | [**connectivity_plus_tizen**](packages/connectivity_plus) | [connectivity_plus](https://github.com/fluttercommunity/plus_plugins/tree/main/packages/connectivity_plus) (1st-party) | [![pub package](https://img.shields.io/pub/v/connectivity_plus_tizen.svg)](https://pub.dev/packages/connectivity_plus_tizen) | No |
 | [**device_info_plus_tizen**](packages/device_info_plus) | [device_info_plus](https://github.com/fluttercommunity/plus_plugins/tree/main/packages/device_info_plus) (1st-party) | [![pub package](https://img.shields.io/pub/v/device_info_plus_tizen.svg)](https://pub.dev/packages/device_info_plus_tizen) | No |
+| [**flutter_blue_tizen**](packages/flutter_blue) | [flutter_blue_tizen](https://github.com/fluttercommunity/plus_plugins/tree/main/packages/flutter_blue) (3rd-party) | [![pub package](https://img.shields.io/pub/v/flutter_blue_tizen.svg)](https://pub.dev/packages/flutter_blue_tizen) | No |
```
HakkyuKim, line 22: The second entry in the table refers to the app-facing package, in the case of `fluttter_blue_tizen`, it's `flutter_blue`, it should be: \[flutter_blue](https://github.com/pauldemarco/flutter_blue) (Also note the url difference).

packages/flutter_blue/README.md
```diff
@@ -1,157 +1,41 @@
...
-FlutterBlue is a bluetooth plugin for [Flutter](https://flutter.dev), a new app SDK to help developers build modern multi-platform apps.
-
-## Alpha version
-
-This library is actively developed alongside production apps, and the API will evolve as we continue our way to version 1.0.
-
-**Please be fully prepared to deal with breaking changes.**
-**This package must be tested on a real device.**
+<h1 align="left"> flutter_blue_tizen </h1>
+<br>
```
HakkyuKim, line 3: Please use markdown syntax for consistency : \# flutter_blue_tizen

packages/flutter_blue/README.md
```diff
@@ -1,157 +1,41 @@
...
 
-Having trouble adapting to the latest API?   I'd love to hear your use-case, please contact me.
+[![pub package](https://img.shields.io/pub/v/flutter_blue_tizen.svg)](https://pub.dev/packages/flutter_blue_tizen)
 
-## Cross-Platform Bluetooth LE
-FlutterBlue aims to offer the most from both platforms (iOS, Android and Tizen).
+The Tizen implementation of <a href="https://github.com/pauldemarco/flutter_blue">flutter_blue</a>.
```
HakkyuKim, line 7: Ditto.

```suggestion
The Tizen implementation of [flutter_blue](https://github.com/pauldemarco/flutter_blue).
```

packages/flutter_blue/README.md
```diff
@@ -1,157 +1,41 @@
...
-flutterBlue.startScan(timeout: Duration(seconds: 4));
 
-// Listen to scan results
-var subscription = flutterBlue.scanResults.listen((results) {
-    // do something with scan results
-    for (ScanResult r in results) {
-        print('${r.device.name} found! rssi: ${r.rssi}');
-    }
-});
+This package is not an endorsed implementation of `flutter_blue`. Therefore, you have to include `flutter_blue_tizen` alongside flutter_blue as dependencies in your `pubspec.yaml` file.
```
HakkyuKim, line 12:
```suggestion
This package is not an _endorsed_ implementation of `flutter_blue`. Therefore, you have to include `flutter_blue_tizen` alongside `flutter_blue` as dependencies in your `pubspec.yaml` file.
```

HakkyuKim:
Regarding C++ code:
## Class definition
https://google.github.io/styleguide/cppguide.html#Declaration_Order
> A class definition should usually start with a public: section, followed by protected:, then private:. Omit sections that would be empty.

## Names
### Class data members are snake_case with trailing underscore
https://google.github.io/styleguide/cppguide.html#Variable_Names

> Data members of classes, both static and non-static, are named like ordinary nonmember variables, but with a trailing underscore.

### Function parameters are snake_case
https://google.github.io/styleguide/cppguide.html#Variable_Names
> The names of variables (including function parameters) and data members are all lowercase, with underscores between words.

### Function names are UpperCamelCase
https://google.github.io/styleguide/cppguide.html#Function_Names

### Abbreviated words are same as normal Enlgish words
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> For the purposes of the naming rules below, a "word" is anything that you would write in English without internal spaces. This includes abbreviations, such as acronyms and initialisms. For names written in mixed case (also sometimes referred to as "[camel case](https://en.wikipedia.org/wiki/Camel_case)" or "[Pascal case](https://en.wiktionary.org/wiki/Pascal_case)"), in which the first letter of each word is capitalized, prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC().

### Do not abbreviate unless it is widely accepted
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> Do not abbreviate by deleting letters within a word. As a rule of thumb, an abbreviation is probably OK if it's listed in Wikipedia. Generally speaking, descriptiveness should be proportional to the name's scope of visibility.

For example: mut -> mutex, mess -> message, s -> service

### Try to avoid forward declaration
https://google.github.io/styleguide/cppguide.html#Forward_Declarations

> Avoid using forward declarations where possible. Instead, [include the headers you need](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use).

There are too many forward declartion at the moment, is it really necessary? Can it be redesigned not to do this?
 
### Place const before type not after
https://google.github.io/styleguide/cppguide.html#Use_of_const

> while we encourage putting const first, we do not require it. But be consistent with the code around you!

HakkyuKim:
@JRazek 
Questions:
- How did you build the proto compiler and runtime for C++?
- What command options need be provided to generate code for files in `external` directory?

HakkyuKim:
@JRazek 
Bluetooth constructs (BluetoothService, BluetoothCharacteristic, BluetoothDescriptor, ...) shouldn't be coupled with protobuf data types. For example, `ToProtoDescriptor` method creates the need to store `BluetoothCharacteristic` as `BluetoothDescriptor`'s private member and forward declare `BluetoothCharacteristic` as a type. I suggest following the design of [Java's implementation](https://github.com/pauldemarco/flutter_blue/blob/master/android/src/main/java/com/pauldemarco/flutter_blue/ProtoMaker.java).

JRazek:
> @JRazek Questions:
> 
>     * Which version of the protobuf is used?
> 
>     * How did you build the proto compiler and runtime (aarch64, armel, i586, x86_64) for C++?
> 
>     * What command options need be provided to generate code for files in `external` directory?


Hello, 
1. I used a 3.14.0 version of protobuf
2. Protobuf from [here](https://review.tizen.org/git/?p=platform/upstream/protobuf.git;a=shortlog;h=refs/heads/tizen) was built using GBS.
3. `protoc proto/flutterblue.proto --cpp_out=./`

packages/flutter_blue/tizen/src/GATT/bluetooth_characteristic.cc
```diff
@@ -0,0 +1,213 @@
...
+  proto.set_uuid(Uuid());
+  proto.set_allocated_properties(new proto::gen::CharacteristicProperties(
+      GetProtoCharacteristicProperties(Properties())));
+  proto.set_value(Value());
+  if (service_.GetType() == ServiceType::kPrimary)
+    proto.set_serviceuuid(service_.Uuid());
+  else {
+    SecondaryService& sec = dynamic_cast<SecondaryService&>(service_);
+    proto.set_serviceuuid(sec.Uuid());
+    proto.set_secondaryserviceuuid(sec.PrimaryUuid());
```
HakkyuKim, line 48: This is reversed.

packages/flutter_blue/tizen/src/bluetooth_device_controller.cc
```diff
@@ -0,0 +1,255 @@
...
+
+  LOG_ERROR("bt_gatt_client_request_att_mtu_change", get_error_message(res));
+
+  if (res) throw BtException(res, "bt_gatt_client_request_att_mtu_change");
+}
+
+void BluetoothDeviceController::NotifyDeviceState() const {
+  proto::gen::DeviceStateResponse device_state;
+  device_state.set_remote_id(cAddress());
+  device_state.set_state(LocalToProtoDeviceState(State()));
```
HakkyuKim, line 233: `GetState()`.

packages/flutter_blue/tizen/src/bluetooth_device_controller.cc
```diff
@@ -0,0 +1,255 @@
...
+
+void BluetoothDeviceController::DiscoverServices() {
+  std::scoped_lock lock(operation_mutex_);
+
+  struct Scope {
+    BluetoothDeviceController& device;
+    std::vector<std::unique_ptr<btGatt::PrimaryService>>& services;
+  };
+  services_.clear();
+  Scope scope{*this, services_};
```
HakkyuKim, line 124: This is in local scope that may not live when callback is called.

packages/flutter_blue/tizen/src/GATT/bluetooth_characteristic.cc
```diff
@@ -0,0 +1,213 @@
+#include <GATT/bluetooth_characteristic.h>
+#include <GATT/bluetooth_descriptor.h>
+#include <GATT/bluetooth_service.h>
+#include <bluetooth_device_controller.h>
+#include <bluetooth_manager.h>
+#include <log.h>
+#include <notifications_handler.h>
+#include <utils.h>
```
HakkyuKim, line 8:
```suggestion
#include "notifications_handler.h"
#include "utils.h"
```

Same for all other files.

packages/flutter_blue/tizen/src/GATT/bluetooth_characteristic.cc
```diff
@@ -0,0 +1,213 @@
+#include <GATT/bluetooth_characteristic.h>
+#include <GATT/bluetooth_descriptor.h>
+#include <GATT/bluetooth_service.h>
+#include <bluetooth_device_controller.h>
+#include <bluetooth_manager.h>
```
HakkyuKim, line 5:
```suggestion
#include "GATT/bluetooth_characteristic.h"
#include "GATT/bluetooth_descriptor.h"
#include "GATT/bluetooth_service.h"
#include "bluetooth_device_controller.h"
#include "bluetooth_manager.h"
```

Same for all other files.

packages/flutter_blue/tizen/src/flutter_blue_tizen_plugin.cc
```diff
@@ -0,0 +1,314 @@
...
+        result->Error(e.what());
+      }
+
+    } else if (method_call.method_name() == "deviceState") {
+      std::string device_id = std::get<std::string>(args);
+      std::scoped_lock lock(bluetooth_manager_->bluetoothDevices().mutex_);
+      auto it = bluetooth_manager_->bluetoothDevices().var_.find(device_id);
+
+      if (it != bluetooth_manager_->bluetoothDevices().var_.end()) {
+        auto& device = it->second;
```
HakkyuKim, line 171: This logic should be handled by `BluetoothManager` by something like `GetBluetoothDevice`. It's exposing internal member and how it's being used.

packages/flutter_blue/tizen/src/flutter_blue_tizen_plugin.cc
```diff
@@ -0,0 +1,314 @@
...
+      } else
+        result->Error("device not available");
+
+    } else if (method_call.method_name() == "discoverServices") {
+      std::string device_id = std::get<std::string>(args);
+      std::scoped_lock lock(bluetooth_manager_->bluetoothDevices().mutex_);
+      auto it = bluetooth_manager_->bluetoothDevices().var_.find(device_id);
+
+      if (it != bluetooth_manager_->bluetoothDevices().var_.end()) {
+        auto& device = it->second;
```
HakkyuKim, line 189: Same here.

packages/flutter_blue/tizen/src/flutter_blue_tizen_plugin.cc
```diff
@@ -0,0 +1,314 @@
...
+        result->Error("device not available");
+
+    } else if (method_call.method_name() == "services") {
+      std::string device_id = std::get<std::string>(args);
+      std::scoped_lock lock(bluetooth_manager_->bluetoothDevices().mutex_);
+
+      auto it = bluetooth_manager_->bluetoothDevices().var_.find(device_id);
+
+      if (it != bluetooth_manager_->bluetoothDevices().var_.end()) {
+        auto& device = it->second;
```
HakkyuKim, line 208: Same here.

packages/flutter_blue/tizen/src/bluetooth_manager.cc
```diff
@@ -0,0 +1,467 @@
...
+      LOG_ERROR("bt_adapter_le_stop_scan", res);
+    }
+
+    LOG_ERROR("bt_adapter_le_is_discovering", res);
+  }
+}
+
+void BluetoothManager::Connect(const proto::gen::ConnectRequest& conn_request) {
+  std::unique_lock lock(bluetooth_devices_.mutex_);
+  auto device = bluetooth_devices_.var_.find(conn_request.remote_id())->second;
```
HakkyuKim, line 266: This line will crash the app when it returns `end()`.

packages/flutter_blue/tizen/src/bluetooth_manager.cc
```diff
@@ -0,0 +1,467 @@
...
+  if (device)
+    device->Connect(conn_request.android_auto_connect());
+  else
+    throw BtException("device not found!");
+}
+
+void BluetoothManager::Disconnect(const std::string& device_id) {
+  std::unique_lock lock(bluetooth_devices_.mutex_);
+  auto device = bluetooth_devices_.var_.find(device_id)->second;
```
HakkyuKim, line 275: Same here.

HakkyuKim (commented):
Also please follow:
- From https://github.com/flutter-tizen/plugins/issues/354:
> Use a local variable ret to store the return code of a Tizen API call.

- Add curly braces to conditionals and loops:
https://google.github.io/styleguide/cppguide.html#Conditionals
https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements

packages/flutter_blue/tizen/src/bluetooth_device_controller.cc
JRazek: `bt_gatt_client_foreach_services()` executes synchronously.

HakkyuKim: I see, API signature kind of suggests that it's asynchronous, would you add a comment that it's synchronous?

tools/pubspec.yaml
HakkyuKim, line 12: Revert the change here to resolve conflict.

packages/flutter_blue/tizen/src/GATT/bluetooth_descriptor.h
```diff
@@ -0,0 +1,44 @@
+#ifndef FLUTTER_BLUE_TIZEN_GATT_BLUETOOTH_DESCRIPTOR_H
+#define FLUTTER_BLUE_TIZEN_GATT_BLUETOOTH_DESCRIPTOR_H
+
+#include <map>
+#include <memory>
+#include <string>
+
+#include "bluetooth.h"
```
HakkyuKim, line 8: `bluetooth.h` is a Tizen system header and should be wrapped in angle brackets.

```suggestion
#include <bluetooth.h>
```

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
```diff
@@ -0,0 +1,103 @@
...
+  const NotificationsHandler& cNotificationsHandler() const noexcept;
+
+ private:
+  /**
+   * @brief all attributes are depentent on this mutex
+   */
+
+  std::mutex operation_mutex_;
+
+  std::vector<proto::gen::BluetoothDevice> proto_bluetooth_devices_;
```
HakkyuKim, line 83: Why does `BluetoothDeviceController` manage multiple `BluetoothDevice`? Isn't this the role of a `BluetoothManager`?

JRazek: Depending on whether the `allow duplicates` option is on, this container would aggregate all the results from `scanCallback`. 
They all describe the same device.

HakkyuKim: Does `BluetoothDeviceController` need to aggregate duplicated scan results? The only reason I see for having `BluetoothDeviceController.proto_bluetooth_devices_` is to later return some equal (not identical) proto::gen::BluetoothDevice objects in `BluetoothManager.GetConnectedProtoBluetoothDevices()`. I don't see why this API should return multiple "Connected Info" instances.

If you want to return scan result duplicates when `allow_duplicate` is true, `BluetoothManager.bluetooth_devices_` is effectively working as a cache of scanned devices. See https://github.com/boskokg/flutter_blue_plus/blob/259ebdb15a03e95a4f98280a7e7c4137f4171f88/android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java#L833-L838.

packages/flutter_blue/tizen/src/GATT/bluetooth_descriptor.h
HakkyuKim: Note that other files need fixing as well.

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
JRazek: I believe that `RSSI` can differ.

JRazek: But the authors decided that they don't care as far as I can see in the code...

HakkyuKim: Isn't rssi sent through `ScanResult` not `BluetoothDevice`? `BluetoothManager.GetConnectedProtoBluetoothDevices()` may return value-wise unique objects, while `BluetoothManager.ScanCallback()` may return `ScanResult`s with same bluetooth device info but with different rssi: https://github.com/HakkyuKim/plugins/commit/e7c7a3bff8cdf5edb1339c0bb5008c4aa50274dd

packages/flutter_blue/tizen/src/bluetooth_manager.cc
```diff
@@ -0,0 +1,479 @@
...
+              get_error_message(ret));
+    return;
+  }
+  ret = bt_initialize();
+  if (ret != 0) {
+    LOG_ERROR("[bt_initialize]", get_error_message(ret));
+    return;
+  }
+  ret = bt_gatt_set_connection_state_changed_cb(
+      &BluetoothDeviceController::ConnectionStateCallback, this);
```
HakkyuKim, line 33: Directly calling `bt_gatt_set_connection_state_changed_cb` here forces the class to hold `NotificationsHandler` and depend on protobuf data types. How about:
- https://github.com/HakkyuKim/plugins/commit/fa940cde3f1331b881bfc4c5be81fac4a354a197
- https://github.com/HakkyuKim/plugins/commit/460f85b9a2ce81af8e94b071feca86b32852d615

packages/flutter_blue/tizen/src/GATT/bluetooth_service.h
```diff
@@ -0,0 +1,81 @@
...
+  BluetoothCharacteristic* GetCharacteristic(const std::string& uuid) const;
+
+ protected:
+  bt_gatt_h handle_;
+
+  std::vector<std::unique_ptr<BluetoothCharacteristic>> characteristics_;
+
+  BluetoothService(bt_gatt_h handle);
+
+  BluetoothService(const BluetoothService&) = default;
```
HakkyuKim, line 38: Is this necessary? Isn't default copy constructor is created by default?

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
```diff
@@ -0,0 +1,103 @@
...
+      const std::string& address,
+      NotificationsHandler& notifications_handler) noexcept;
+
+  BluetoothDeviceController(
+      const char* address,
+      NotificationsHandler& notifications_handler) noexcept;
+
+  ~BluetoothDeviceController() noexcept;
+
+  BluetoothDeviceController() = delete;
```
HakkyuKim, line 34: This is not needed, default constructor will not be created when there's a custom constructor.

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
```diff
@@ -0,0 +1,103 @@
...
+
+  BluetoothDeviceController(const BluetoothDeviceController& address) = delete;
```
HakkyuKim, line 36: Any reason for deleting copy constructor?

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
```diff
@@ -0,0 +1,103 @@
...
+  void RequestMtu(uint32_t mtu, const requestMtuCallback& callback);
+
+  void NotifyDeviceState() const;
+
+  const NotificationsHandler& cNotificationsHandler() const noexcept;
+
+ private:
+  /**
+   * @brief all attributes are depentent on this mutex
+   */
```
HakkyuKim, line 79: What do you mean by attributes?

packages/flutter_blue/tizen/src/bluetooth_device_controller.h
HakkyuKim: I don't follow, what's got to do with `std::vector<std::unique_ptr<btGatt::PrimaryService>> services_`?

JRazek: `BluetoothDeviceController`, has `std::unique_ptr<>` in a vector container. It makes it non copyable object and default copy ctor or `operator=` are implicitly deleted.

HakkyuKim:
@JRazek 
There was a double free issue when setting notification callback for characteristics, https://github.com/HakkyuKim/plugins/commit/875f7b734093e27bdbaa4e10f05109b0dc1f3bf3 fixes it, but there's another bug when reading characteristic after setting notification.

HakkyuKim:
The code compiled with an incompatible argument type is because `BluetoothCharacteristic` had a non-explicit constructor.

HakkyuKim:
The current package is not fully complete in implementing `flutter_blue`, I'll be OK merging this without releasing just yet, or releasing it without a major version (0.1.0).

Have you give any thoughts to the comment https://github.com/flutter-tizen/plugins/pull/355#issuecomment-1115775744? Should we release the package for `flutter_blue` and not `flutter_blue_plus`?

There's also [quick_blue](https://github.com/woodemi/quick_blue) which follows federated architecture without protobuf, though I'm not sure how stable it is.

JRazek:
> The current package is not fully complete in implementing `flutter_blue` (state_channel), I'll be OK merging this without releasing just yet, or releasing it without a major version (0.1.0).

Since plugin is pretty much functional, releasing with without a major version would work fine I guess.
In the meantime we could focus on migrating it to a `flutter_blue_tizen_plus` version or do some research on how the `quick_blue` is performing on android/linux devices.

JRazek:
We've decided to postpone the release of the bluetooth plugin on Tizen platform until the flutter_blue_plus version is implemented.

I will use this PR for further changes in the future.

JRazek:
@HakkyuKim 

I've implemented the features that are possible to implement on the Tizen platform. 
Reading RSSI was tricky due to unpublished API's and it had to be done using scanning.
I've also improved a few features such as fixing buggy filters in scan and added event channel handling.

Disabling/enabling the adapter is not implemented on Tizen. It is a private API which cannot be published.

In order to use pairing, we must wait for the merge of https://github.com/boskokg/flutter_blue_plus/pull/109.
It is implemented natively though.

Please have a look and let me know if I should perform any major changes.
I'll be testing the stability of the plugin now.

https://github.com/JRazek/plugins/tree/flutter_blue_plus

HakkyuKim:
@JRazek 
Okay, I'll take a look.

HakkyuKim:
@JRazek 
I'll test the features on an actual device when I get back to office on Thursday (I'm currently working at home).

But even if all features work correctly, the code is not ready for review at this stage as there are many simple mistakes that can easily be fixed on your end.

- Calling unnecessary log functions `LOG_DEBUG`.
- Unused private member `handle_` in `state_handler.h`.
- Inconsistent comment syntax (`/** */` and `//`) in `state_handler.h`
- Unnecessary marking methods with `static` in `bluetooth_manager.h`, such as `GetBluetoothState`.
- Regarding using `std::move` to initialize private member `event_name_` in `SystemEventHandler`, did you intend to remove the value of the passing parameter?
- ...

Please refer to [Google style guide](https://google.github.io/styleguide/cppguide.html), [Use consistent styling across packages](https://github.com/flutter-tizen/plugins/issues/354), and other plugin packages in this repository and try to implement with a consistent style. Then send a PR request if it's ready for review so reviewers can comment on it.

JRazek:
Ok, ill just prepare a PR in such case.

JRazek:
> Unnecessary marking methods with static in bluetooth_manager.h, such as GetBluetoothState.

Why do you consider static member functions to be unnecessary here? 

> Regarding using std::move to initialize private member event_name_ in SystemEventHandler, did you intend to remove the value of the passing parameter?

`std::string event_name` is passed by value in the constructor, which allows moving it when creating a `SystemEventHandler`. Later when it's already passed (either moved or copied), the string is moved again to the `std::string` constructor. 
I applied the same for callback functions setters - passing by value and then moving rather then by `const&` and they copying. It potentially removes unnecessary copy.

HakkyuKim:
> Why do you consider static member functions to be unnecessary here?

`flutter_blue_plus_tizen.cc` doesn't call any static methods of `BluetoothManager`, I don't know why some methods are required as static, can't they all be instance methods?

> std::string event_name is passed by value in the constructor, which allows moving it when creating a SystemEventHandler. Later when it's already passed (either moved or copied), the string is moved again to the std::string constructor.
I applied the same for callback functions setters - passing by value and then moving rather then by const& and then copying. It potentially removes unnecessary copy.

I don't see how the current implementation can remove unnecessary copy compared to

```
SomeClassConstructor(const std::string& s): s_(s) {}
```

Current implementation copies once when passing parameter as value, the implementation above copies by calling a copy constructor. Please pass `const std::string&` for consistency.

Is `SystenEventHandler` really required? Can't the logic just be inside `StateHandler`? See https://github.com/flutter-tizen/plugins/blob/master/packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc#L46.


Did you test `SystemEventHandler`? Only 'off' events are called, I'm not sure if this an API bug or if there's a mistake in the implementation.

JRazek:
> `flutter_blue_plus_tizen.cc` doesn't call any static methods of BluetoothManager, I don't know why some methods are required as static, can't they all be instance methods?

Even if that is true, I don't see any difference. As these static member functions do not use non static members, there's no need to make them belong to instance rather then a class. I will refactor the code as you wish though.

> I don't see how the current implementation can remove unnecessary copy compared to

```cpp
SomeClassConstructor(const std::string& s): s_(s) {}
```
vs
```cpp
SomeClassConstructor(std::string s): s_(std::move(s)) {}
```
It will move the rvalue string rather than copy it, when you call:
```cpp
SomeClassConstructor obj(std::string("foo");
```
[This thread explains it](https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference). I can make all places where this can be applied with `pass by value and move` for consistency.

> Is SystenEventHandler really required? Can't the logic just be inside StateHandler? See https://github.com/flutter-tizen/plugins/blob/master/packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc#L46.

No, it is not required, but it's a modern C++ wrapper on very bug prone [bundle](https://docs.tizen.org/application/native/api/mobile/6.5/group__CORE__LIB__BUNDLE__MODULE.html#gae6bcd5be7bb61f5fa47341d34f5e61c0)
which I was afraid to touch without it. It could be reused in other projects in the future, but it's up to you if I should use it.

> Did you test SystemEventHandler? Only 'off' events are called, I'm not sure if this an API bug or if there's a mistake in the implementation.

Tizen API seems to be unresponsive very often. When I tested events it very often happened that callbacks came after repeated attempts to call API or not at all. I will contact Tizen API authors to verify it, but I believe it's an API issue.

HakkyuKim:
> Even if that is true, I don't see any difference. As these static member functions do not use non static members, there's no need to make them belong to instance rather then a class. I will refactor the code as you wish though.

Generally if client code doesn't need to call the methods by class name, it would not be marked `static`. By your logic, every method should be marked `static`.

> [This thread explains it](https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference). I can make all places where this can be applied with pass by value and move for consistency.

Ok, I understand that there's one less copy in case the parameter was passed as rvalue. Though I'm not quite convinced that this type of optimization is needed in our case since none of other packages in our repo follow this practice. If you want to use this, you can at least make it consistent in `flutter_blue_plus_tizen`.

> Tizen API seems to be unresponsive very often. When I tested events it very often happened that callbacks came after repeated attempts to call API or not at all. I will contact Tizen API authors to verify it, but I believe it's an API issue.

Ok. I believe this bug needs to be fixed before release as app developers may need to listen whether Bluetooth module is available or not.

> No, it is not required, but it's a modern C++ wrapper on very bug prone [bundle](https://docs.tizen.org/application/native/api/mobile/6.5/group__CORE__LIB__BUNDLE__MODULE.html#gae6bcd5be7bb61f5fa47341d34f5e61c0)
which I was afraid to touch without it. It could be reused in other projects in the future, but it's up to you if I should use it.

Having fewer classes and files reduce mistakes in my opinion. For example, `StateHandler::OnCancelInternal` doesn't deregister the `SystemEventCallback` so the program will crash later on `event_sink->Success`.

JRazek:
> Generally if client code doesn't need to call the methods by class name, it would not be marked static. By your logic, every method should be marked static.

I don't see it that way. The only methods that can be static are the ones that do not use non-static members. As these methods don't use them I do not see a reason to make them belong to an instance. They can be as well free functions, but I believe it's easier to keep them in one place.

But ok, I am going to make the unused by client code non-static here. 

> Ok, I understand that there's one less copy in case the parameter was passed as rvalue. Though I'm not quite convinced that this type of optimization is needed in our case since none of other packages in our repo follow this practice. If you want to use this, you can at least make it consistent in flutter_blue_plus_tizen.

AFAIK, that's what clang-tidy embraces. As it is unnecessary copy, why not reduce it. I will make it consistent. :)

> Having fewer classes and files reduce mistakes in my opinion. For example, StateHandler::OnCancelInternal doesn't deregister the SystemEventCallback so the program will crash later on event_sink->Success.

Thanks, haven't noticed it :).
I still embrace that the code should be more modular though. What fixes this bug is a simple setter in `StateHandler::onCancelInternal`.
```cpp
system_event_handler_.SetCallback(nullptr);
```
If all the logic would be in `StateHandler`, I do not think it would change anything. The callback should be deregistered anyway, so the only thing it reduces is modularity here.

> Ok. I believe this bug needs to be fixed before release as app developers may need to listen whether Bluetooth module is available or not.

Ok, I am waiting for a feedback from authors.

JRazek:
I have looked into the static member functions.

the function:
```cpp
  static void BluetoothManager::StartBluetoothDeviceScanLE(const BleScanSettings& scan_settings, ScanCallback callback);
``` 
uses  
```cpp
  static void BluetoothManager::StopBluetoothDeviceScanLE();
  static BluetoothState BluetoothManager::GetBluetoothState();
```
As `BluetoothManager::StartBluetoothDeviceScanLE` is required by RSSI functionality as static, I need these members to be instance independent as well.
I cannot simply make them non-static without some tricks or cycle dependencies.


The only member I can make non-static here is
```cpp
  static bool BluetoothManager::IsBLEAvailable();
```

If you prefer, they can be made as free functions as well.

PR #360, opened 2022-04-15
Implement recipe logic properly

HakkyuKim:
Follows the original semantics of recipe:
- plugins not in the recipe file are not recognized by the tool during integration test.
- plugins with empty list are marked 'excluded' during integration test.

### DIFF
--- a/tools/lib/src/recipe.dart
+++ b/tools/lib/src/recipe.dart
@@ -8,45 +8,38 @@ import 'tizen_sdk.dart';
 
 /// A class that holds data parsed from recipe file.
 class Recipe {
-  Recipe._(Map<String, List<Profile>> profilesPerPackages,
-      {List<String> excluded = const <String>[]})
-      : _profilesPerPackage = profilesPerPackages,
-        _excluded = excluded;
+  Recipe._(Map<String, List<Profile>> profilesPerPackage)
+      : _profilesPerPackage = profilesPerPackage;
 
   /// Creates a [Recipe] instance by parsing [yamlMap].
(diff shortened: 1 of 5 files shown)


tools/lib/src/recipe.dart
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+    return Recipe._(
+      profilesPerPackage,
+      excluded: excluded,
+    );
+  }
+
+  final List<String> _excluded;
+  final Map<String, List<Profile>> _profilesPerPackage;
```
swift-kim, line 39: You may simplify this class by removing `_excluded` and allowing `_profilesPerPackage` to contain empty lists.

In that case `isExcluded` will be just
```dart
  bool isExcluded(String plugin) =>
      isRecognized(plugin) && _profilesPerPackage[plugin]!.isEmpty;
```

tools/test/recipe_test.dart
```diff
@@ -0,0 +1,32 @@
...
+  b: ["wearable-5.5", "tv-6.0"]
+
+  c: []
+''';
+  });
+
+  test('correctly parses recipe', () {
+    final Recipe recipe = Recipe.fromYaml(loadYaml(yamlString) as YamlMap);
+
+    expect(recipe.isRecognized('c'), true);
```
swift-kim, line 25: You may also test for `'d'` (not present in the recipe).

swift-kim (commented):
Just some nitpicks.

HakkyuKim:
Analyze workflow will pass after #372.

swift-kim (approved):
Thank you!

PR #362, opened 2022-04-16
[connectivity_plus] Add bluetooth as connectivity result

swift-kim:
Adds support for `ConnectivityResult.bluetooth`.

Also reverts part of https://github.com/flutter-tizen/plugins/pull/352. The `EventChannel` instance must be preserved in memory during the plugin lifetime. See https://github.com/flutter-tizen/plugins/pull/134.

### DIFF
--- a/packages/connectivity_plus/tizen/src/connection.cc
+++ b/packages/connectivity_plus/tizen/src/connection.cc
@@ -14,6 +14,8 @@ static ConnectionType ToConnectionType(connection_type_e type) {
       return ConnectionType::kMobile;
     case CONNECTION_TYPE_ETHERNET:
       return ConnectionType::kEthernet;
+    case CONNECTION_TYPE_BT:
+      return ConnectionType::kBluetooth;
     case CONNECTION_TYPE_DISCONNECTED:
     default:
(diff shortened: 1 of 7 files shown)


packages/connectivity_plus/tizen/src/connection.cc
```diff
@@ -14,6 +14,8 @@ static ConnectionType ToConnectionType(connection_type_e type) {
       return ConnectionType::kMobile;
     case CONNECTION_TYPE_ETHERNET:
       return ConnectionType::kEthernet;
+    case CONNECTION_TYPE_BT:
+      return ConnectionType::kBluetooth;
```
swift-kim, line 18: Should this function return `ConnectionType::kBluetooth` if the system reported network type is `CONNECTION_TYPE_NET_PROXY`? As far as I understand the proxy type means that the wearable device is using the bluetooth-connected phone's connection.

bwikbs: Relax! :smile: 
Perhaps it is a gimmick that was entered because of a some scenario in the past.

bwikbs (approved):
:+1:

PR #371, opened 2022-04-22
[video_player] Minor code cleanup

swift-kim:
Just some style fixes. Please see the commit history for details.

- Apply naming rules. See the below note for the exceptions.
- Use a fixed-width integer type `int64_t` instead of `long`.
- Remove unnecessary logs. See the below note for details.
- Remove unused buffering-related callbacks.

Notes:
- The types and methods in `message.h` are expected to be auto-generated from [the common interface definition](https://github.com/flutter/plugins/blob/main/packages/video_player/video_player_android/pigeons/messages.dart) and thus do not follow the language-specific naming convention.
- Not all debug logs have been removed because they can be useful for later debugging. I removed only redundant logs that do not carry extra information.
- This plugin does not follow the error code format in https://github.com/flutter-tizen/plugins/issues/354, because providing the failed API name helps identifying the underlying issue quickly.

Contributes to https://github.com/flutter-tizen/plugins/issues/354.

### DIFF
--- a/packages/video_player/tizen/src/message.h
+++ b/packages/video_player/tizen/src/message.h
@@ -160,9 +160,8 @@ class VideoPlayerApi {
   virtual void setMixWithOthers(
       const MixWithOthersMessage &mixWithOthersMsg) = 0;
 
-  static void setup(flutter::BinaryMessenger *binaryMessenger,
-                    VideoPlayerApi *api);
-  static flutter::EncodableValue wrapError(const VideoPlayerError &error);
+  static void SetUp(flutter::BinaryMessenger *messenger, VideoPlayerApi *api);
+  static flutter::EncodableValue WrapError(const VideoPlayerError &error);
 };
 
 #endif  // VIDEO_PLAYER_MESSAGE_H_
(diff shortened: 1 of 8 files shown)


xiaowei-guan (approved):

packages/video_player/tizen/src/message.h
```diff
@@ -1,5 +1,5 @@
-#ifndef VIDEO_PLAYER_MESSAGE_H_
-#define VIDEO_PLAYER_MESSAGE_H_
+#ifndef FLUTTER_PLUGIN_VIDEO_PLAYER_MESSAGE_H_
```
HakkyuKim, line 1: This is a question, is there a reason to prefix header guard with `FLUTTER_PLUGIN`? Should other plugins also do this?

swift-kim: There's not much reason as it's not a public header and thus there's little chance of name conflicts. It's just a convention of this repo. It's not a strict rule so it's up to you.

HakkyuKim (approved):
LGTM

PR #373, opened 2022-05-05
Fixed warnings found with static analysis tool

pkosko:
Fixed warnings:
* not initialized database_ pointer
* dereference without checking null


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -831,7 +831,7 @@ void WebView::InitWebView() {
 void WebView::HandleMethodCall(
     const flutter::MethodCall<flutter::EncodableValue>& method_call,
     std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
-  if (!webview_instance_) {
+  if (!webview_instance_ || !method_call.arguments()) {
```
swift-kim, line 834: Approving for now, but the overall logic should be improved and refactored soon. The `HandleMethodCall` method should not return without sending a result.

swift-kim (approved):

PR #142, opened 2021-07-02
[permission_handler] Revise OpenAppSettings

bbrto21:
* Show own permissions view instead system setting
* Remove limitation about openAppSettings in README

Signed-off-by: Boram Bae <boram21.bae@samsung.com>

PR #379, opened 2022-05-10
[permission_handler] Refactor the C++ code

swift-kim:

`app_settings_manager.cc`:
- Remove the `AppPermissions` class and move its logic to `AppSettingsManager::OpenAppSettings`. The class has been first introduced in #142, but adds unnecessary complexity to the code.
- Replace the `PackageName` class with a simple function `GetPackageName`.

`permission_handler_tizen_plugin.cc` & `permission_manager.cc`:
- The code has almost been rewritten from scratch (reduced 200~ lines of code).
- The `PermissionManager` class is a copy of image_picker_tizen's `PermissionManager`. (See https://github.com/flutter-tizen/plugins/pull/353.)
  - The only difference is that `PermissionStatus` and `PermissionResult` are now a single type, in accordance with the type defined in the [platform interface package](https://github.com/Baseflow/flutter-permission-handler/blob/master/permission_handler_platform_interface/lib/src/permission_status.dart).
- Rename `PermissionGroup` to `Permission`.
- Remove all `on_success` and `on_error` callbacks as everything is now done synchronously.
- Iterating over requested permissions is done inside `HandleMethodCall`'s if-else block.
- Do not redundantly check for the permission status before requesting it.
- Remove [duplication](https://github.com/flutter-tizen/plugins/blob/dd5d56ac7449dc994b66b32ed3f4fc5b0e5af1cd/packages/permission_handler/tizen/src/permission_manager.cc#L197-L203) of `kLocation` results as it's not expected by anybody.

`service_manager.cc`:
- Make `ServiceManager::CheckServiceStatus` synchronous.

### DIFF
--- a/packages/permission_handler/CHANGELOG.md
+++ b/packages/permission_handler/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Code refactoring.
+
 ## 1.2.0
 
 * Update permission_handler to 8.3.0.
--- a/packages/permission_handler/tizen/src/app_settings_manager.cc
+++ b/packages/permission_handler/tizen/src/app_settings_manager.cc
@@ -1,126 +1,86 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 #include "app_settings_manager.h"
 
 #include <app_common.h>
 #include <app_control.h>
 #include <package_manager.h>
 
+#include <string>
+
 #include "log.h"
 
 namespace {
-constexpr char kPackageID[] = "pkgId";
-constexpr char kSettingAppID[] = "com.samsung.clocksetting.apps";
 
-class AppPermissions {
- public:
-  AppPermissions() { Init(); }
+constexpr char kSettingAppId[] = "com.samsung.clocksetting.apps";
 
-  ~AppPermissions() { Deinit(); }
-
-  bool Launch(const std::string& package_name) {
-    if (package_name.length() == 0) {
-      return false;
-    }
-
-    int ret = app_control_add_extra_data(app_control_, kPackageID,
-                                         package_name.c_str());
-    if (ret != APP_CONTROL_ERROR_NONE) {
-      LOG_ERROR("Failed to add key[%s]:value[%s]", kPackageID,
-                package_name.c_str());
-      return false;
-    }
-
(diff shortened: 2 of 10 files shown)


packages/permission_handler/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
```
bbrto21, line 1: When should we use `NEXT` in the change log?

swift-kim: The convention has been taken from:
https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog

It's used when the change doesn't involve a version update but still needs to be documented.

bbrto21: Oh, thank you for your answer.
According to the guide, it seems like every PR should include a change log as much as possible(however it doesn't seem to write a change log enough even in upstream.). what do you think about this?

swift-kim: > This policy exists both to make it easier for maintainers to see a record of all changes to a package, and because some changes (e.g., updates to examples) that do not need to be published may still be of interest to clients of a package.

It's a good practice to record ALL changes but not every member cares about it.😢 At least I'm trying my best.

HakkyuKim (approved):
Looks nice! LGTM.

packages/permission_handler/tizen/src/app_settings_manager.cc
```diff
@@ -1,126 +1,86 @@
...
-
-    ret = app_control_set_app_id(app_control_, kSettingAppID);
-    if (ret != APP_CONTROL_ERROR_NONE) {
-      LOG_ERROR("Failed to set app id[%s]", kSettingAppID);
-    }
+  package_info_h package_info;
+  ret = package_info_create(app_id, &package_info);
+  free(app_id);
+  if (ret != PACKAGE_MANAGER_ERROR_NONE) {
+    LOG_ERROR("Failed to get the package info.");
```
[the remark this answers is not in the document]
swift-kim: Done :)

bbrto21 (approved):

swift-kim:
Let me test this change on TV tomorrow. Now this plugin is expected to work on TV (although not necessary because all permissions are already granted) and we don't have to explicitly exclude the permission_handler dependency from other plugins tests.

swift-kim:
Tested on a TV emulator and it worked.

PR #381, opened 2022-05-11
Fix an error when run without recipe and update the recipe to test on 6.5 TV emulators

HakkyuKim:
Left out `else`. Tool currently fails if ran without recipe file.

### DIFF
--- a/.github/recipe.yaml
+++ b/.github/recipe.yaml
@@ -1,21 +1,21 @@
 plugins:
   audioplayers: ["wearable-5.5"]
   battery_plus: ["wearable-5.5"]
-  connectivity_plus: ["wearable-5.5", "tv-6.0"]
-  device_info_plus: ["wearable-5.5", "tv-6.0"]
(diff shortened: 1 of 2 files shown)


HakkyuKim:
After upgrading the package manager, tv-6.0 emulator package is replaced with tv-6.5.

swift-kim:
The TV SDK is automatically updated by the Package Manager and we don't have control over it. What happens if you don't specify the TV emulator version in the recipe file? The documentation says the version value is optional.

https://github.com/flutter-tizen/plugins/blob/41ea7d58aa9e9b86aaa22c0a7d2b0f01bc720f54/tools/lib/src/tizen_sdk.dart#L85

HakkyuKim:
The original plan was to support testing on all available platform versions by just specifying "device type", `Profile` was designed that way but the command code didn't implement that feature. It currently doesn't work if you omit platform version in recipe.

PR #387, opened 2022-05-16
[geolocator] Refactor the C++ code

JSUYA:
Part of [the code refactoring project.](https://github.com/flutter-tizen/plugins/issues/354)

- Remove unused header
- Change header's macro name
- Fix wrong file name
- Use typedefs of long commonly-used types
- Cleanup error message


JSUYA:
@swift-kim Thank you for tip

and I have a question regarding some github action failures related to this PR.
Could you please check these?

Please let me know if there are any additional actions I need to take

[build test]
When I ran this command in my local environment, the build test became success.
`./tools/tools_runner.sh build-examples     --run-on-changed-packages     --base-sha=$(git rev-parse HEAD^)`
But the github action failed with this log.

```
[   +4 ms] Failed to download https://github.com/flutter-tizen/engine/releases/download/6ac2e23/tizen-common.zip. Ensure you 
have network connectivity and then try again.
```

[dart_analyze fail]
https://github.com/flutter-tizen/plugins/runs/6449786679?check_suite_focus=true
sqflite/example is fail


[integration test]
https://github.com/flutter-tizen/plugins/runs/6449786692?check_suite_focus=true
```
Cannot cache tizen artifacts used for integration-test.
Error: Process completed with exit code 1.
```

swift-kim:
@JSUYA Please check my M Chat message. It could be a DNS issue that should be resolved by the GitHub team.

packages/geolocator/tizen/src/setting.cc
```diff
@@ -9,6 +9,8 @@
 #include <package_manager.h>
 
 namespace {
+constexpr char kSettingAppId[] = "com.samsung.clocksetting.apps";
```
swift-kim, line 12: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

Blank lines immediately inside a declaration of a namespace improves readability.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -51,10 +53,9 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
  private:
   void SetupGeolocatorServiceUpdatesChannel(
       flutter::BinaryMessenger *messenger) {
-    geolocator_service_updates_channel_ =
-        std::make_unique<flutter::EventChannel<flutter::EncodableValue>>(
-            messenger, "flutter.baseflow.com/geolocator_service_updates",
-            &flutter::StandardMethodCodec::GetInstance());
+    geolocator_service_updates_channel_ = std::make_unique<FlEventChannel>(
```
swift-kim, line 56: The redundant `geolocator_` prefix might be omitted for better readability.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -90,111 +90,101 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-  void OnIsLocationServiceEnabled(
-      std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
+  void OnIsLocationServiceEnabled() {
     bool is_enabled = false;
     TizenResult ret = location_manager_->IsLocationServiceEnabled(&is_enabled);
 
     // TODO : add location service listener
     if (!ret) {
-      result->Error("Failed to check service enabled.", ret.message());
+      SendErrorResult(ret.message(), "Failed to check service enabled.");
```
swift-kim, line 138: Why using an error message as an error code? Since the Dart side caller can clearly know that the failed method name is `isLocationServiceEnabled`, you don't have to repeat the information in the error message and can just use this format:

```cpp
SendErrorResult(std::to_string(ret.error_code()), ret.message());
// or
SendErrorResult("Operation failed", ret.message());
```

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -90,111 +90,101 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
     geolocator_updates_channel_->SetStreamHandler(std::move(handler));
```
swift-kim, line 90: Please consider extending the `flutter::StreamHandler` (`FlStreamHandler`) class to create a stream handler class as I've done for [battery_plus_tizen](https://github.com/flutter-tizen/plugins/pull/351) and [connectivity_plus_tizen](https://github.com/flutter-tizen/plugins/pull/352).

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -11,32 +11,34 @@
...
     auto plugin = std::make_unique<GeolocatorTizenPlugin>();
 
     plugin->SetupGeolocatorServiceUpdatesChannel(registrar->messenger());
     plugin->SetupGeolocatorUpdatesChannel(registrar->messenger());
 
     channel->SetMethodCallHandler(
-        [plugin_pointer = plugin.get()](const auto &call, auto result) {
-          plugin_pointer->HandleMethodCall(call, std::move(result));
+        [plugin_ptr = plugin.get()](const auto &call, auto result) {
+          plugin_ptr->HandleMethodCall(call, std::move(result));
```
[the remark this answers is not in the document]
JSUYA: Thanks I changed it

packages/geolocator/tizen/src/permission_manager.h
```diff
@@ -2,11 +2,10 @@
 // Use of this source code is governed by a BSD-style license that can be
```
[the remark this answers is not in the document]
JSUYA: I changed the PermissionManager by referencing another package.
I added `ChangePermissionResulttoLocationPermission` and `ChangePermissionStatustoLocationPermission` because we have to keep the enum order used by the geolocator.

packages/geolocator/tizen/src/setting.cc
```diff
@@ -9,6 +9,8 @@
 #include <package_manager.h>
 
 namespace {
```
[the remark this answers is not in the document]
JSUYA: I changed to app_settings_manager.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: I haven't applied the StreamHandler class yet. Because I still don't perfect understand both StreamHandler and LocationManager. If that's okay with you, I'd like to make this a separate PR.

packages/geolocator/tizen/src/permission_manager.h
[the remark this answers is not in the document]
swift-kim: You can change the values of the `PermissionStatus` enum as necessary (and modify the implementation of `PermissionManager` accordingly) without having to convert between the types. Please refer to https://github.com/flutter-tizen/plugins/pull/379. The `PermissionManager` class in the permission_handler_tizen package has been copied from image_picker_tizen, but has only a single return type `PermissionStatus` because its corresponding Dart type is already defined in the plugin's platform interface.

packages/geolocator/tizen/src/app_settings_manager.cc
```diff
@@ -0,0 +1,155 @@
...
+  if (!SendLauchRequest()) {
+    DestroyAppControl();
+    return false;
+  }
+
+  DestroyAppControl();
+  return true;
+}
+
+bool AppSettingsManager::OpenLocationSetting() {
```
swift-kim, line 81: nit:
```suggestion
bool AppSettingsManager::OpenLocationSettings() {
```

packages/geolocator/tizen/src/app_settings_manager.cc
```diff
@@ -0,0 +1,155 @@
...
+  if (ret != APP_CONTROL_ERROR_NONE) {
+    LOG_ERROR("Failed to send a launch request. (%s)", get_error_message(ret));
+    return false;
+  }
+  return true;
+}
+
+void AppSettingsManager::DestroyAppControl() {
+  app_control_destroy(app_control_);
+}
```
swift-kim, line 155: Every file should have a newline at the end.

packages/geolocator/tizen/src/setting.cc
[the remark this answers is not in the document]
swift-kim: Well, the original `AppControl` class seems to make more sense than individual methods in new `AppSettingsManager` because it can better represent the lifecycle of an app_control handle. Just error handling was missing there. In that way you will have two classes:

- `AppSettingsManager` (has only public methods)
- `AppControl` (statically defined in `app_settings_manager.cc`, used by `AppSettingsManager`)

packages/geolocator/tizen/src/permission_manager.cc
```diff
@@ -1,97 +1,87 @@
...
   }
-  return TizenResult();
+#endif
 }
 
-void PermissionManager::RequestPermssion(const OnSuccess &on_success,
-                                         const OnFailure &on_failure) {
-  const char *permission = kPrivilegeLocation;
-  PermissionResponse response;
+PermissionResult PermissionManager::RequestPermssion(
```
swift-kim, line 40:
```suggestion
PermissionResult PermissionManager::RequestPermission(
```
This is my mistake. I'll update image_picker and permission_handler as well.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
swift-kim: In order to refactor the `LocationManager` class, you may refer to

- `Connection::StartListen()` of connectivity_plus_tizen which returns false on immediate error and takes a result callback which returns a `ConnectionType` on completion (Note that the `ConnectionType` enum has a `kError` value to represent an error state.)
- `ServiceManager` of permission_handler_tizen whose return type is `ServiceStatus`

packages/geolocator/tizen/src/permission_manager.h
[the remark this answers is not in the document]
JSUYA: I changed the PermissionStatus enum of PermissionManager.

packages/geolocator/tizen/src/app_settings_manager.cc
```diff
@@ -0,0 +1,122 @@
...
+  if (ret != APP_CONTROL_ERROR_NONE) {
+    LOG_ERROR("Failed to set an app ID.");
+    app_control_destroy(app_control);
+    return false;
+  }
+
+  ret = app_control_set_operation(app_control, kLocationSettingOperationId);
+  if (ret != APP_CONTROL_ERROR_NONE) {
+    LOG_ERROR("Failed to set operation. (%s)", get_error_message(ret));
+    return false;
```
bbrto21, line 111:
```suggestion
    LOG_ERROR("Failed to set operation. (%s)", get_error_message(ret));
    app_control_destroy(app_control);
    return false;
```

packages/geolocator/tizen/src/setting.cc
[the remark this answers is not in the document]
JSUYA: I changed app_control_h to a local variable and deleted the private methods of AppSettingsManager.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: I extended 'LocationStream' and 'ServiceStatusStream' from 'LocationManager' to StreamHandler.
And I changed Error handling of `LocationManager` from `TizenResult` to throw exception. (For minimize use of output parameters)

packages/geolocator/tizen/src/app_settings_manager.h
```diff
@@ -0,0 +1,21 @@
...
+#include <string>
+
+class AppSettingsManager {
+ public:
+  AppSettingsManager() {}
+  ~AppSettingsManager() {}
+
+  bool OpenAppSettings();
+  bool OpenLocationSettings();
+};
```
bbrto21, line 19: If this class does not have any member variables and does not have any status, why should this class be provided as a class which It needs to be instantiated?

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,117 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-    PermissionStatus permission_status;
-    TizenResult ret =
-        permission_manager_->CheckPermissionStatus(&permission_status);
-    if (!ret) {
-      result->Error("Failed to check permssion status.", ret.message());
+  void OnCheckPermission() {
+    PermissionStatus result =
+        permission_manager_->CheckPermission(kPrivilegeLocation);
+    if (result == PermissionStatus::kError) {
+      SendErrorResult("Operation failed", "Failed to request permission.");
```
bbrto21, line 177:
```suggestion
      SendErrorResult("Operation failed", "Failed to check permission.");
```

JSUYA:
oh.. I will upload a patch for fix this soon
https://github.com/flutter-tizen/plugins/runs/6518068677?check_suite_focus=true#step:5:49

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,117 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-
-    // TODO : add location service listener
-    if (!ret) {
-      result->Error("Failed to check service enabled.", ret.message());
+  void OnIsLocationServiceEnabled() {
+    try {
+      bool is_enabled = location_manager_->IsLocationServiceEnabled();
+      SendResult(flutter::EncodableValue(is_enabled));
+    } catch (const LocationManagerError &error) {
+      // TODO : add location service listener
```
bbrto21, line 189: This comment doesn't seem to have anything to do with error handling.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,117 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-          geolocator_service_updates_event_sink_->Success(value);
-        });
-    if (!tizen_result) {
-      LOG_ERROR("Failed to set OnServiceStateChanged, %s.",
-                tizen_result.message().c_str());
-      geolocator_service_updates_event_sink_ = nullptr;
+  void OnGetCurrentPosition() {
+    auto result_ptr = result_.get();
+    try {
+      location_manager_->GetCurrentPosition(
```
bbrto21, line 220: When the `GetCurrentPosition` is performed without errors, when is the `nullptr` assigned to `result_`?

What if another method request arrives before the callback is executed?

I think `GeolocatorTizenPlugin` should give up ownership of `result_` completely rather than pass reference of `result_` to the callback.

bbrto21: if you capture this pointer in `lambda' for callback functions and still want to keep result as a member, I think you have to add some guard to prevent the result pointer from being replaced with a new one.

Let's assume the following situation.
 1. Register a callback for asynchronously processed requests, the plugin still has the `result` pointer that has not yet sent a result as a member.
 2. a new method request arrives to plugin, but the registered callback is still not executed.
 3. The result pointer is replaced by a new one.
 4. Finally the callback is executed, sending the result to the wrong place.

JSUYA: I changed it to release() `result_` unique ptr.

bbrto21 (approved):

packages/geolocator/tizen/src/setting.cc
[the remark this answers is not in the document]
swift-kim: Alright. I'm okay with either way but you might discuss with the author of the original code (@bbrto21) about the change just in case.

packages/geolocator/tizen/src/location_manager.cc
```diff
@@ -0,0 +1,178 @@
...
+
+          self->location_callback_(position);
+        }
+
+        self->location_callback_ = nullptr;
+        self->location_error_listener_ = nullptr;
+      },
+      this);
+  if (LOCATIONS_ERROR_NONE != ret) {
+    throw LocationManagerError(ret);
```
swift-kim, line 98: When do you use an exception and when do you use the `LastError` pattern? Please choose only one if possible.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,127 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
+            if (result_ptr) {
+              result_ptr->Error("Operation failed", error.GetErrorString());
+              delete result_ptr;
+            }
+          });
+    } catch (const LocationManagerError &error) {
+      if (result_ptr) {
+        result_ptr->Error("Operation failed", error.GetErrorString());
+        delete result_ptr;
+      }
```
swift-kim, line 238: Please use only one error handling mechanism at a time (either error callback or exception). You can remove the try-catch block here and modify `GetCurrentPosition` accordingly.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -11,28 +11,122 @@
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
 
-#include <memory>
-#include <string>
```
swift-kim, line 15: These standard libs are used in this file. Please don't remove the includes. We don't rely on transitive inclusion.

https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

packages/geolocator/tizen/src/app_settings_manager.h
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_PLUGIN_APP_SETTINGS_MANAGER_H_
+#define FLUTTER_PLUGIN_APP_SETTINGS_MANAGER_H_
+
+#include <app_control.h>
+
+#include <string>
```
swift-kim, line 10: Please move these includes to the implementation file.

packages/geolocator/tizen/src/location_manager.h
```diff
@@ -0,0 +1,68 @@
+// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_PLUGIN_LOCATION_MANAGER_H_
+#define FLUTTER_PLUGIN_LOCATION_MANAGER_H_
+
+#include <locations.h>
```
swift-kim, line 8: Same here.

```suggestion
#include <tizen_error.h>
```

packages/geolocator/tizen/src/location_manager.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <functional>
```
swift-kim, line 10:
```suggestion
#include <string>
#include <functional>
```

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,127 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-    if (!ret) {
-      result->Error("Failed to check permssion status.", ret.message());
-      return;
+  void OnCheckPermission() {
+    try {
+      PermissionStatus result =
+          permission_manager_->CheckPermission(kPrivilegeLocation);
+      LOG_INFO("permission_status is %d", result);
+      SendResult(flutter::EncodableValue(static_cast<int>(result)));
+    } catch (const LocationManagerError &error) {
```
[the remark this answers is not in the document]
JSUYA: My mistake, Fixed

packages/geolocator/tizen/src/location_manager.h
```diff
@@ -0,0 +1,68 @@
...
+  LocationManagerError(int error_code) : error_code_(error_code) {}
+  std::string GetErrorString() const { return get_error_message(error_code_); }
+
+ private:
+  int error_code_;
+};
+
+typedef std::function<void(Position)> LocationCallback;
+typedef std::function<void(ServiceStatus)> ServiceStatusCallback;
+typedef std::function<void(LocationManagerError)> LocationErrorListener;
```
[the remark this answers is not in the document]
JSUYA: I changed it with `Callback`

packages/geolocator/tizen/src/location_manager.h
```diff
@@ -0,0 +1,68 @@
...
+  ~LocationManager();
+
+  bool IsLocationServiceEnabled();
+
+  Position GetLastKnownPosition();
+
+  void GetCurrentPosition(LocationCallback location_updated_callback,
+                          LocationErrorListener location_error_listener);
+
+  bool StartServiceUpdatedListen(
```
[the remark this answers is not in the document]
JSUYA: I changed these like this
`StartServiceUpdatedListen` -> `StartListenServiceStatusUpdate`
`StopServiceUpdatedListen` -> `StopListenServiceStatusUpdate`
`StartLocationUpdatesListen` -> `StartListenLocationUpdate`
`StopLocationUpdatesListen` -> `StopListenLocationUpdate`
Please check again if these are appropriate

packages/geolocator/tizen/src/location_manager.cc
JSUYA: I removed `LastError` pattern

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: I have modified the header include codes by referring to the guide. please check again

packages/geolocator/tizen/src/permission_manager.cc
```diff
@@ -1,97 +1,88 @@
...
+      return PermissionStatus::kDeniedForever;
+    case PRIVACY_PRIVILEGE_MANAGER_REQUEST_RESULT_DENY_ONCE:
+      return PermissionStatus::kDenied;
+    case PRIVACY_PRIVILEGE_MANAGER_REQUEST_RESULT_ALLOW_FOREVER:
+      return PermissionStatus::kAlways;
     default:
-      LOG_ERROR("Unknown ppm_request_result_e.");
-      on_failure(TizenResult(ret));
-      break;
+      throw PermissionManagerError("Failed to request permission.");
```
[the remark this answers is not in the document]
JSUYA: I added a `kError` enum and modified it to return this. and removed `PermissionManagerError` class

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: The error callback is the error handling of the callback function of `location_manager_request_single_location` called in `GetCurrentPosition`. As far as i know, since this callback is called asynchronously, it was not possible to throw an exception inside the callback.

There is a way to change `GetCurrentPosition`'s return type to `bool` type and not use exception. (maybe the way you want)
I thought these three methods of `LocationManager` should be implemented in the same context: `GetCurrentPosition`, `GetLastKnownPosition`, `IsLocationServiceEnabled`.
(Reduce output arguments. Use Exceptions Rather Than Return Codes(maybe clean code?))
I don't think there is any particular problem with this way. Do you still think you should fix `GetCurrentPosition` to type `bool` and remove exception throwing? If you want to change it, I'll change it. Or please let me know if there is another way

packages/geolocator/tizen/src/app_settings_manager.h
swift-kim: I mean the both. You missed the first one.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,126 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-    PermissionStatus permission_status;
-    TizenResult ret =
-        permission_manager_->CheckPermissionStatus(&permission_status);
-    if (!ret) {
-      result->Error("Failed to check permssion status.", ret.message());
+  void OnCheckPermission() {
+    PermissionStatus result =
+        permission_manager_->CheckPermission(kPrivilegeLocation);
+    if (result == PermissionStatus::kError) {
+      SendErrorResult("Operation failed", "Permission request failed.");
```
swift-kim, line 177:
```suggestion
      SendErrorResult("Operation failed", "Permission check failed.");
```

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,126 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
       return;
     }
-    LOG_INFO("permission_status is %d", permission_status);
-    result->Success(
-        flutter::EncodableValue(static_cast<int>(permission_status)));
+    LOG_INFO("permission_status is %d", result);
```
swift-kim, line 180: Could you remove this line or format the message properly?

packages/geolocator/tizen/src/location_manager.cc
```diff
@@ -0,0 +1,164 @@
...
+  position.heading = direction;
+  position.speed = speed;
+
+  return position;
+}
+
+void LocationManager::GetCurrentPosition(
+    LocationCallback on_location_callback,
+    LocationErrorCallback on_error_callback) {
+  on_location_callback_ = on_location_callback;
```
swift-kim, line 69: Callback arguments can be named like `on_foo` or `foo_callback` but you shouldn't mix the two.

```suggestion
    LocationCallback on_location_update,
    LocationErrorCallback on_error) {
  location_callback_ = on_location_update;
```

This is what I meant in https://github.com/flutter-tizen/plugins/pull/387#discussion_r880054471. I don't think we need to rename `LocationCallback` to `LocationUpdateCallback` and `location_callback_` to `location_update_callback_`.

You might also refer to https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#naming-rules-for-typedefs-and-function-variables although it's not specifically for C++ code.

packages/geolocator/tizen/src/permission_manager.cc
```diff
@@ -1,97 +1,90 @@
...
     case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_DENY:
     case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_ASK:
-      *permission_status = PermissionStatus::kDenied;
-      break;
+      return PermissionStatus::kDenied;
     case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_ALLOW:
     default:
-      *permission_status = PermissionStatus::kAlways;
-      break;
+      return PermissionStatus::kAlways;
```
swift-kim, line 37:
```cpp
    case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_ALLOW:
      return PermissionStatus::kAlways;
    case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_ASK:
    case PRIVACY_PRIVILEGE_MANAGER_CHECK_RESULT_DENY:
    default:
      return PermissionStatus::kDenied;
```

packages/geolocator/tizen/src/location_manager.cc
JSUYA: @swift-kim
Thank you for detail review.
I have a question. about "the method that is called" in link.
According to the link, the `location_callback_` you suggest should be `handle_location_`, shouldn't it?

And there are two cases of using LocationCallback. (stream, getCurrentPosition).
So I think this is appropriate. What do you think?
```cpp
LocationCallback handle_location_update_;
LocationCallback handle_location_;
...
StartListenLocationUpdate(LocationCallback on_location_update)
handle_location_update_ = on_location_update;
...
GetCurrentPosition(LocationCallback on_location,LocationErrorCallback on_error
handle_location_ = on_location;
```

swift-kim: > `location_callback_` should be `handle_location_`, shouldn't it?

No, `handleFoo` in the document refers to an explicit method or function like this:

```c++
static void HandleFoo(void* value) {
  // Handle the value.
}

SetFooHandler(HandleFoo);
```

> What do you think?

Yes, that looks good but the data member names would be:

```cpp
LocationCallback location_callback_;
LocationCallback location_update_callback_;
```

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: I removed it. it is unnecessary

packages/geolocator/tizen/src/location_manager.cc
JSUYA: Thank you for guide. I changed it

packages/geolocator/tizen/src/permission_manager.cc
```diff
@@ -1,97 +1,90 @@
...
 
+#ifndef TV_PROFILE
 #include <Ecore.h>
 #include <privacy_privilege_manager.h>
+#include <tizen.h>
+#endif
+#include <tizen_error.h>
 
-#include <vector>
+#include <string>
```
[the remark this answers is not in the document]
JSUYA: I modified it

packages/geolocator/tizen/src/permission_manager.cc
```diff
@@ -1,97 +1,90 @@
...
+      return PermissionStatus::kDeniedForever;
+    case PRIVACY_PRIVILEGE_MANAGER_REQUEST_RESULT_DENY_ONCE:
+      return PermissionStatus::kDenied;
+    case PRIVACY_PRIVILEGE_MANAGER_REQUEST_RESULT_ALLOW_FOREVER:
+      return PermissionStatus::kAlways;
     default:
-      LOG_ERROR("Unknown ppm_request_result_e.");
-      on_failure(TizenResult(ret));
-      break;
+      return PermissionStatus::kError;
```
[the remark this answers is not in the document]
JSUYA: I changed the order of the cases to reduce code diff.
But in the old code, the default case returned `KAlways`. So I kept it.
[permission_manager.cc](https://github.com/flutter-tizen/plugins/blob/master/packages/geolocator/tizen/src/permission_manager.cc)

packages/geolocator/tizen/src/permission_manager.cc
JSUYA: Same reason, the default case returned error. So I kept it.
[permission_manager.cc](https://github.com/flutter-tizen/plugins/blob/master/packages/geolocator/tizen/src/permission_manager.cc)

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
JSUYA: I modified it according to your guide.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -49,200 +143,118 @@ class GeolocatorTizenPlugin : public flutter::Plugin {
...
-      geolocator_service_updates_event_sink_;
-  std::unique_ptr<flutter::EventChannel<flutter::EncodableValue>>
-      geolocator_updates_channel_;
-  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>>
-      geolocator_updates_event_sink_;
+  std::unique_ptr<AppSettingsManager> app_settings_manager_;
+  std::unique_ptr<FlEventChannel> service_updates_channel_;
+  std::unique_ptr<FlEventSink> service_updates_event_sink_;
+  std::unique_ptr<FlEventChannel> updates_channel_;
+  std::unique_ptr<FlEventSink> updates_event_sink_;
```
swift-kim, line 257: `service_updates_event_sink_` and `updates_event_sink_` are unused.

packages/geolocator/tizen/src/location_manager.cc
```diff
@@ -0,0 +1,165 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "location_manager.h"
+
+#include "log.h"
```
swift-kim, line 7: Unused include.

packages/geolocator/tizen/src/geolocator_tizen_plugin.cc
```diff
@@ -14,25 +14,119 @@
...
+    events_ = std::move(events);
+    LocationCallback callback = [this](Position position) -> void {
+      events_->Success(position.ToEncodableValue());
+    };
+    try {
+      location_manager_.StartListenLocationUpdate(callback);
+    } catch (const LocationManagerError &error) {
+      return std::make_unique<FlStreamHandlerError>(
+          std::to_string(error.GetErrorCode()), error.GetErrorString(),
+          nullptr);
```
swift-kim, line 49: This part is causing a hard crash (SIGABRT) due to the issue mentioned in https://github.com/flutter-tizen/plugins/pull/351#discussion_r847005275. battery_plus_tizen and connectivity_plus_tizen also had the same problem but we ignored because the APIs used by the plugins had very low chance of getting an error, but unfortunately it's not the case for this plugin (e.g. you will get an error if a permission is not granted to the app).

You may modify the code as follows for now (until the issue gets resolved):
```suggestion
      LOG_ERROR("Failed to set a location update callback: %s",
                error.GetErrorString().c_str());
      return nullptr;
```
I'll check if there's any better way to handle this.

swift-kim: How about this?

```cpp
    } catch (const LocationManagerError &error) {
      // This is an intentional leak.
      // Issue: https://github.com/flutter/flutter/issues/101682
      std::string *code = new std::string(std::to_string(error.GetErrorCode()));
      std::string *message = new std::string(error.GetErrorString());
      return std::make_unique<FlStreamHandlerError>(*code, *message, nullptr);
    }
```

swift-kim (approved):
Thank you very much!

PR #396, opened 2022-05-25
[tizen_app_manager] Support terminating background applications

swift-kim:
Adds an optional `background` parameter to `AppRunningContext.terminate()`. If `background` is true, the method internally invokes `app_manager_request_terminate_bg_app`, otherwise invokes `app_manager_terminate_app` as usual.

There's no test for the new API because testing the API requires a multi-project configuration. I'm planning to unify tizen_app_control and tizen_app_manager into a single package and the test will be added then.

### DIFF
--- a/packages/tizen_app_manager/CHANGELOG.md
+++ b/packages/tizen_app_manager/CHANGELOG.md
@@ -1,4 +1,8 @@
+## 0.1.1
+
+* Add an optional parameter `background` to `AppRunningContext.terminate()`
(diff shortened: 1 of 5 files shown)


HakkyuKim (approved):
LGTM, are you planning to move `app_manager.dart` to `src` in next breaking change?

swift-kim:
@HakkyuKim Yes, that's the plan. The same will be done for tizen_app_control as well.

PR #401, opened 2022-06-10
[flutter_tts] Refactor the C++ code

JSUYA:
Part of [the code refactoring project.](https://github.com/flutter-tizen/plugins/issues/354)

- Extract method.
- Hiding tizen api in plugin.cc
- Minimize use of output parameters
- Change method name(Init-> TtsCreate, DeInit -> TtsDestroy, ...)

### DIFF
--- a/packages/flutter_tts/tizen/inc/flutter_tts_tizen_plugin.h
+++ b/packages/flutter_tts/tizen/inc/flutter_tts_tizen_plugin.h
@@ -1,3 +1,7 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 #ifndef FLUTTER_PLUGIN_FLUTTER_TTS_TIZEN_PLUGIN_H_
 #define FLUTTER_PLUGIN_FLUTTER_TTS_TIZEN_PLUGIN_H_
 
--- a/packages/flutter_tts/tizen/inc/flutter_tts_tizen_plugin.h
+++ b/packages/flutter_tts/tizen/inc/flutter_tts_tizen_plugin.h
@@ -1,7 +1,3 @@
-// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
-
 #ifndef FLUTTER_PLUGIN_FLUTTER_TTS_TIZEN_PLUGIN_H_
 #define FLUTTER_PLUGIN_FLUTTER_TTS_TIZEN_PLUGIN_H_
 
--- a/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
+++ b/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
@@ -9,6 +9,7 @@
 #include <flutter/standard_method_codec.h>
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -136,9 +137,9 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
   }
 
   void OnSpeak(const flutter::EncodableValue &arguments) {
-    TtsState state = tts_->GetState();
-    if (state == TtsState::kPlaying || state == TtsState::kError) {
-      if (state == TtsState::kPlaying) {
+    std::optional<TtsState> state = tts_->GetState();
+    if (!state.has_value() || state == TtsState::kPlaying) {
+      if (state.has_value() && state == TtsState::kPlaying) {
         LOG_ERROR("[TTS] : You cannot speak again while speaking.");
       }
       SendResult(flutter::EncodableValue(0));
--- a/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
+++ b/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
@@ -222,8 +222,8 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
 
   void OnSetVolume(const flutter::EncodableValue &arguments) {
     if (std::holds_alternative<double>(arguments)) {
-      double rate = std::get<double>(arguments);
-      if (tts_->SetVolume(rate)) {
+      double volume = std::get<double>(arguments);
+      if (tts_->SetVolume(volume)) {
         SendResult(flutter::EncodableValue(1));
         return;
       }
--- a/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
+++ b/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
@@ -183,11 +183,15 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
   void OnGetSpeechRateValidRange() {
     int32_t min = 0, normal = 0, max = 0;
     tts_->GetSpeedRange(&min, &normal, &max);
+    // FIXME: The value of "platform" is temporarily set to "android" instead of
+    // "tizen". Because it is not declared in TextToSpeechPlatform of
+    // TextToSpeech example.
     flutter::EncodableMap map = {
         {flutter::EncodableValue("min"), flutter::EncodableValue(min)},
         {flutter::EncodableValue("normal"), flutter::EncodableValue(normal)},
         {flutter::EncodableValue("max"), flutter::EncodableValue(max)},
-        {flutter::EncodableValue("platform"), flutter::EncodableValue("tizen")},
+        {flutter::EncodableValue("platform"),
+         flutter::EncodableValue("android")},
     };
     SendResult(flutter::EncodableValue(map));
   }
--- a/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
+++ b/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
@@ -45,7 +45,7 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
       tts_ = nullptr;
       return;
     }
-    tts_->SetOnStateChanagedCallback(
+    tts_->SetStateChanagedCallback(
         [this](tts_state_e previous, tts_state_e current) -> void {
           std::unique_ptr<flutter::EncodableValue> value =
               std::make_unique<flutter::EncodableValue>(true);
@@ -68,15 +68,14 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
           }
         });
 
-    tts_->SetOnUtteranceCompletedCallback([this](int utt_id) -> void {
-      LOG_INFO("[TTS] Utterance (%d) is completed", utt_id);
+    tts_->SetUtteranceCompletedCallback([this](int32_t utt_id) -> void {
       std::unique_ptr<flutter::EncodableValue> args =
           std::make_unique<flutter::EncodableValue>(true);
       channel_->InvokeMethod("speak.onComplete", std::move(args));
       HandleAwaitSpeakCompletion(1);
     });
 
-    tts_->SetErrorCallback([](int utt_id, tts_error_e reason) -> void {
+    tts_->SetErrorCallback([](int32_t utt_id, tts_error_e reason) -> void {
       // It seems unnecessary for now.
     });
   }
@@ -140,7 +139,7 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
     std::optional<TtsState> state = tts_->GetState();
     if (!state.has_value() || state == TtsState::kPlaying) {
       if (state.has_value() && state == TtsState::kPlaying) {
-        LOG_ERROR("[TTS] : You cannot speak again while speaking.");
+        LOG_ERROR("You cannot speak again while speaking.");
       }
       SendResult(flutter::EncodableValue(0));
       return;
@@ -156,7 +155,6 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
 
     if (tts_->Speak()) {
       if (await_speak_completion_ && !result_for_await_speak_completion_) {
-        LOG_DEBUG("Store result ptr for await speak completion");
         result_for_await_speak_completion_ = std::move(result_);
       } else {
         SendResult(flutter::EncodableValue(1));
@@ -183,23 +181,20 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
   }
 
   void OnGetSpeechRateValidRange() {
-    int min = 0, normal = 0, max = 0;
+    int32_t min = 0, normal = 0, max = 0;
     tts_->GetSpeedRange(&min, &normal, &max);
-    flutter::EncodableMap map;
-    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
-        "min", min));
-    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
-        "normal", normal));
-    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
-        "max", max));
-    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
-        "platform", "tizen"));
-    SendResult(flutter::EncodableValue(std::move(map)));
+    flutter::EncodableMap map = {
+        {flutter::EncodableValue("min"), flutter::EncodableValue(min)},
+        {flutter::EncodableValue("normal"), flutter::EncodableValue(normal)},
+        {flutter::EncodableValue("max"), flutter::EncodableValue(max)},
+        {flutter::EncodableValue("platform"), flutter::EncodableValue("tizen")},
+    };
+    SendResult(flutter::EncodableValue(map));
   }
 
   void OnSetSpeechRate(const flutter::EncodableValue &arguments) {
     if (std::holds_alternative<double>(arguments)) {
-      int speed = (int)std::get<double>(arguments);
+      int32_t speed = std::get<double>(arguments);
       tts_->SetTtsSpeed(speed);
       SendResult(flutter::EncodableValue(1));
       return;
@@ -244,9 +239,8 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
     result_ = nullptr;
   }
 
-  void HandleAwaitSpeakCompletion(int value) {
+  void HandleAwaitSpeakCompletion(int32_t value) {
     if (await_speak_completion_) {
-      LOG_DEBUG("Send result for await speak completion[%d]", value);
       result_for_await_speak_completion_->Success(
           flutter::EncodableValue(value));
       result_for_await_speak_completion_ = nullptr;
--- a/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
+++ b/packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
@@ -36,14 +36,11 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
     });
 
     tts_ = std::make_unique<TextToSpeech>();
-    try {
-      tts_->Initialize();
-    } catch (const TextToSpeechError &error) {
+    if (!tts_->Initialize()) {
       // TODO : Handle initialization failure cases
       // Rarely, initializing TextToSpeech can fail. we should consider catching
       // the exception and propagating it to the flutter side. however, I think
       // this is optional because flutter side is not expecting any errors.
-      LOG_ERROR("Operation failed : %s", error.GetErrorString().c_str());
       tts_ = nullptr;
       return;
     }
@@ -89,6 +86,13 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
   void HandleMethodCall(
       const flutter::MethodCall<flutter::EncodableValue> &method_call,
       std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
+    // Keep in sync with the return values implemented in:
+    // https://github.com/dlutton/flutter_tts/blob/master/android/src/main/java/com/tundralabs/fluttertts/FlutterTtsPlugin.java.
+    // In principle, MethodResult was designed to call MethodResult.Error() to
+    // notify the dart code of any method call failures from the host platform.
+    // However, in the case of flutter_tts, it expects a return value 0 on
+    // failure(and value 1 on success). Therefore in the scope of this plugin,
+    // we call SendResult(flutter::EncodableValue(0)); to notify errors.
     const auto method_name = method_call.method_name();
     const auto &arguments = *method_call.arguments();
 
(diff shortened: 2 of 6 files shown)


HakkyuKim:
I'm not sure changing to exception throws is a good idea when the original code worked fine without them. If you are going to use exceptions, they must all be catched (it's easy to miss in callbacks) as @bbrto21 mentioned.

nit: remove [WIP] in commit message.

packages/flutter_tts/tizen/inc/flutter_tts_tizen_plugin.h
```diff
@@ -1,3 +1,7 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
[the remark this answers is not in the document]
swift-kim: No, in general.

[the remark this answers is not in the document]
bbrto21: @swift-kim OK, should the license notice be added to the header files of other packages as well?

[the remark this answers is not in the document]
swift-kim: ?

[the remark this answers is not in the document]
bbrto21: There is no license notice in the automatically generated header file. 
as the license notice is added in this PR, should we add license notice to other packages? or should we remove this fix from this PR?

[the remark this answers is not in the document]
swift-kim: > should we remove this fix from this PR?

Yes, I think.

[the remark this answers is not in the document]
bbrto21: Thanks you for answer.
CC @JSUYA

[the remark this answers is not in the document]
JSUYA: I will remove this part.
And I did the same thing when refactoring the geolocator_tizen package.
https://github.com/flutter-tizen/plugins/pull/410 PR is fix it.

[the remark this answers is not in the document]
swift-kim: @bbrto21 Basically because no portion of the code is written by the specified copyright holder. If you want to add the information to the plugin header file, the same thing should be added to the `log.h` file for consistency. I don't think that's what you want.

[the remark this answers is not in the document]
bbrto21: Thank you :) It's just that I haven't figured out how to explain this to newcomers.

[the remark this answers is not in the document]
JSUYA: Oh I have a question related to this.
Should I remove the license notice from XXX_plugin.cc and log.h as well?

```shell
~/dev/os/f-project/plugins (dev/flutter_tts_refactor) 18:09:37 $ grep -nr "Copyright 202" ./ | grep "log.h"
./packages/tizen_audio_manager/tizen/src/log.h:  1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.

~/dev/os/f-project/plugins (dev/flutter_tts_refactor) 18:09:40 $ grep -nr "Copyright 202" ./ | grep "plugin.h"
./packages/tizen_audio_manager/tizen/inc/tizen_audio_manager_plugin.h:  1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
./packages/messageport/tizen/inc/messageport_tizen_plugin.h:  1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_package_manager/tizen/inc/tizen_package_manager_plugin.h:  1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
./packages/geolocator/tizen/inc/geolocator_tizen_plugin.h:  1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/battery_plus/tizen/inc/battery_plus_tizen_plugin.h:  1:	// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_app_manager/tizen/inc/tizen_app_manager_plugin.h:  1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.


~/dev/os/f-project/plugins (dev/flutter_tts_refactor) 18:09:46 $ grep -nr "Copyright 202" ./ | grep "plugin.cc"
./packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/sensors_plus/tizen/src/sensors_plus_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc:    1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
./packages/messageport/tizen/src/messageport_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/webview_flutter/tizen/src/webview_flutter_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/image_picker/tizen/src/image_picker_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_package_manager/tizen/src/tizen_package_manager_plugin.cc:    1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
./packages/geolocator/tizen/src/geolocator_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc:    1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc:   1:	// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
./packages/network_info_plus/tizen/src/network_info_plus_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/audioplayers/tizen/src/audioplayers_tizen_plugin.cc:    1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_notification/tizen/src/tizen_notification_plugin.cc:    1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/sqflite/tizen/src/sqflite_plugin.cc:    1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc:    1:	// Copyright 2022 Samsung Electronics Co., Ltd. All rights reserved.
./packages/wakelock/tizen/src/wakelock_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/package_info_plus/tizen/src/package_info_plus_tizen_plugin.cc:   1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
./packages/video_player/tizen/src/video_player_tizen_plugin.cc:   1:	// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
./packages/integration_test/tizen/src/integration_test_plugin.cc:   1:	// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
./packages/connectivity_plus/tizen/src/connectivity_plus_tizen_plugin.cc:   1:	// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
./packages/camera/tizen/src/camera_plugin.cc:    1:	// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
~/dev/os/f-project/plugins (dev/flutter_tts_refactor) 18:09:56 $ 

```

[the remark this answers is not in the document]
JSUYA: >@bbrto21 Basically because no portion of the code is written by the specified copyright holder. If you want to add the information to the plugin header file, the same thing should be added to the log.h file for consistency. I don't think that's what you want.

In https://github.com/flutter-tizen/plugins/pull/410 , I added a commit that removes copyright from the log.

JSUYA:
@bbrto21 @HakkyuKim Thank you for review. I reverted the code related to the exception.

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -109,13 +121,24 @@ std::vector<std::string> &TextToSpeech::GetSupportedLanaguages() {
...
   if (ret != TTS_ERROR_NONE) {
-    LOG_ERROR("[TTS] tts_prepare failed: %s", get_error_message(ret));
+    LOG_ERROR("[TTS] tts_get_state failed: %s", get_error_message(ret));
+    return TtsState::kError;
+  }
+  switch (state) {
+    case TTS_STATE_CREATED:
+      return TtsState::kCreated;
+    case TTS_STATE_READY:
+    default:
```
HakkyuKim, line 135: Just one suggestion based on style guide: https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements.
> If the default case should never execute, treat this as an error.

JSUYA: Thank you for review. I updated the commit.

packages/flutter_tts/tizen/src/text_to_speech.h
```diff
@@ -11,30 +11,24 @@
 #include <string>
 #include <vector>
 
+enum class TtsState { kCreated, kReady, kPlaying, kPaused, kError };
```
bbrto21, line 14: https://docs.tizen.org/application/native/api/mobile/5.5/group__CAPI__UIX__TTS__MODULE.html#ga3290b8dabf73cc17a1baac641465f28f

According to the document, there are no errors state.

bbrto21, line 14: How about this?
```suggestion
enum class TtsState {
  kCreated = TTS_STATE_CREATED,
  kReady = TTS_STATE_READY,
  kPlaying = TTS_STATE_PLAYING,
  kPaused = TTS_STATE_PAUSED,
};
```
then I guess you can use like this
``` c++
  tts_state_e state;
  int ret = tts_get_state(tts_, &state);
  TtsState tts_state(state);
```

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -109,13 +121,25 @@ std::vector<std::string> &TextToSpeech::GetSupportedLanaguages() {
...
 }
 
-tts_state_e TextToSpeech::GetState() {
+TtsState TextToSpeech::GetState() {
   tts_state_e state;
   int ret = tts_get_state(tts_, &state);
   if (ret != TTS_ERROR_NONE) {
-    LOG_ERROR("[TTS] tts_prepare failed: %s", get_error_message(ret));
+    LOG_ERROR("[TTS] tts_get_state failed: %s", get_error_message(ret));
+    return TtsState::kError;
```
bbrto21, line 129: It does not seem appropriate for API call failures turn into the state of TTS.

packages/flutter_tts/tizen/src/text_to_speech.h
HakkyuKim: I believe it's implementing error-as-return value approach without using something like [StatusOr](https://abseil.io/docs/cpp/guides/status#overview-of-abslstatus). Similar implementation is done in [geolocator](https://github.com/flutter-tizen/plugins/blob/5671d565376eeca8a5701e96ffdf648a67a04bd4/packages/geolocator/tizen/src/permission_manager.h#L13-L19).

bbrto21: My view is slightly different and I'd like to politely say that it doesn't matter that there's a precedent already merged. 

first, this enum and related method made to represent and to get TTS life-cycle on Tizen. therefore, using some state that does not exist on a state machine causes confusion. 

second, the return value of `tts_get_state` already represent the success of the function call and various error situations, 
so for error-as-return, It seems reasonable to use return value of `tts_get_state` as it is.

thirdly, furthermore, there seems to be a difference between using `StatusOr<T>` and not. This is similar to the `ErrorOr<T>` that we have been familiar. and the more generalized type is `std::expected` , and it has to be explicit like these. 
```
std::expected<TTsStatus, TizenError> TextToSpeech::GetState()
```

swift-kim: As far as I know this style originated from the [battery_plus_tizen](https://github.com/flutter-tizen/plugins/blob/2bed8cbf8c2e9914531ff8db354e050f37c260a4/packages/battery_plus/tizen/src/device_battery.h#L14) and [connectivity_plus_tizen](https://github.com/flutter-tizen/plugins/blob/2bed8cbf8c2e9914531ff8db354e050f37c260a4/packages/connectivity_plus/tizen/src/connection.h#L20) plugins.

There's no reason for the values of `device_battery_status_e` and `BatteryStatus` to exactly match with each other's. They are just different types by design.

```cpp
enum class BatteryStatus { kFull, kCharging, kDischarging, kUnknown, kError };

typedef enum {
	DEVICE_BATTERY_STATUS_CHARGING, /**< Battery is charging */
	DEVICE_BATTERY_STATUS_DISCHARGING, /**< Battery is discharging */
	DEVICE_BATTERY_STATUS_FULL, /**< Battery is fully charged */
	DEVICE_BATTERY_STATUS_NOT_CHARGING, /**< Battery is not charging */
} device_battery_status_e;
```

The user of the `DeviceBattery` class should only care about the public interface of the class but not its underlying implementation.

bbrto21: That's right, the two types don't necessarily have to be symmetrical.

However, IMO, I'd like to politely say that the error is not the status. the error mentioned here only indicates that it was interrupted by an error during a function call. The same goes for the battery status. The current definitions is ambiguous and confuses those who use the API. In addition, it is not fully documented.

swift-kim: Well, an error status is also a possible result of a function and I think everyone can understand what an `kError` state is without difficulty.

bbrto21: Of course it can be. but when newcomers try to use it, he can feel the ambiguity. because he doesn't know if this is a function call failure or if this module can actually go into an error state.

I think there is nothing wrong with keeping to use this style in terms of program behavior.
In that case, however, additional comments are required.

JSUYA: I agree that the meaning of Error in State is ambiguous. And I don't think we have to necessarily follow the definition of tts_state_e also.
But in this case, as per @bbrto21's comment. I think using std::optional can make it a more clearer.
If you agree with this method, I will try to apply it to other code.

packages/flutter_tts/tizen/src/text_to_speech.cc
JSUYA: done

packages/flutter_tts/tizen/src/text_to_speech.h
bbrto21: @JSUYA Thank you but you don't need to propagate this to every plugin :)

bbrto21 (approved):

JSUYA:
@swift-kim Any opinions?

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -82,102 +93,155 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
...
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "max", max));
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "platform", "tizen"));
+    SendResult(flutter::EncodableValue(std::move(map)));
+  }
+
+  void OnSetSpeechRate(const flutter::EncodableValue &arguments) {
+    if (std::holds_alternative<double>(arguments)) {
+      int speed = (int)std::get<double>(arguments);
```
swift-kim, line 202:
```suggestion
      int32_t speed = std::get<double>(arguments);
```

packages/flutter_tts/tizen/src/text_to_speech.h
```diff
@@ -8,33 +8,28 @@
...
 
   ~TextToSpeech() {
     UnregisterTtsCallback();
-    Deinit();
+    TtsDestroy();
   }
 
+  bool Initialize();
+
   void SetOnStateChanagedCallback(OnStateChangedCallback callback) {
```
swift-kim, line 33: https://github.com/flutter-tizen/plugins/pull/387#discussion_r882489624
```suggestion
  void SetStateChanagedCallback(StateChangedCallback callback) {
```

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -30,15 +30,27 @@ void OnError(tts_h tts, int utt_id, tts_error_e reason, void *user_data) {
 
 }  // namespace
```
swift-kim, line 31: You might just inline the callbacks and move the code to `RegisterTtsCallback` for simplicity. Please also remove unnecessary casts `(void *)` in `RegisterTtsCallback` and `GetSupportedLanaguages`.

swift-kim (commented):
Some nitpicks. Otherwise looks good to me.

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -82,102 +93,155 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
...
+      return;
+    }
+    SendResult(flutter::EncodableValue(0));
+  }
+
+  void OnSpeak(const flutter::EncodableValue &arguments) {
+    std::optional<TtsState> state = tts_->GetState();
+    if (!state.has_value() || state == TtsState::kPlaying) {
+      if (state.has_value() && state == TtsState::kPlaying) {
         LOG_ERROR("[TTS] : You cannot speak again while speaking.");
```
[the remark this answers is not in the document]
JSUYA: I removed [TTS] and [SOUNDMANAGER] tags

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -82,102 +93,155 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
...
-        }
-      } else {
-        result->Success(flutter::EncodableValue(0));
-      }
-    } else if (method_name.compare("stop") == 0) {
-      if (tts_->Stop()) {
-        result->Success(flutter::EncodableValue(1));
+    if (tts_->Speak()) {
+      if (await_speak_completion_ && !result_for_await_speak_completion_) {
+        LOG_DEBUG("Store result ptr for await speak completion");
```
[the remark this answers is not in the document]
JSUYA: removed

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -82,102 +93,155 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
...
+    flutter::EncodableMap map;
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "min", min));
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "normal", normal));
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "max", max));
+    map.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+        "platform", "tizen"));
+    SendResult(flutter::EncodableValue(std::move(map)));
```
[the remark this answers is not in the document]
JSUYA: done

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
JSUYA: done

packages/flutter_tts/tizen/src/text_to_speech.h
```diff
@@ -81,8 +76,8 @@ class TextToSpeech {
   int GetUttId() { return utt_id_; }
 
  private:
-  void Init();
-  void Deinit();
+  bool TtsCreate();
+  void TtsDestroy();
   void Prepare();
   void RegisterTtsCallback();
   void UnregisterTtsCallback();
```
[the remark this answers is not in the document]
JSUYA: I removed TtsCreate and TtsDestory functions

packages/flutter_tts/tizen/src/text_to_speech.h
JSUYA: done

packages/flutter_tts/tizen/src/text_to_speech.cc
JSUYA: done

swift-kim:
Calling `getSpeechRateValidRange` fails with a StateError on Tizen.
```dart
[ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: Bad state: No element
#0      ListMixin.firstWhere (dart:collection/list.dart:167)
#1      FlutterTts.getSpeechRateValidRange (package:flutter_tts/flutter_tts.dart:320)
<asynchronous suspension>
#2      _MyAppState._speak (package:flutter_tts_tizen_example/main.dart:90)
<asynchronous suspension>
```
Maybe we can
- set the value of `platform` to `"android"` (instead of `"tizen"`), or
- just unsupport the method and automatically set `tts_speed_` based on the platform's speed range (proportional to 0.0~1.0). For example, if the max speed is 10 and the given speech rate is 0.5, set the `tts_speed_` to 5.

How do you think? The implementation can be done in a separate PR.

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -149,27 +164,27 @@ bool TextToSpeech::Stop() {
...
   int ret = tts_pause(tts_);
   if (ret != TTS_ERROR_NONE) {
-    LOG_ERROR("[TTS] tts_pause failed: %s", get_error_message(ret));
+    LOG_ERROR("tts_pause failed: %s", get_error_message(ret));
     return false;
   }
   return true;
 }
 
 bool TextToSpeech::SetVolume(double volume_rate) {
```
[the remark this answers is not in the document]
JSUYA: I updated argument name. 
And I tested it on a watch device and a mobile emulator and the volume works fine.

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -56,51 +42,68 @@ void TextToSpeech::Prepare() {
   }
   ret = sound_manager_get_max_volume(SOUND_TYPE_VOICE, &system_max_volume_);
```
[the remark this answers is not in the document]
JSUYA: done

JSUYA:
> Calling `getSpeechRateValidRange` fails with a StateError on Tizen.
> 
> ```dart
> [ERROR:flutter/lib/ui/ui_dart_state.cc(198)] Unhandled Exception: Bad state: No element
> #0      ListMixin.firstWhere (dart:collection/list.dart:167)
> #1      FlutterTts.getSpeechRateValidRange (package:flutter_tts/flutter_tts.dart:320)
> <asynchronous suspension>
> #2      _MyAppState._speak (package:flutter_tts_tizen_example/main.dart:90)
> <asynchronous suspension>
> ```
> 
> Maybe we can
> 
> * set the value of `platform` to `"android"` (instead of `"tizen"`), or
> * just unsupport the method and automatically set `tts_speed_` based on the platform's speed range (proportional to 0.0~1.0). For example, if the max speed is 10 and the given speech rate is 0.5, set the `tts_speed_` to 5.
> 
> How do you think? The implementation can be done in a separate PR.

Could you please explain more how to reproduce this log?
As far as I know, the example in flutter_tts doesn't have the code to call `getSpeechRateValidRange`.
Do I need to test it separately? If the problem is not caused by this refactoring, I think it is better to handle it in another PR.

swift-kim:
@JSUYA You can add
```dart
await FlutterTts.getSpeechRateValidRange;
```
to anywhere in the example app (like `initState` or `_speak`). And yes, the error has not been caused by this PR, but I mentioned here because the first solution in the above requires only a single line of change. If you prefer the second solution, we can work on it in another PR.

swift-kim (approved):
Thank you!

Issue #389, opened 2022-05-19
Update analysis_options.yaml


swift-kim:
Recently in the upstream repo, some updates have been made to the analysis_options.yaml file and example apps. We need to sync our code again accordingly.

https://github.com/flutter/plugins/commits/main/analysis_options.yaml

swift-kim:
@bbrto21 @bwikbs @seungsoo47 Could you remove references to `analysis_options_legacy.yaml` in `camera_tizen`, `google_maps_flutter_tizen`, and `webview_flutter_tizen` as well? You will first need to resync all their Dart code (the example app and etc.) and update app-facing and platform interface packages as I've done in https://github.com/flutter-tizen/plugins/pull/392.

bbrto21:
@swift-kim 
I'll check `camera_tizen` but depending on my work priority, this may be processed late.
Is that okay?

swift-kim:
@bbrto21 That's okay. This whole work is a quite demanding task and my expectation is that it will take at least a month to close this issue.

PR #402, opened 2022-06-13
[google_maps_flutter] Fix various bugs

seungsoo47:
* Remove analysis_options.yaml.
* google_maps_flutter version up.
* Fix crash issues and bugs.

Signed-off-by: leess <seungsoo47.lee@samsung.com>

### DIFF
--- a/packages/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 0.1.2
 
 * Fix variouse bugs.
+* Update webview_flutter to 3.0.4.
+* Update webview_flutter_tizen to 0.5.1.
 
 ## 0.1.1
 
--- a/packages/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.2
+
+* Fix variouse bugs.
+
 ## 0.1.1
 
 * Update webview_flutter_tizen to 0.3.8.
--- a/packages/google_maps_flutter/README.md
+++ b/packages/google_maps_flutter/README.md
@@ -20,8 +20,8 @@ This package is not an _endorsed_ implementation of `google_maps_flutter`. There
 
 ```yaml
 dependencies:
-  google_maps_flutter: ^2.0.8
-  google_maps_flutter_tizen: ^0.1.1
+  google_maps_flutter: ^2.1.7
+  google_maps_flutter_tizen: ^0.1.2
 ```
 
 In addition, you need a Maps JavaScript API Key to use this plugin. Please refer to the site below to get the API key.  
--- a/packages/google_maps_flutter/analysis_options.yaml
+++ b/packages/google_maps_flutter/analysis_options.yaml
@@ -1,1 +1,0 @@
-include: ../../analysis_options_legacy.yaml
--- a/packages/google_maps_flutter/example/assets/map.html
+++ b/packages/google_maps_flutter/example/assets/map.html
@@ -18,7 +18,7 @@
       let map;
       console.log("before load");
     </script>
-    <script src="https://maps.googleapis.com/maps/api/js?v=beta&key=YOUR_API_KEY">
+    <script src="https://maps.googleapis.com/maps/api/js?v=beta&key=AIzaSyBqXEo_1aZqXj6Gr7u2ImhegZjbOMmKbzg">
     </script>
   </body>
 </html>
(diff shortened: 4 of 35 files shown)


seungsoo47:
This patch should be merged after the latest engine/webview plugin is released. It is because the interfaces of the two modules do not match.

seungsoo47:
https://github.com/flutter-tizen/plugins/issues/389

HakkyuKim:
Since `0.1.2` is not released, all changes can be part of `0.1.2` and not `0.1.3`. Do you have a reason to release two versions here?

seungsoo47:
> Since `0.1.2` is not released, all changes can be part of `0.1.2` and not `0.1.3`. Do you have a reason to release two versions here?

I thought that @bbrto21 already released `0.1.2` in the below patch.
https://github.com/flutter-tizen/plugins/pull/369/files
However, it just seems the version was not written in Changelog.md.

HakkyuKim:
Ah, I guess that was supposed to be released with CHANGELOG, since it wasn't, we can release `0.1.2` together with the current changes.

seungsoo47:
> Ah, I guess that was supposed to be released with CHANGELOG, since it wasn't, we can release `0.1.2` together with the current changes.

Thank you very much :)

bbrto21:
> * google_maps_flutter version up.

Thanks Thank you for your hard work. but I have a question.
Shouldn't we update the example to the latest one instead of resolving issues of  lint that exists in current example?
If not, do you have a plan to update the example with implementing additional feature of this plugin in the future?

bbrto21:
I'm asking that if it's necessary to copy the latest original example to this plug-in.
I tried it based on this PR, and it seems to be no problem with the execution and behavior of the example, but some lint issues have been reported after the analyze.

seungsoo47:
> I'm asking that if it's necessary to copy the latest original example to this plug-in. I tried it based on this PR(I just replaced lib dir of the example.), and it seems to be no problem with the execution and behavior of the example, but some lint issues have been reported after the analyze.

Thank you for your kind review. 😀 I am not sure that you remember what I said for this plugin. As you know, this plugin is based on webview plugin and uses google map web apis. However, the original examples are based on google map apis and the apis are different from web version apis a little bit. So, I removed some examples related to unsupported apis. By the way, if I miss some examples based on supported apis, please let me know it. ^^

packages/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.1.3
+
+* Fix variouse bugs.
+* Update webview_flutter_tizen to 0.5.1.
+
+## 0.1.2
+
+* Update dependency package.
+
```
HakkyuKim, line 9: Is the version related changes in progress? 

```suggestion
## 0.1.2

* Fix variouse bugs.
* Update webview_flutter to 3.0.4.
* Update webview_flutter_tizen to 0.5.1.

```

And all version related changes in `pubspec.yaml`s and `README`s.

seungsoo47: Oh, I am sorry for being too late. I fixed them. Thank you 😀

packages/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## 0.1.2
 
 * Fix variouse bugs.
+* Update webview_flutter to 3.0.4.
+* Update webview_flutter_tizen to 0.5.1.
+* Update dependency package.
```
HakkyuKim, line 6: Thanks for the changes, I'm not sure if there are other meanings to "Update dependency package" other than bumping up `webview_flutter` and `webview_flutter_tizen`. This line is not needed if you meant those two packages.

seungsoo47: Ah~ As I mentioned above, actually @bbrto21 's patch was 0.1.2 and it will be released with this PR. So I thought I needed `Update dependency package` about @bbrto21 patch. What do you think? :)

bbrto21: @seungsoo47 I think you don't need a last line. 
```suggestion
```

swift-kim:
Any updates? This package is the last package that depends on the `analysis_options_legacy.yaml` file. The file will be deleted after this PR is merged.

seungsoo47:
> Any updates? This package is the last package that depends on the `analysis_options_legacy.yaml` file. The file will be deleted after this PR is merged.
No more updates. Please merge this pr.

bbrto21 (approved):

PR #398, opened 2022-06-02
[webview_flutter] Apply PlatformView API change

bbrto21:
Signed-off-by: Boram Bae <boram21.bae@samsung.com>

PR #404, opened 2022-06-14
[webview_flutter] Handle back key

bbrto21:
* Perform go back when back key pressed if navigation stack has a item.
  if not, the app will be terminated by the engine.


bbrto21:
This PR is based on [previous discussion](https://github.com/flutter-tizen/plugins/pull/398#discussion_r895493710).

swift-kim (approved):

Issue #251, opened 2021-10-26
[audioplayers] Missing implementations


swift-kim:
Partially implemented:
- [ ] `AudioPlayer.play` (not supported arguments: `isLocal`, `respectSilence`, `stayAwake`, `duckAudio`, `recordingActive`)
- [ ] `AudioPlayer.playBytes` (not supported arguments: `isLocal`, `respectSilence`, `stayAwake`, `duckAudio`, `recordingActive`)
- [ ] `AudioPlayer.setUrl` (not supported arguments: `isLocal`, `respectSilence`, `recordingActive`)

Not implemented:
- [ ] `Logger.changeLogLevel` (generates the following error when invoked)<p>
  ```
  Unhandled Exception: PlatformException(Invalid Player ID, Invalid Player ID for method changeLogLevel, null, null)
  ```
- [x] `NotificationService` (not applicable for Tizen)

Needs improvement:
- [ ] `AudioPlayer.playBytes` (the limitation mentioned in README may be relieved by adding a dartPluginClass)

This issue is just for tracking purpose. We're not going to implement all the functions.

PR #406, opened 2022-06-14
[audioplayers] Update audioplayers

### LINKED ISSUE #382: [audioplayers] Fix a crash issue on TV
~~Calling `player_destroy()` in the `OnPlayCompletedcallback` function crashes the player module on TV. Instead, `player_unprepare()` seems like a more appropriate function to call in `Release()` according to the documentation of [ReleaseMode](https://github.com/bluefireteam/audioplayers/blob/a9d2d01b8aaab1a57437291c446913957f845224/packages/audioplayers_platform_interface/lib/api/release_mode.dart#L9).~~ See commit message in https://github.com/flutter-tizen/plugins/pull/406/commits/10ea3dd7f0f996c8a4825242414c153beaf3157b for reasons why some issues occurred on TV devices.

~~As there are quite a bit of change in audioplayers 1.0.0,~~ This issue may be fixed when supporting audioplayers 1.0.0 version.

HakkyuKim:
Closes #382. And possibly #251?

## Summary
- Update implementation for `audioplayers_platform_interface` package.
- Make `audioplayers_tizen` stable for TV devices.

## Existing bug fixes
- TV devices now send complete event properly when audio play finishes.
- `setPlaybackRate` doesn't fail if triggered within last 3 seconds.
- `seek` method will work for different playback rates.

## Previously existed but newly discovered bug(?)
These are documented in the [README](https://github.com/flutter-tizen/plugins/pull/406/commits/f8d61db3cb4ad87429b41ca9734e1ca08d3b151c) file which may be resolved in the future.
- `setVolume` has no effect on TV devices, I suspect TV devices only allow system volume changes.
- If player is set in loop mode by `player_set_looping`, it will not fire callback added by `player_set_completed_cb`. Also playback rate, and possibly other configurations, is reset before audio is replayed. This doesn't seem like an API bug, it seems intended.

### TODO:
- [x] Fix integration test failure.
- [x] Update README.
- [x] Fix https://github.com/flutter-tizen/plugins/issues/382.
- [x] Fix limitations mentioned in https://github.com/flutter-tizen/plugins/tree/master/packages/audioplayers#limitations.


packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -167,29 +166,25 @@ void AudioPlayer::SetDataSource(std::vector<uint8_t> &data) {
...
-      int ret = player_set_playback_rate(player_, rate);
-      if (ret != PLAYER_ERROR_NONE) {
-        throw AudioPlayerError("player_set_playback_rate failed",
-                               get_error_message(ret));
-      }
+  playback_rate_ = rate;
+  player_state_e state = GetPlayerState();
+  if (state == PLAYER_STATE_READY || state == PLAYER_STATE_PLAYING ||
+      state == PLAYER_STATE_PAUSED) {
+    int ret = player_set_playback_rate(player_, rate);
```
bbrto21, line 184: for consistency
```suggestion
    int ret = player_set_playback_rate(player_, playback_rate_);
```

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -167,29 +166,25 @@ void AudioPlayer::SetDataSource(std::vector<uint8_t> &data) {
 }
 
 void AudioPlayer::SetVolume(double volume) {
-  if (volume_ != volume) {
```
bbrto21, line 170: You removed the comparison between old and new values in general. could you please explain the reason in detail?

HakkyuKim: The old value (cached value) becomes invalid when audio is replayed in loop mode because "playback rate" (and possibly volume, I haven't checked) is automatically reset to 1.0. If playback rate was set to 2.0 during previous iteration, calling `SetVolume(2.0)` at next iteration will not change the playback rate properly because of the cached value.

packages/audioplayers/tizen/src/audio_player.cc
HakkyuKim: Done!

HakkyuKim:
> To minimize asynchronous calls, you might consider using `player_prepar` instead of `player_prepar_async`.

I actually like the idea of changing prepare to synchronous, I'll check how it would affect the return time of `Future`s in dart side.

packages/audioplayers/tizen/src/audioplayers_tizen_plugin.cc
```diff
@@ -90,89 +90,68 @@ class AudioplayersTizenPlugin : public flutter::Plugin {
...
       } else if (method_name == "setVolume") {
         player->SetVolume(GetRequiredArg<double>(arguments, "volume"));
-        result->Success(flutter::EncodableValue(1));
-      } else if (method_name == "setUrl") {
-        player->SetUrl(GetRequiredArg<std::string>(arguments, "url"));
-        result->Success(flutter::EncodableValue(1));
+        result->Success();
+      } else if (method_name == "setSourceUrl") {
+        bool isLocal = false;
+        GetValueFromEncodableMap(arguments, "playerId", isLocal);
```
swift-kim, line 127: `isLocal` should be `is_local`.

Also why does `playerId` map to a boolean value?

packages/audioplayers/README.md
```diff
@@ -53,40 +54,18 @@ For detailed information on Tizen privileges, see [Tizen Docs: API Privileges](h
...
+- [ ] `AudioPlayer.global.setGlobalAudioContext` (not supported)
 
 ## Limitations
 
-This plugin has some limitations on TV devices.
-
-- The `setPlaybackRate` method will fail if triggered within last 3 seconds.
-- The playback rate will reset to 1.0 when audio is replayed in loop mode.
-- The `seek` method works only when playback rate is 1.0, and it sets audio position to the nearest key frame which may differ from the passed argument.
+- `onPlayerComplete` event will not be fired when `ReleaseMode` is set to loop which differs from the behavior specified in the [documentation](https://github.com/bluefireteam/audioplayers/blob/e6532371a3372c0f208abb24f13e1e82a9c1e040/packages/audioplayers/lib/src/audioplayer.dart#L64). And playback rate will reset to 1.0 when audio is replayed.
```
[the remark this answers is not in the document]
HakkyuKim: Yes it would, fixed.

packages/audioplayers/example/integration_test/audioplayers_test.dart
```diff
@@ -4,6 +4,7 @@ import 'package:audioplayers/audioplayers.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
+const String assetAudio = 'nasa_on_a_mission.mp3';
 const Duration _kPlayDuration = Duration(seconds: 1);
```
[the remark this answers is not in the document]
HakkyuKim: Done.

packages/audioplayers/tizen/src/audioplayers_tizen_plugin.cc
HakkyuKim: Oh, this is a mistake. Fixed!

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -334,79 +326,101 @@ player_state_e AudioPlayer::GetPlayerState() {
...
-    if (ret != PLAYER_ERROR_NONE) {
-      player->error_listener_(player->player_id_, "player_start failed.");
-      return;
-    }
-    player->EmitPositionUpdates();
-    player->should_play_ = false;
-  }
+  // On TV devices, callbacks are not executed on the main loop. Therefore
+  // we explicitly transfer the callback to the main loop to avoid any race
+  // conditions and to allow creating timer objects in EmitPositionUpdates.
```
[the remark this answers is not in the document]
HakkyuKim: Forgot to change the method name is comments, fixed!

packages/audioplayers/tizen/src/audio_player.cc
HakkyuKim: Done.

HakkyuKim:
@bbrto21 
I've checked implementation of Android and it uses asynchronous prepare method. I don't think it's necessary to differ the logic here though I agree that synchronous method may make code simple (and more sense for `Future` returning APIs), I'm thinking of leaving as it is.

bbrto21:
@HakkyuKim Thank you for your answer.

bbrto21 (approved):

PR #407, opened 2022-06-15
[tizen_app_control] Remove AppManager

swift-kim:
There are some breaking changes:
- Increase the package version to 0.2.0.
- The deprecated class `AppManager` has been removed completely.
- `app_control.dart` has been moved to the implementation directory (`src`), and should not be imported from outside of the library.


bbrto21 (approved):
LGTM

PR #408, opened 2022-06-15
[tizen_app_manager] Refactoring the C++ code

Swanseo0:
Signed-off-by: swan.seo <swan.seo@samsung.com>


packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -7,63 +7,65 @@
 #include "tizen_app_manager.h"
 
 static bool AppInfoCB(app_info_h app_info, void *user_data) {
```
bbrto21, line 9: > Consider inlining a C-style callback using a lambda expression if it makes the code cleaner.

we prefer lambda instead a C-style callback.

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
```diff
@@ -21,12 +21,12 @@ class TizenAppManagerPlugin : public flutter::Plugin {
...
-  flutter::EncodableList m_applications;
-  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> m_launch_events;
+  flutter::EncodableList applications_;
+  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> launch_events_;
   std::unique_ptr<flutter::EventSink<flutter::EncodableValue>>
-      m_terminate_events;
-  bool m_registered_event_cb;
-  int m_registered_cnt;
+      terminate_events_;
+  bool registered_event_cb_;
```
bbrto21, line 28: suffix `cb` is typically used when a variable is a function pointer used as a callback. please consider renaming.

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -7,63 +7,65 @@
...
-    TizenAppManagerPlugin *plugin = (TizenAppManagerPlugin *)user_data;
-    flutter::EncodableMap value;
-    int ret = application_utils::GetAppData(app_info, value);
-    if (ret == APP_MANAGER_ERROR_NONE) {
-      plugin->m_applications.push_back(flutter::EncodableValue(value));
-    }
+  if (app_info == nullptr) {
+    return false;
+  }
+  TizenAppManagerPlugin *plugin = (TizenAppManagerPlugin *)user_data;
```
bbrto21, line 13: Please use C++ style type cast.

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -157,30 +159,28 @@ void TizenAppManagerPlugin::GetApplicationInfo(
...
-cleanup:
-  if (app_info) {
-    app_info_destroy(app_info);
-  }
 }
 
 void TizenAppManagerPlugin::ApplicationIsRunning(
     const flutter::EncodableValue &arguments, MethodResultPtr result) {
   std::string id = "";
   const char *app_id;
```
bbrto21, line 183:
```suggestion
  const char *app_id = nullptr;
```

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -157,30 +159,28 @@ void TizenAppManagerPlugin::GetApplicationInfo(
...
-  char *err_msg;
   bool running;
```
bbrto21, line 184:
```suggestion
  bool running = false;
```

bbrto21: Please initialize local variables of `GetApplicationInfo` too.

packages/tizen_app_manager/tizen/src/application_utils.cc
```diff
@@ -32,117 +32,166 @@ bool AppMetaDataCB(const char *key, const char *value, void *user_data) {
   return true;
 }
 
-int GetAppData(app_info_h app_info, flutter::EncodableMap &value) {
+static char *GetAppId(app_info_h app_info, flutter::EncodableMap &value) {
```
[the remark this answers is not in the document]
Swanseo0: I will apply it. Thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: Initialized the parts you mentioned. Thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: it has been modified by using static_cast. Thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
Swanseo0: I agree that it is inappropriate to put cb as a suffix.
In addition, we will change it to has_registered_event_ to clearly convey what the variable means.

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: the callback function registration method using lambda has been changed to use.

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
```diff
@@ -21,12 +21,12 @@ class TizenAppManagerPlugin : public flutter::Plugin {
...
+  int registered_cnt_;
```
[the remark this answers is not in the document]
Swanseo0: `applications_`, `registered_event_cb_` and `registered_cnt_` have been changed to private areas. However, `launch_events_` and `terminate_events_` have not been modified yet. Because `launch_events_` and `terminate_events_` are unique_ptr.

swift-kim:
There's a typo in the PR title.

Swanseo0:
> There's a typo in the PR title.

Thank you for letting me know.
i will modify it to Refactring->Refactoring.

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
```diff
@@ -16,18 +16,15 @@ class TizenAppManagerPlugin : public flutter::Plugin {
   using MethodResultPtr =
       std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>>;
   static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar);
+  flutter::EncodableList &Applications();
+  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> launch_events_;
+  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>>
+      terminate_events_;
```
JSUYA, line 22: Like `batter_plus` you can declare `flutter::EventSink<flutter::EncodableValue>` as `FlEventSink`.
Please refer here.

https://github.com/flutter-tizen/plugins/blob/2bed8cbf8c2e9914531ff8db354e050f37c260a4/packages/battery_plus/tizen/src/battery_plus_tizen_plugin.cc#L21-L27

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -76,6 +23,10 @@ void TizenAppManagerPlugin::RegisterWithRegistrar(
   registrar->AddPlugin(std::move(plugin));
 }
 
+flutter::EncodableList &TizenAppManagerPlugin::Applications() {
+  return this->applications_;
+}
+
 void TizenAppManagerPlugin::HandleMethodCall(
     const flutter::MethodCall<flutter::EncodableValue> &method_call,
     std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
```
JSUYA, line 32: Prefer == over .compare() when comparing strings.
Please refer this.  https://github.com/flutter-tizen/plugins/issues/354
```cpp
const auto &method_name = method_call.method_name();
if (method_name == "wifiName") {
  ...
}
```

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -98,24 +49,68 @@ void TizenAppManagerPlugin::HandleMethodCall(
...
+                 app_context, clone_context, int_address);
+        TizenAppManagerPlugin *plugin =
+            static_cast<TizenAppManagerPlugin *>(user_data);
+        flutter::EncodableMap msg;
+        msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+            "appId", std::string(app_id)));
+        msg.insert(std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+            "handle", int_address));
+        if (event == APP_CONTEXT_EVENT_LAUNCHED &&
+            plugin->launch_events_ != nullptr) {
```
JSUYA, line 87: Prefer if (value) over if (value != nullptr).

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -98,24 +49,68 @@ void TizenAppManagerPlugin::HandleMethodCall(
...
     return;
   }
-  m_registered_event_cb = true;
+  has_registered_event_ = true;
 }
 
 void TizenAppManagerPlugin::UnregisterObserver() {
   LOG_DEBUG("UnregisterObserver");
-  if (m_registered_event_cb == true) {
+  if (has_registered_event_ == true) {
```
JSUYA, line 108: I think `if (has_registered_event_)` would be better

JSUYA (commented):
I did some reviews based on the rule of this link. Please refer
https://github.com/flutter-tizen/plugins/issues/354

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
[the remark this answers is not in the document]
bbrto21: > `launch_events_` and `terminate_events_` are unique_ptr

I'm sorry, but I don't understand this.
Would you like to refer to another plugin implementation?

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -98,24 +49,68 @@ void TizenAppManagerPlugin::HandleMethodCall(
...
+      },
+      (void *)this);
   if (ret != APP_MANAGER_ERROR_NONE) {
     char *err_msg = get_error_message(ret);
     LOG_ERROR("Failed app_manager_set_app_context_event_cb : %s", err_msg);
     events->Error("Failed to add callback", std::string(err_msg));
     return;
   }
-  m_registered_event_cb = true;
+  has_registered_event_ = true;
```
bbrto21, line 103: I think the `observer` is more appropriate for name than the `event` in context.

packages/tizen_app_manager/tizen/src/application_utils.cc
```diff
@@ -4,145 +4,194 @@
...
-        std::pair<flutter::EncodableValue, flutter::EncodableValue>(
-            std::string(key), std::string(value)));
-    LOG_INFO("AppMetaDataCB key: %s, value: %s", key, value);
-  }
-  return true;
-}
-
-int GetAppData(app_info_h app_info, flutter::EncodableMap &value) {
+namespace {
+char *GetAppId(app_info_h app_info, flutter::EncodableMap &value) {
```
bbrto21, line 8: Can you read [this document](https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace) about the vertical white space and apply it to this plug-in as a whole?
I think If you look at other plug-in implementations, you'll get help.

packages/tizen_app_manager/tizen/src/application_utils.cc
```diff
@@ -4,145 +4,194 @@
...
+          flutter::EncodableMap *metadata =
+              static_cast<flutter::EncodableMap *>(user_data);
+          metadata->insert(
+              std::pair<flutter::EncodableValue, flutter::EncodableValue>(
+                  std::string(key), std::string(value)));
+          LOG_INFO("AppMetaDataCB key: %s, value: %s", key, value);
+        }
+        return true;
+      },
+      (void *)&metadata);
```
bbrto21, line 150: Please don't use C-style cast.

packages/tizen_app_manager/tizen/src/application_utils.cc
```diff
@@ -4,145 +4,194 @@
...
+  try {
+    app_id = GetAppId(app_info, value);
+    GetPkgId(app_info, value);
+    GetLabel(app_info, value);
+    GetType(app_info, value);
+    GetIconPath(app_info, value);
+    GetExecutablePath(app_info, value);
+    GetSharedResourcePath(app_id, value);
+    GetIsNoDisplay(app_info, value);
+    GetForEachMetadata(app_info, value);
```
swift-kim, line 188: Why don't you make `AppInfo` into a class?

packages/tizen_app_manager/tizen/src/application_utils.cc
```diff
@@ -4,145 +4,194 @@
...
+        return true;
+      }
+    }
   }
-  if (exec_path) {
-    free(exec_path);
+  return false;
+}
+
+int GetAppData(app_info_h app_info, flutter::EncodableMap &value) {
```
swift-kim, line 176: https://github.com/flutter-tizen/plugins/issues/354

Ideally the type `flutter::EncodableMap` should not appear in this file. Consider moving `ExtractValueFromMap` and `GetAppData` to `tizen_app_manager_plugin.cc`.

swift-kim (changes requested):

packages/tizen_app_manager/tizen/src/application_utils.cc
Swanseo0: > The classes should have a clean design so that their roles can be easily known from their signatures. Ideally, the plugin class should be the only one that interacts with the underlying platform channel (processes method arguments and generates results), and other classes should not touch any Flutter API (flutter/*.h).

As you said, GetAppData() and ExtractValueFromMap() were moved to tizen_app_manager_plugin.cc.

packages/tizen_app_manager/tizen/src/application_utils.cc
Swanseo0: Created `AppData` class. This class supports `GetAppData ()` and stores the app_info and value mentioned as member variables.
It seems to have been modified in a different direction from the direction you reviewed. Please review the code again to see if the current code contains the intention of the review.

packages/tizen_app_manager/tizen/src/application_utils.cc
Swanseo0: It has been modified to use static_cast.
`(void *)&metadata` -> `static_cast<void *>(&metadata)`

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: modified to if(value) form. Thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: This has also been modified in the form of if (value). Thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: The string comparison part of HandleMethodCall() has been modified to a string comparison method using ==.

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
Swanseo0: I applied it partially using the method you mentioned. Thank you.

packages/tizen_app_manager/tizen/src/application_utils.cc
swift-kim: 1. `AppData` → `AppInfo`
2. Move the class into a separate file.
3. Remove the `GetAppData` method and the private member `flutter_map_`. You'll need to refactor the code a bit.
4. Do not use C++ exceptions.
5. Do not abbreviate words: `GetPkgId` → `GetPackageId`.

Please ask me or @bbrto21 offline if you need a rationale or detailed explanation.

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
Swanseo0: I agree that the `observer` is more appropriate than the `event`.
Is it correct that you gave the opinion because the flag is for `RegisterObserver()` and `UnregisterObserver()` operation?

packages/tizen_app_manager/tizen/src/application_utils.cc
Swanseo0: > Blank lines immediately inside a declaration of a namespace or block of namespaces may help readability by visually separating the load-bearing content from the (largely non-semantic) organizational wrapper. Especially when the first declaration inside the namespace(s) is preceded by a comment, this becomes a special case of the previous rule, helping the comment to "attach" to the subsequent declaration.

I added a blank line just below the namespace. thank you!

packages/tizen_app_manager/tizen/src/tizen_app_manager.h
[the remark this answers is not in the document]
Swanseo0: I think there was something I misunderstood. The other two variables have also been changed to private area.

packages/tizen_app_manager/tizen/src/application_utils.cc
Swanseo0: I applied the reviews mentioned.
1. `AppData`-> `TizenAppInfo`
2. The defined class is separated into a separate file. Tizen_app_info.h was created to distinguish from App_info.h, and the prefix Tizen was added to the class name.
3. Private Member `Flutter_map_` was removed from the class. It seems to be inappropriate because it was a private member, but it could be affected outside of the class.
In the review, you requested to remove the `GetAppdata` function, but we have made a modification.
4. The C ++ Exceptions I used was removed.
5. The part where the compression was used is modified.

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
bbrto21: yes!

swift-kim (approved):
Let me do some additional work based on this change.

packages/tizen_app_manager/tizen/src/tizen_app_info.h
```diff
@@ -5,38 +5,32 @@
...
+  std::string GetType();
+  std::string GetIconPath();
+  std::string GetExecutablePath();
+  std::string GetSharedResourcePath();
+  bool GetIsNoDisplay();
+  void GetForEachMetadata(app_info_metadata_cb callback, void *user_data);
 
+ private:
   app_info_h app_info_ = nullptr;
   char *app_id_ = nullptr;
```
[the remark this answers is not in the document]
Swanseo0: The reason I used `char *` instead of `std :: string` was that the first App_id data type was `char *`, and the data type of ㅁ`app_id_` required by the `app_manager_get_shared_resource_path()` using it was `char *`.
Also, I haven't thought about using `std :: string`.
As I think about it now, using `std :: string` will solve the problem with the position of `free(app_id_)` and the code will be cleaner.

packages/tizen_app_manager/tizen/src/tizen_app_info.cc
```diff
@@ -4,190 +4,126 @@
...
+  std::string result = std::string(executable_path);
+  free(executable_path);
+  return result;
 }
 
-bool TizenAppInfo::GetSharedResourcePath(flutter::EncodableMap &value) {
+std::string TizenAppInfo::GetSharedResourcePath() {
   char *shared_res_path = nullptr;
 
   int ret = app_manager_get_shared_resource_path(app_id_, &shared_res_path);
```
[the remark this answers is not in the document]
Swanseo0: The Tizen API used in this plugin was `app_manager_*`, `app_info_*`, `app_context_*`.
I declared and used classes that call each Tizen API, but I wondered if this is right. So I erased what I implemented and declared only the `TizenAppinfo` class.
The API mentioned in the form of `app_manager_*`, but I thought it was a data that I had to get when requesting application information and belonged to a large frame app info.

packages/tizen_app_manager/tizen/src/tizen_app_info.cc
```diff
@@ -4,190 +4,126 @@
...
-        if (key) {
-          flutter::EncodableMap *metadata =
-              static_cast<flutter::EncodableMap *>(user_data);
-          metadata->insert(FlValuePair(std::string(key), std::string(value)));
-          LOG_INFO("AppMetaDataCB key: %s, value: %s", key, value);
-        }
-        return true;
-      },
-      static_cast<void *>(&metadata));
+void TizenAppInfo::GetForEachMetadata(app_info_metadata_cb callback,
```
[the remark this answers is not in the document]
Swanseo0: If so ... `Tizenappinfo::GetForeachmetadata()` returned `std::map<std::string, std::string>` and then extracted information from the map and put information in `flutter::EncodableMap`?

packages/tizen_app_manager/tizen/src/tizen_app_manager_plugin.cc
```diff
@@ -4,316 +4,386 @@
...
+    }
   }
-}
 
-void TizenAppManagerPlugin::GetApplicationInfo(
-    const flutter::EncodableValue &arguments, MethodResultPtr result) {
-  std::string id = "";
-  char *err_msg = nullptr;
-  app_info_h app_info = nullptr;
+  void RegisterObserver(std::unique_ptr<FlEventSink> &&events) {
```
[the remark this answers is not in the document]
Swanseo0: `void SetupChannels (flutter::PluginRegistrar *registrar)`

Please also refactoring this method.
There were several questions in the method.

-The callback of `launch_event_channel_handler`, `terminal_event_channel_handler` may be performed twice in `std::move (events)` in the callback of Launch.
-`launch_event_channel`, `terminate_event_channel` is not stored
- When registering a callback in the same form as the `BatterystatusStreamHandler` class, how to handle `registered_cnt_`

These are the problems that could not be solved because i could not determine how to proceed ...

PR #409, opened 2022-06-16
Update Tizen package dependencies in pubspecs

swift-kim:
- Update the `tizen_app_control` version to 0.2.0 in all dependent packages.
- Let example apps resolve package dependencies from paths when possible.

[Here](https://github.com/flutter-tizen/plugins/commit/8f86c976d4ed2b2711469314f9d4a1fee6b60635) is a comparison of two strategies on how to manage internal package dependencies. Which way do you guys prefer?
- Explicit version: The versions should be manually updated, but is more robust to backward incompatible API changes.
- Relative path: All dependent packages should be updated at once whenever there's a breaking API change, but deprecated/obsolete API usage can be easily detected.

### DIFF
--- a/packages/audioplayers/example/pubspec.yaml
+++ b/packages/audioplayers/example/pubspec.yaml
@@ -10,8 +10,7 @@ dependencies:
     sdk: flutter
   http: ^0.13.1
   path_provider: ^2.0.1
-  path_provider_tizen:
-    path: ../../path_provider/
+  path_provider_tizen: ^2.0.1
   provider: ^5.0.0
 
 dev_dependencies:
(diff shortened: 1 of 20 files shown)


HakkyuKim (approved):
LGTM, I personally favor relative paths as packages that need update after breaking changes can be detected easily if CI builds for all packages (currently, merging to master branch only builds for changed packages). It will give us a chance to look over other packages when there's a breaking change to a dependent package.

bbrto21:
Could you please update CHANGELOG files?

bbrto21 (approved):

PR #426, opened 2022-08-25
[webview_flutter] Change the project type to sharedLib

swift-kim:
Based on https://github.com/flutter-tizen/flutter-tizen/pull/407.

With this change, the webview plugin will not be included in `libflutter_plugins.so` and instead `libwebview_flutter_tizen_plugin.so` will be generated and included in a TPK.

@flutter-tizen/maintainers The experimental "implicit user library linking" support for "staticLib" type plugins will be removed in the future. Please migrate to the "sharedLib" type if you have inhouse (or work-in-progress) plugins that use such an old configuration by referring to this change.

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,4 +1,4 @@
-## NEXT
+## 0.5.4
 
 * Change the project type to sharedLib.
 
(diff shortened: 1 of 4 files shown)


HakkyuKim (approved):
LGTM. Should be merged after #432 for CI to release it.

swift-kim:
https://github.com/flutter-tizen/plugins/runs/8245888702?check_suite_focus=true#step:4:157

> Connection closed before full header was received

It seems this CI error usually occurs for large packages (a few megabytes) and can be resolved by running `pub get` before the `pub publish` command. We need to update the `PublishPluginCommand` class in the plugin tools.

PR #427, opened 2022-08-25
Bump ffi dependencies

swift-kim:
This helps resolving dependencies of apps that have ffi ^2.0.0 as a dependency. We may drop support for ffi ^1.1.2 and only support ^2.0.0 in the future.


HakkyuKim (approved):
LGTM

PR #429, opened 2022-08-26
[tizen_app_control] Improve FFI symbol resolution

swift-kim:
We shouldn't dlopen the embedder library by its name because the soname may possibly have the "_nui" suffix (https://github.com/flutter-tizen/engine/pull/328).

This change simplifies DynamicLibrary loading and symbol resolution by assuming that the embedder has already been loaded by the running process.


HakkyuKim (approved):
LGTM

PR #433, opened 2022-09-02
[video_player] Add a queue for storing decoded media packets

xiaowei-guan:
For fixing frame tearing issue on RB5:
1.Should destory media packet in sequence.
2.Should not destory current rendering packet when rendering finished.

### DIFF
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.1
+
+* Implement a new message loop for storing media packet.
+
 ## 2.4.0
 
 * Update video_player to 2.4.2.
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 2.4.1
+## 2.4.2
 
 * Implement a new message loop for storing media packet.
+
+## 2.4.1
+
 * Apply new texture APIs.
 
 ## 2.4.0
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.4.2
 
-* Implement a new message loop for storing media packet.
+* Add a queue for storing decoded media packet.
 
 ## 2.4.1
 
--- a/packages/video_player/tizen/src/video_player.cc
+++ b/packages/video_player/tizen/src/video_player.cc
@@ -12,12 +12,6 @@
 #include "log.h"
 #include "video_player_error.h"
 
-struct Message {
-  Eina_Thread_Queue_Msg head;
-  MessageEvent event;
-  media_packet_h media_packet;
-};
-
 static std::string RotationToString(player_display_rotation_e rotation) {
   switch (rotation) {
     case PLAYER_DISPLAY_ROTATION_NONE:
@@ -263,17 +257,6 @@ void VideoPlayer::Dispose() {
   event_sink_ = nullptr;
   event_channel_->SetStreamHandler(nullptr);
 
-  if (player_) {
-    player_unprepare(player_);
-    player_unset_media_packet_video_frame_decoded_cb(player_);
-    player_unset_buffering_cb(player_);
-    player_unset_completed_cb(player_);
-    player_unset_interrupted_cb(player_);
-    player_unset_error_cb(player_);
-    player_destroy(player_);
-    player_ = 0;
-  }
-
   while (!packet_queue_.empty()) {
     media_packet_destroy(packet_queue_.front());
     packet_queue_.pop();
@@ -287,6 +270,16 @@ void VideoPlayer::Dispose() {
   if (previous_media_packet_) {
     media_packet_destroy(previous_media_packet_);
     previous_media_packet_ = nullptr;
+  }
+  if (player_) {
+    player_unprepare(player_);
+    player_unset_media_packet_video_frame_decoded_cb(player_);
+    player_unset_buffering_cb(player_);
+    player_unset_completed_cb(player_);
+    player_unset_interrupted_cb(player_);
+    player_unset_error_cb(player_);
+    player_destroy(player_);
+    player_ = 0;
   }
 
   if (texture_registrar_) {
--- a/packages/video_player/tizen/src/video_player.cc
+++ b/packages/video_player/tizen/src/video_player.cc
@@ -446,7 +446,7 @@ void VideoPlayer::OnError(int code, void *data) {
 void VideoPlayer::OnVideoFrameDecoded(media_packet_h packet, void *data) {
   auto *player = reinterpret_cast<VideoPlayer *>(data);
   std::lock_guard<std::mutex> lock(player->mutex_);
-  player->SendMessage(kMessageOnFrameDecoded, packet);
+  player->SendMessage(kMessageFrameDecoded, packet);
 }
 
 void VideoPlayer::RunMediaPacketLoop(void *data, Ecore_Thread *thread) {
@@ -470,7 +470,7 @@ void VideoPlayer::RunMediaPacketLoop(void *data, Ecore_Thread *thread) {
       break;
     }
 
-    if (message->event == kMessageOnFrameDecoded) {
+    if (message->event == kMessageFrameDecoded) {
       if (message->media_packet != nullptr) {
         packet_queue.push(message->media_packet);
       }
@@ -524,5 +524,5 @@ void VideoPlayer::SendRenderFinishedMessage() {
     media_packet_destroy(previous_media_packet_);
     previous_media_packet_ = nullptr;
   }
-  SendMessage(kMessageOnRenderFinished, nullptr);
+  SendMessage(kMessageRenderFinished, nullptr);
 }
--- a/packages/video_player/tizen/src/video_player.cc
+++ b/packages/video_player/tizen/src/video_player.cc
@@ -13,13 +13,9 @@
 #include "log.h"
 #include "video_player_error.h"
 
-constexpr int kMessageQuit = -1;
-constexpr int kMessageOnFrameDecoded = 0;
-constexpr int kMessageOnRenderFinished = 1;
-
 struct Message {
   Eina_Thread_Queue_Msg head;
-  int event;
+  MessageEvent event;
   media_packet_h media_packet;
 };
 
@@ -92,7 +88,6 @@ FlutterDesktopGpuBuffer *VideoPlayer::ObtainGpuBuffer(size_t width,
 VideoPlayer::VideoPlayer(flutter::PluginRegistrar *plugin_registrar,
                          flutter::TextureRegistrar *texture_registrar,
                          const std::string &uri, VideoPlayerOptions &options) {
-  is_initialized_ = false;
   texture_registrar_ = texture_registrar;
 
   texture_variant_ =
@@ -470,6 +465,7 @@ void VideoPlayer::RunMediaPacketLoop(void *data, Ecore_Thread *thread) {
     Message *message = static_cast<Message *>(
         eina_thread_queue_wait(packet_thread_queue, &ref));
     if (message->event == kMessageQuit) {
+      LOG_INFO("Message quit.");
       eina_thread_queue_wait_done(packet_thread_queue, ref);
       break;
     }
@@ -504,7 +500,7 @@ void VideoPlayer::RunMediaPacketLoop(void *data, Ecore_Thread *thread) {
   }
 }
 
-void VideoPlayer::SendMessage(int event, media_packet_h media_packet) {
+void VideoPlayer::SendMessage(MessageEvent event, media_packet_h media_packet) {
   if (!packet_thread_ || ecore_thread_check(packet_thread_)) {
     LOG_ERROR("Invalid packet thread.");
     return;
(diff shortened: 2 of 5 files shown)


packages/video_player/tizen/src/video_player.h
```diff
@@ -13,6 +14,7 @@
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <queue>
```
HakkyuKim, line 17: This may be moved to the implementation file.

packages/video_player/tizen/src/video_player.cc
```diff
@@ -423,13 +450,84 @@ void VideoPlayer::OnError(int code, void *data) {
...
-  player->texture_registrar_->MarkTextureFrameAvailable(player->texture_id_);
+
+  self->packet_thread_queue_ = packet_thread_queue;
+  std::queue<media_packet_h> packet_queue;
+  while (!ecore_thread_check(thread)) {
+    void *ref;
+    Message *message = static_cast<Message *>(
+        eina_thread_queue_wait(packet_thread_queue, &ref));
+    if (message->event == kMessageQuit) {
+      LOG_ERROR("Message quit.");
```
HakkyuKim, line 473: This isn't an error and may be removed.

packages/video_player/tizen/src/video_player.cc
```diff
@@ -79,6 +92,7 @@ VideoPlayer::VideoPlayer(flutter::PluginRegistrar *plugin_registrar,
                          flutter::TextureRegistrar *texture_registrar,
                          const std::string &uri, VideoPlayerOptions &options) {
   is_initialized_ = false;
+  is_rendering_ = false;
```
HakkyuKim, line 95: How about initializing them inline in header file?

HakkyuKim (commented):
Otherwise looks good to me.

packages/video_player/tizen/src/video_player.cc
```diff
@@ -8,6 +8,7 @@
 #include <flutter/standard_method_codec.h>
 
 #include <algorithm>
+#include <queue>
```
HakkyuKim, line 11: `#include <queue>` can be removed from `video_player.h`,

packages/video_player/tizen/src/video_player.cc
HakkyuKim: Did you decide to leave the `LOG_ERROR` message? I've suggested it because it's an expected flow of the program, not an error or exception.

packages/video_player/tizen/src/video_player.cc
```diff
@@ -8,10 +8,21 @@
...
 
 #include "log.h"
 #include "video_player_error.h"
 
+constexpr int kMessageQuit = -1;
+constexpr int kMessageOnFrameDecoded = 0;
+constexpr int kMessageOnRenderFinished = 1;
```
swift-kim, line 18: This is okay for now but you may consider creating a dedicated enum type for constants like this in future implementations. (I think you just followed the current `TizenVsyncWaiter` implementation.)

packages/video_player/tizen/src/video_player.cc
```diff
@@ -423,13 +450,83 @@ void VideoPlayer::OnError(int code, void *data) {
...
+  message->media_packet = media_packet;
+  eina_thread_queue_send_done(packet_thread_queue_, ref);
+}
+
+void VideoPlayer::SendRenderFinishedMessage() {
+  if (previous_media_packet_) {
+    media_packet_destroy(previous_media_packet_);
+    previous_media_packet_ = nullptr;
+  }
+  SendMessage(kMessageOnRenderFinished, nullptr);
```
swift-kim, line 531: What is the `kMessageOnRenderFinished` message for?

swift-kim (commented):
Please also add a patch note in the CHANGELOG.md file. If you don't want to increase the version, you can write:

```markdown
## NEXT

* ...

## 2.4.0
```

packages/video_player/tizen/src/video_player.cc
xiaowei-guan: Ok, I change to INFO

packages/video_player/tizen/src/video_player.cc
xiaowei-guan: I have removed it from header file.

packages/video_player/tizen/src/video_player.h
```diff
@@ -63,7 +62,7 @@ class VideoPlayer {
   media_packet_h current_media_packet_ = nullptr;
   media_packet_h previous_media_packet_ = nullptr;
   bool is_initialized_;
-  bool is_rendering_;
+  bool is_rendering_ = false;
```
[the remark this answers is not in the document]
xiaowei-guan: OK

packages/video_player/tizen/src/video_player.cc
xiaowei-guan: I have created message event enum.

packages/video_player/tizen/src/video_player.cc
xiaowei-guan: sent this message to request render if message queue not empty.

swift-kim: Hmm, in that case "OnRenderFinished" and "request rendering" do not match in their meaning. "OnRenderFinished" might need better naming.

Also is there any real world case that the packet queue is not empty when the "OnRenderFinished" message is sent? As far as I can see a media packet is queued when a new frame is decoded by the player and then it's immediately sent to the engine for rendering, so the queue should remain empty in most cases. Even if the `lock` is acquired again by `OnVideoFrameDecoded` before the thread acquires the `lock`, the frames will still be rendered in order and the queue will be emptied anyway.

xiaowei-guan: > Hmm, in that case "OnRenderFinished" and "request rendering" do not match in their meaning. "OnRenderFinished" might need better naming.
> 
> Also is there any real world case that the packet queue is not empty when the "OnRenderFinished" message is sent? As far as I can see a media packet is queued when a new frame is decoded by the player and then it's immediately sent to the engine for rendering, so the queue should remain empty in most cases. Even if the `lock` is acquired again by `OnVideoFrameDecoded` before the thread acquires the `lock`, the frames will still be rendered in order and the queue will be emptied anyway.

Also is there any real world case that the packet queue is not empty when the "OnRenderFinished" message is sent?
==>when decoding speed is faster than rendering speed.

swift-kim: I was not talking about the "On" prefix in the name. I was asking why the rendering "finished" message would "trigger" rendering.

> when decoding speed is faster than rendering speed.

Okay, but this still doesn't explain why the `kMessageRenderFinished` message is used for scheduling a single frame. Let's say the decoding speed is faster than the rendering speed and some `kMessageFrameDecoded` messages have been ignored due to the `is_rendering_` check. If a `kMessageRenderFinished` message is received by the thread in this situation, a packet will be dequeued from `packet_queue`, but even after that `packet_queue` will still remain non-empty if there were more than one packets in the queue. Is this what you expected? Please take more time to explain your code so that it can be understood by average readers.

xiaowei-guan: About render queue, we have three types event:
1. FrameDecoded event
  when receive decoded callback we will push the packet to packet queue.
```cpp
    if (message->event == kMessageFrameDecoded) {
      if (message->media_packet != nullptr) {
        packet_queue.push(message->media_packet);
      }
    }
```
  If there is no rendering operation at this time， we can request to render.
```cpp
    if (packet_queue.empty() || self->is_rendering_) {
      continue;
    }
    if (self->texture_registrar_->MarkTextureFrameAvailable(
            self->texture_id_)) {
      self->is_rendering_ = true;
      self->previous_media_packet_ = self->current_media_packet_;
      self->current_media_packet_ = packet_queue.front();
      packet_queue.pop();
    }
```
2. kMessageRenderFinished
we may receive frame decoded callback during rendering operation, we just push the packet to the packet queue, but not requet render. so after rendering, we can can send RenderFinished event to message loop, if  packet queue not empty, we need request to render.
```cpp
    if (packet_queue.empty() || self->is_rendering_) {
      continue;
    }
    if (self->texture_registrar_->MarkTextureFrameAvailable(
            self->texture_id_)) {
      self->is_rendering_ = true;
      self->previous_media_packet_ = self->current_media_packet_;
      self->current_media_packet_ = packet_queue.front();
      packet_queue.pop();
    }
```
3. kMessageQuit
destory message loop

swift-kim: Thank you for the explanation. So the render finished event is sent from the `ReleaseMediaPacket` callback to request rendering of the next packet in the queue (if any). In that case the name `SendRenderFinishedMessage` makes sense when used within the release callback but not in `ObtainGpuBuffer` because no rendering actually happened. This is the part that confused me.

Another concern is that if the decoding speed can be faster than the rendering speed, the queue may grow more and more over time and the app may encounter its memory limit. Is that possible? (It will also slow down the video playback and `GetPosition` will report a wrong value as a consequence.) If the packets must be destroyed in decoding order, how about destroying all stale packets in the queue (except the latest one) when `ReleaseMediaPacket` is called?

packages/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.4.1
 
 * Implement a new message loop for storing media packet.
+* Apply new texture APIs.
```
HakkyuKim, line 4: version 2.4.1 is already published, this needs to be published as 2.4.2.

packages/video_player/tizen/src/video_player.cc
xiaowei-guan: I asked MM team member, they said for HW decoder:
The number of output buffer is fixed, so decoded buffer will not be come if app keeps too many buffers not to rendered.
They suggest that plugin can control this situation by dropping new buffer decoded callback if too many buffers are queued.

swift-kim: > dropping new buffer decoded callback

Isn't this the way `OnVideoFrameDecoded` was implemented before this change?

xiaowei-guan: > > dropping new buffer decoded callback
> 
> Isn't this the way `OnVideoFrameDecoded` was implemented before this change?

Yes, but message queue solution has another difference, it doesn't destory current media packet when render finished,
These are MM team member guide me to do so.

swift-kim: I don't still clearly understand the solution but wouldn't the result be the same without the new thread (`RunMediaPacketLoop`)? i.e. `packet_queue.push` could be moved to `OnVideoFrameDecoded` and `RunMediaPacketLoop` may become a simple method call (might be named `RequestRendering`). `OnVideoFrameDecoded`, `ObtainGpuSurface`, and `ReleaseMediaPacket` are all called asynchronously and the thread seems to add unnecessary complexity on top of it.

xiaowei-guan: > I don't still clearly understand the solution but wouldn't the result be the same without the new thread (`RunMediaPacketLoop`)? i.e. `packet_queue.push` could be moved to `OnVideoFrameDecoded` and `RunMediaPacketLoop` may become a simple method call (might be named `RequestRendering`). `OnVideoFrameDecoded`, `ObtainGpuSurface`, and `ReleaseMediaPacket` are all called asynchronously and the thread seems to add unnecessary complexity on top of it.

Yes, you are right, actually we don't need create a new thread, I try update a new version and test it on RB5 and TV devices.

swift-kim (commented):
Thank you for the fix and overall looks good to me except for the following.

packages/video_player/tizen/src/video_player.h
```diff
@@ -5,6 +5,7 @@
 #ifndef FLUTTER_PLUGIN_VIDEO_PLAYER_H_
 #define FLUTTER_PLUGIN_VIDEO_PLAYER_H_
 
+#include <Ecore.h>
```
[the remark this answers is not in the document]
xiaowei-guan: Has been removed.

packages/video_player/tizen/src/video_player.h
```diff
@@ -13,10 +14,17 @@
...
+#include <queue>
 #include <string>
 
 #include "video_player_options.h"
 
+typedef enum {
+  kMessageQuit = 0,
+  kMessageFrameDecoded,
+  kMessageRenderFinished,
+} MessageEvent;
```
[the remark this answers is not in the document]
xiaowei-guan: Has been removed.

packages/video_player/tizen/src/video_player.cc
```diff
@@ -12,6 +12,12 @@
 #include "log.h"
 #include "video_player_error.h"
 
+struct Message {
```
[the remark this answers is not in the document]
xiaowei-guan: Has been removed.

packages/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.4.2
+
+* Implement a new message loop for storing media packet.
```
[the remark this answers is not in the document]
xiaowei-guan: Update

packages/video_player/tizen/src/video_player.cc
```diff
@@ -266,11 +274,21 @@ void VideoPlayer::Dispose() {
     player_ = 0;
   }
 
+  while (!packet_queue_.empty()) {
+    media_packet_destroy(packet_queue_.front());
+    packet_queue_.pop();
+  }
```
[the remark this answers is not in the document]
xiaowei-guan: Destory media packet before destory palyer.

swift-kim (approved):

PR #439, opened 2022-09-19
[path_provider] Add tizen_interop dependency

swift-kim:
Reimplement using the tizen_interop package.

Contributes to https://github.com/flutter-tizen/tizen_interop/issues/9.

PR #440, opened 2022-09-19
[shared_preferences] Add tizen_interop dependency

swift-kim:
Similar to https://github.com/flutter-tizen/plugins/pull/439.

Closes https://github.com/flutter-tizen/tizen_interop/issues/9.


WonyoungChoi (approved):
:+1:

PR #442, opened 2022-09-20
[webview_flutter] Update README

swift-kim:


packages/webview_flutter/README.md
```diff
@@ -31,30 +27,25 @@ dependencies:
...
   @override
   WebViewExampleState createState() => WebViewExampleState();
 }
 
 class WebViewExampleState extends State<WebViewExample> {
-  @override
-  void initState() {
-    super.initState();
-  }
-
```
bbrto21, line 47: If this part is no longer needed, why don'y you remove them from the example too?

swift-kim: Done!

bbrto21 (approved):

PR #450, opened 2022-09-27
[webview_flutter] Set runtime path to $ORIGIN

swift-kim:
Fixes runtime errors of plugins that depend on the webview_flutter plugin (such as google_maps_flutter).


bbrto21:
google_maps_flutter using this PR works fine on TW3 and emulators.

PR #452, opened 2022-09-27
Update README

swift-kim:
- Update the "Device limitations" section in README with minimum API level info.
- Update google_maps_flutter_tizen's README.

### DIFF
--- a/packages/google_maps_flutter/example/assets/map.html
+++ b/packages/google_maps_flutter/example/assets/map.html
@@ -1,25 +1,29 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <style>
-      html,body {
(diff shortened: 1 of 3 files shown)


packages/google_maps_flutter/example/assets/map.html
```diff
@@ -1,25 +1,29 @@
...
+  <div id="map"></div>
+  <script>
+    let map;
+    console.log("before load");
+  </script>
+  <script src="https://maps.googleapis.com/maps/api/js?key=">
+  </script>
+</body>
+
+</html>
```
swift-kim, line 29: Auto-formatted by VS Code.

bbrto21 (approved):

PR #449, opened 2022-09-27
Update tizen-manifest.xml files

swift-kim:
- Change the organization prefix in package and application IDs (`com.example` → `org.tizen`).
- Remove unnecessary `api-version` attributes from `ui-application` elements.
- Remove unnecessary `tv-info` elements.

PR #454, opened 2022-09-28
[tizen_bundle] General cleanups

swift-kim:
- Update outdated information in README.
- Remove redundant test cases.
- Simplify the example app.
- Export the `Uint8List` type from `tizen_bundle.dart`.
- Compute `Bundle.length`, `Bundle.isEmpty`, and `Bundle.isNotEmpty` more efficiently.
- Code formatting and cleanups.

### DIFF
--- a/packages/tizen_bundle/CHANGELOG.md
+++ b/packages/tizen_bundle/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.1
+
+* General cleanups.
+
 ## 0.1.0
 
 * Initial release.
--- a/packages/tizen_bundle/README.md
+++ b/packages/tizen_bundle/README.md
@@ -2,7 +2,7 @@
 
 [![pub package](https://img.shields.io/pub/v/tizen_bundle.svg)](https://pub.dev/packages/tizen_bundle)
 
-Tizen bundle APIs.
+Tizen [Data Bundle](https://docs.tizen.org/application/native/guides/app-management/data-bundles) APIs.
 
 ## Usage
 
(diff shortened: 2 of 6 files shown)


HakkyuKim (approved):
LGTM

bbrto21:
@swift-kim Using this PR on TV emulator, I can't quit the example using 'q' key. could you please cross check this?

HakkyuKim:
> Using this PR on TV emulator, I can't quit the example using 'q' key.

I tried it as well, it doesn't quit the app.

> I also got an error message "Failed to stop app"after integration test

I cannot reproduce this though, both using `flutter-tizen test integration_test` and `./tools/tools_runner.sh integration-test --packages tizen_bundle`.

bbrto21:
@HakkyuKim I used below command
```
flutter-tizen drive --driver=test_driver/integration_test.dart --target=integration_test/bundle_test.dart
```

HakkyuKim:
@bbrto21 
Ah, I can reproduce it with that command, thanks.

swift-kim:
@HakkyuKim @bbrto21 Nice catch. The error is a result of the following defects (which are not related to this change).

(1) The application ID (`org.tizen...`) of this package doesn't match its package ID (`com.example...`). This is [technically not allowed](https://docs.tizen.org/application/native/index#package-id-and-application-id) on Tizen but it seems no error actually happens during the installation of the package:

> The application ID must be the same as the package ID if the package contains a single application.

(2) Unlike the `app_launcher -k` command, the secure command `kill` takes a package ID as an argument instead of an application ID. I thought the two commands were equivalent but apparently it's not. The code is [here](https://github.com/flutter-tizen/flutter-tizen/blob/8bc9483c0aa9f65efdb7e4928c7434ebe6623159/lib/tizen_device.dart#L482-L483).

(1) has already been fixed by https://github.com/flutter-tizen/plugins/pull/449. I'll make a patch for (2) soon.

bbrto21 (approved):

PR #455, opened 2022-09-29
Hold memory of StreamHandlerError's member

bbrto21:
* Use a derived class to hold memory not guaranteed by StreamHandlerError.
* This patch prevents crashes caused by dereferences to invalid memory.
* Update version.

### DIFF
--- a/packages/battery_plus/CHANGELOG.md
+++ b/packages/battery_plus/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.1
+
+* Hold memory of StreamHandlerError's member to prevent crash.
+
 ## 1.1.0
 
 * Update battery_plus to 2.1.3.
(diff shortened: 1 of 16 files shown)


bbrto21:
This PR is workaround for https://github.com/flutter-tizen/plugins/pull/351#discussion_r847005275.
If the right solution is applied upstream, then I will revert it.

swift-kim:
You may consider applying this to messageport_tizen, tizen_audio_manager, video_player_tizen, and wearable_rotary as well. They will need a bit of refactoring.

bbrto21:
> You may consider applying this to messageport_tizen, tizen_audio_manager, video_player_tizen, and wearable_rotary as well. They will need a bit of refactoring.

I think packages you mentioned are too many to include this PR. 
it would be better to open a PR that performs refactoring and apply this for each. If you don't mind, I'll go ahead with other PRs.

swift-kim (approved):

PR #458, opened 2022-10-04
[wearable_rotary] Refactor the C++ code

bbrto21:
This patch includes:
* Simplify WearableRotaryPlugin Implementation.
* Use derived class from flutter::StreamHandler to set stream handlers.
* Use derived class from flutter::StreamHandlerError to hold references safely.

This change contribute #354

### DIFF
--- a/packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
+++ b/packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
@@ -8,15 +8,72 @@
 #include <flutter/encodable_value.h>
 #include <flutter/event_channel.h>
 #include <flutter/event_sink.h>
-#include <flutter/event_stream_handler_functions.h>
+#include <flutter/event_stream_handler.h>
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
 
 #include <memory>
+#include <string>
 
 #include "log.h"
 
-static constexpr char kChannelName[] = "flutter.wearable_rotary.channel";
+namespace {
+
+typedef flutter::EventChannel<flutter::EncodableValue> FlEventChannel;
+typedef flutter::EventSink<flutter::EncodableValue> FlEventSink;
+typedef flutter::StreamHandler<flutter::EncodableValue> FlStreamHandler;
+typedef flutter::StreamHandlerError<flutter::EncodableValue>
+    FlStreamHandlerError;
+
+class WearableRotaryStreamHandlerError : public FlStreamHandlerError {
+ public:
+  WearableRotaryStreamHandlerError(const std::string &error_code,
+                                   const std::string &error_message,
+                                   const flutter::EncodableValue *error_details)
+      : error_code_(error_code),
+        error_message_(error_message),
+        FlStreamHandlerError(error_code_, error_message_, error_details) {}
+
+ private:
+  std::string error_code_;
+  std::string error_message_;
+};
+
+class WearableRotaryStreamHandler : public FlStreamHandler {
+ public:
+  WearableRotaryStreamHandler() {}
+
+ protected:
+  std::unique_ptr<FlStreamHandlerError> OnListenInternal(
+      const flutter::EncodableValue *arguments,
+      std::unique_ptr<FlEventSink> &&events) override {
+    events_ = std::move(events);
+    Eina_Bool ret = eext_rotary_event_handler_add(RotaryEventCallBack, this);
+    if (ret == EINA_FALSE) {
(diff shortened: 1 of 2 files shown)


packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
```diff
@@ -5,18 +5,80 @@
 #include "wearable_rotary_plugin.h"
 
 #include <efl-extension/wearable/circle/efl_extension_rotary.h>
-#include <flutter/encodable_value.h>
```
swift-kim, line 8: Why removed

packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
```diff
@@ -5,18 +5,80 @@
...
 #include <flutter/event_channel.h>
 #include <flutter/event_sink.h>
-#include <flutter/event_stream_handler_functions.h>
+#include <flutter/event_stream_handler.h>
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
 
 #include <memory>
+#include <optional>
```
swift-kim, line 15: Why added

HakkyuKim:
@bbrto21 
Nice work on refactoring the package!

Before reviewing the changes, we may need to discuss about whether we're still supporting wearable devices as this package is only viable those type of devices.

packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
```diff
@@ -5,18 +5,80 @@
...
+
+ protected:
+  std::unique_ptr<FlStreamHandlerError> OnListenInternal(
+      const flutter::EncodableValue *arguments,
+      std::unique_ptr<FlEventSink> &&events) override {
+    events_ = std::move(events);
+    Eina_Bool ret = eext_rotary_event_handler_add(RotaryEventCallBack, this);
+    if (ret == EINA_FALSE) {
+      return std::make_unique<WearableRotaryStreamHandlerError>(
+          std::to_string(TIZEN_ERROR_UNKNOWN),
```
swift-kim, line 58:
```suggestion
          "Operation failed",
```

swift-kim (commented):
Otherwise looks good to me.

swift-kim:
@HakkyuKim Watch is supported.

- https://github.com/flutter-tizen/plugins#device-limitations
- https://github.com/flutter-tizen/flutter-tizen#supported-devices

If you want to discontinue the support, please open a dedicated issue in https://github.com/flutter-tizen/flutter-tizen/issues.

bbrto21:
> Before reviewing the changes, we may need to discuss about whether we're still supporting wearable devices as this package is only viable those type of devices.

That's right, I had the same concern. but since we are still officially supporting wearable devices, I stopped concerning about it soon.

packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
bbrto21: I removed it. thanks!

packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
bbrto21: my mistake, I was thought there is no place to use it directly. I add this line.

packages/wearable_rotary/tizen/src/wearable_rotary_plugin.cc
bbrto21: I applied it.

swift-kim: No, you didn't :<

bbrto21: oh, I misunderstood. this parameter is for error code not error description. why do you want to replace the error code with a message? (also I already search the error number for operation failed, but I couldn't find it in Tizen common error)

swift-kim: I think you're misunderstanding the error code parameter. An error code is a string that indicates the type of the error, such as "Invalid argument" or "Permission denied" (see https://github.com/flutter-tizen/plugins/issues/354). It isn't necessarily a number defined in the Tizen API.

bbrto21: ok I fixed it.

swift-kim (approved):

PR #462, opened 2022-10-11
[network_info_plus] Do not use deprecated APIs

swift-kim:
Update the example app and integration_test to not use iOS-specific permission methods (deprecated as of network_info_plus 2.3.2).

Fixes CI analyze failures.


HakkyuKim (approved):
LGTM

PR #464, opened 2022-10-13
[google_maps_flutter] Use completer to get map id

bbrto21:
* Map generation callback is an asynchronous call, so a completeer is required.
* This patch fixes test failures that occur on the device while testing.

### DIFF
--- a/packages/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Fix integration test failures
+
 ## 0.1.3
 
 * Resolve linter warnings.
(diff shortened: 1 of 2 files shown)


bwikbs (approved):
👏

swift-kim:
Why don't you update the test code (and etc.) to the latest?

https://github.com/flutter/plugins/blob/main/packages/google_maps_flutter/google_maps_flutter/example/integration_test/google_maps_test.dart

bbrto21:
@swift-kim 
Of course I considered it first, but I couldn't get the latest tests as is. because some extra work is needed to make the test successful and I don't have much time right now. I'll keep tracking this issue and do it later.

swift-kim (approved):

PR #466, opened 2022-10-13
[google_sign_in] Improve UI

HakkyuKim:
- ~~Images support different pixel densities.~~ Used svg.
- Use two text style, bodyLarge for normal texts and titleLarge for "verification url" and "user code".
- Shows consistent UI on TV, wearable, and mobile.


bbrto21:
I think it would be nice if you provide QR code with login URL in pop up.

bbrto21 (approved):
Looks good to me.

packages/google_sign_in/lib/src/device_flow_widget.dart
```diff
@@ -33,43 +33,102 @@ void showDeviceFlowWidget({
...
+          padding: const EdgeInsets.symmetric(horizontal: 15, vertical: 25),
+          decoration: BoxDecoration(
+              border: Border.all(color: Colors.grey),
+              borderRadius: const BorderRadius.all(Radius.circular(5))),
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            children: <Widget>[
+              Image.asset(
+                'packages/google_sign_in_tizen/assets/images/google.png',
+              ),
```
swift-kim, line 56: or maybe
```suggestion
              const Image(
                image: AssetImage(
                  'assets/images/google.png',
                  package: 'google_sign_in_tizen',
                ),
              ),
```

I don't know which is a preferred way.

By the way, can we use the logo without Google's approval?

packages/google_sign_in/lib/src/device_flow_widget.dart
```diff
@@ -33,43 +33,102 @@ void showDeviceFlowWidget({
...
-              },
-            ),
-          ],
+        scrollable: true,
+        contentPadding: const EdgeInsets.all(10),
+        content: Container(
+          padding: const EdgeInsets.symmetric(horizontal: 15, vertical: 25),
+          decoration: BoxDecoration(
+              border: Border.all(color: Colors.grey),
+              borderRadius: const BorderRadius.all(Radius.circular(5))),
```
swift-kim, line 50: Missing trailing comma.

```suggestion
              borderRadius: const BorderRadius.all(Radius.circular(5)),),
```

Also you could remove the border itself (and padding) and save some space on devices with small screens.

packages/google_sign_in/lib/src/device_flow_widget.dart
```diff
@@ -33,43 +33,102 @@ void showDeviceFlowWidget({
...
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            children: <Widget>[
+              Image.asset(
+                'packages/google_sign_in_tizen/assets/images/google.png',
+              ),
+              Container(
+                margin: const EdgeInsets.only(top: 15),
```
swift-kim, line 58: You could use `SizedBox(height: 15)` to avoid extra indentation (nesting with `Container`).

packages/google_sign_in/lib/src/device_flow_widget.dart
```diff
@@ -33,43 +33,102 @@ void showDeviceFlowWidget({
...
+                        'From a PC, phone, or tablet go to:',
+                        style: bodyStyle,
+                      ),
+                    ),
+                    const SizedBox(height: 5),
+                    FittedBox(
+                      child: Text(
+                        '$verificationUrl',
+                        style: titleStyle,
+                      ),
```
swift-kim, line 72: nit: You may prefer
```suggestion
                      child: Text('$verificationUrl', style: titleStyle),
```
when it fits on one line.

swift-kim (commented):
Only nitpicks.

packages/google_sign_in/lib/src/device_flow_widget.dart
HakkyuKim: > I don't know which is a preferred way.

I don't think they have a preferred way when using assets in the package itself:
- https://docs.flutter.dev/development/ui/assets-and-images#from-packages
- https://docs.flutter.dev/development/ui/assets-and-images#bundling-of-package-assets

But the first one (the one you suggested) is more clear on where the asset comes from, I'll fix it.

> By the way, can we use the logo without Google's approval?

I've actually read this document incorrectly and we can't use the "Google" logo without their permission, we can only use the "G" logo with the text "Sign In With" for buttons. I'll check the Google's [Brand Resource Center](https://about.google/brand-resource-center/brand-elements/).

HakkyuKim:
> I think it would be nice if you provide QR code with login URL in pop up later.

Thanks for the suggestion! I'll add it in the next update.

packages/google_sign_in/lib/src/device_flow_widget.dart
HakkyuKim: Done, I removed the border decoration and container padding but left the content padding for `AlertDialog`.

packages/google_sign_in/lib/src/device_flow_widget.dart
HakkyuKim: Done.

packages/google_sign_in/lib/src/device_flow_widget.dart
HakkyuKim: Done.

HakkyuKim:
> I think it would be nice if you provide QR code with login URL in pop up later.

I've updated the UI with QR code instead of Google logo due to brand use permission.

PR #438, opened 2022-09-16
[tizen_rpc_port] Add tizen_rpc_port package

upple:
Tizen applications can communicate with each other using `tizen_rpc_port` APIs,
which allows applications to send and receive data after a connection is established.

PR #467, opened 2022-10-18
Use the tools analyze command for code analysis

swift-kim:
Use the `tools_runner.sh analyze` command instead of `dart analyze` for analyzing Dart source code, to ensure that `dart pub get` is run for all example apps before the analysis.

c.f. This change is for the "analyze" workflow. f065d1f6658cb35ad824c1160b52cec24787b72c is for the "build" workflow.

Resolves the analyze error of https://github.com/flutter-tizen/plugins/pull/438.


HakkyuKim (approved):
LGTM

PR #481, opened 2022-11-17
[tizen_notification] Bump version to 0.2.0

swift-kim:
- Rename some classes and properties. For example,
  - `NotificationImage` → `NotificationIcons`
  - `Property` → `NotificationProperty`
  - `DisplayApplist` → `NotificationStyle`
- Move `TizenNotificationDetails` and enum types (`SoundType` and `VibrationType`) to `types.dart`.
- Set `NotificationProperty.disableAppLaunch` automatically if `appControl` is not set.
- Improve the documentation.
- Code cleanup and minor bug fixes.


packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
-  static const int ticker = 1 << 1;
-
-  /// Lock screen.
+  /// Display in the lock screen.
   static const int lock = 1 << 2;
 
-  /// Indicator.
-  static const int indicator = 1 << 3;
+  /// Display in the indicator area (the top of the screen).
+  static const int indicator = 1 << 1 | 1 << 3;
```
swift-kim, line 33: `NOTIFICATION_DISPLAY_APP_TICKER` and `NOTIFICATION_DISPLAY_APP_INDICATOR` do not work in the way they are [documented](https://docs.tizen.org/application/native/guides/notification/notifications/#indicator-notifications) and now they are a single property.
- `NOTIFICATION_DISPLAY_APP_INDICATOR` has no effect on mobile (only `NOTIFICATION_DISPLAY_APP_TICKER` works in an unexpected way).
- `NOTIFICATION_DISPLAY_APP_TICKER` is effectively the same as `NOTIFICATION_DISPLAY_APP_INDICATOR` on RPi.

packages/tizen_notification/tizen/src/tizen_notification_plugin.cc
```diff
@@ -40,478 +43,389 @@ class TizenNotificationPlugin : public flutter::Plugin {
...
-            flutter::EncodableValue extra_value;
-            std::string key;
-            ret = app_control_create(&app_control);
-            if (ret != APP_CONTROL_ERROR_NONE) {
-              FreeNotification(noti_handle);
-              LOG_ERROR("app_control_create failed : %s",
+      flutter::EncodableMap sound;
+      if (GetValueFromEncodableMap(arguments, "sound", sound)) {
+        std::string type;
+        GetValueFromEncodableMap(&sound, "type", type);
```
JSUYA, line 249: This case and the "vibration" case do not handle false return.
Is it an intentional?
(use `NOTIFICATION_{SOUND/VIBRATION}}_TYPE_NONE` for invalid argument(?))
If not, how about adding a casefor Invalid argument?

packages/tizen_notification/tizen/src/tizen_notification_plugin.cc
```diff
@@ -40,478 +43,389 @@ class TizenNotificationPlugin : public flutter::Plugin {
...
-
-  bool GetEncodableValueFromArgs(const flutter::EncodableValue *args,
-                                 const char *key,
-                                 flutter::EncodableValue &out) {
-    if (std::holds_alternative<flutter::EncodableMap>(*args)) {
-      flutter::EncodableMap map = std::get<flutter::EncodableMap>(*args);
-      if (map.find(flutter::EncodableValue(key)) != map.end()) {
-        out = map[flutter::EncodableValue(key)];
+  static bool GetValueFromEncodableMap(const flutter::EncodableMap *map,
+                                       const char *key, T &out) {
```
JSUYA, line 47: We usually declare this function globally with an anonymous namespace.
How about moving?

https://github.com/flutter-tizen/plugins/issues/354 (Enclose the plugin class within an anonymous namespace)

packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
+  /// The path to the user vibration data file.
+  ///
+  /// Only applicable if the [type] is [VibrationType.userData].
   String? path;
 
-  /// Returns [NotificationVibration] member fields in a map format.
-  Map<String, dynamic> toMap() => <String, dynamic>{
+  /// Converts to a map.
+  Map<String, Object?> toMap() => <String, Object?>{
         'type': type.name,
```
bbrto21, line 124: If you use the actual values in `notification_type.h` ​​corresponding to each enumerable value instead of `name`, you don't have to convert name string to enum in native imple.

packages/tizen_notification/tizen/src/tizen_notification_plugin.cc
swift-kim: Not setting the `sound` argument or setting it to null will result in `notification_set_sound` not being called and the notification module will fall back to its default value (`NOTIFICATION_SOUND_TYPE_NONE` according to the documentation). Currently if the `NotificationSound` property is not set in the Dart side, a null value will be passed (it's a valid input) and the default sound will be used. So yes, this is an intended behavior.

packages/tizen_notification/tizen/src/tizen_notification_plugin.cc
swift-kim: Oh, I just forgot. Thank you for reminding.

packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
   static const int lock = 1 << 2;
 
-  /// Indicator.
-  static const int indicator = 1 << 3;
+  /// Display in the indicator area (the top of the screen).
+  static const int indicator = 1 << 1 | 1 << 3;
 
-  /// All display application.
+  /// All of the above.
   static const int all = (1 << 4) - 1;
```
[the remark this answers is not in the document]
swift-kim: Actually the `NOTIFICATION_DISPLAY_APP_ALL` constant already exists in `notification_type.h` and I don't exactly know what the original author's intention was, but anyway I like this idea and I will apply your suggestion.

packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
-import 'enums.dart';
+export 'package:tizen_app_control/tizen_app_control.dart';
 
-/// Properties that configure how the system handles notification at
-/// certain scenarios.
-class Property {
-  /// Display only SIM card inserted.
-  static const int displayOnlySimMode = 1 << 0;
+/// How the system handles the notification in certain scenarios.
+class NotificationProperty {
```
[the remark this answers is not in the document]
swift-kim: Sure.

packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
+  /// The path to the lock screen icon file.
+  final String? lockIcon;
 
-  /// Returns [NotificationImage] member fields in a map format.
-  Map<String, dynamic> toMap() => <String, dynamic>{
-        'icon': iconPath,
-        'iconForIndicator': indicatorPath,
-        'iconForLock': lockPath,
+  /// Converts to a map.
+  Map<String, Object?> toMap() => <String, Object?>{
```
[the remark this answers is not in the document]
swift-kim: Okay. No problem.

packages/tizen_notification/lib/src/types.dart
```diff
@@ -2,102 +2,189 @@
...
+  final AppControl? appControl;
+
+  /// Converts to a map.
+  Map<String, Object?> toMap() => <String, Object?>{
+        'image': icons?.toMap(),
+        'sound': sound?.toMap(),
+        'vibration': vibration?.toMap(),
+        'properties': appControl != null
+            ? properties
+            : properties | NotificationProperty.disableAppLaunch,
```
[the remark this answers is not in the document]
swift-kim: I changed the code to set the property from outside of the `toMap` method (in `TizenNotificationPlugin.show`) for clarity. How do you think?

packages/tizen_notification/lib/src/types.dart
swift-kim: I personally think it's not a good practice to use raw native constants in Dart code as I mentioned above.

packages/tizen_notification/lib/src/types.dart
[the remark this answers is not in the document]
bbrto21: I think it's much better :)

packages/tizen_notification/lib/src/types.dart
bbrto21: Ok, I agree with you.

bbrto21 (approved):
LGTM

JSUYA (approved):
LGTM

PR #483, opened 2022-11-17
[google_maps_flutter] Change dependency to webview_flutter_lwe

bbrto21:
* This plugin still depends on a lightweight web engine.


bwikbs (approved):
Good!

PR #486, opened 2022-11-22
Disable permission popups during testing

swift-kim:
Disable permission popups when testing on a wearable emulator, by creating the `askuser_disable` file. This enables testing of `path_provider_tizen` and `permission_handler_tizen`, but not `geolocator_tizen` because the location service is not available on emulator.

### DIFF
--- a/.github/recipe.yaml
+++ b/.github/recipe.yaml
@@ -6,10 +6,12 @@ plugins:
   flutter_app_badger: ["wearable-5.5"]
   flutter_secure_storage: ["wearable-5.5", "tv-6.5"]
   flutter_tts: ["wearable-5.5", "tv-6.5"]
+  google_maps_flutter: ["wearable-5.5", "tv-6.5"]
   integration_test: ["wearable-5.5", "tv-6.5"]
   messageport: ["wearable-5.5", "tv-6.5"]
   package_info_plus: ["wearable-5.5", "tv-6.5"]
-  path_provider: ["tv-6.5"]
+  path_provider: ["wearable-5.5", "tv-6.5"]
+  permission_handler: ["wearable-5.5"]
   sensors_plus: ["wearable-5.5"]
   shared_preferences: ["wearable-5.5", "tv-6.5"]
   sqflite: ["wearable-5.5", "tv-6.5"]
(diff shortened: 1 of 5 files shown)


tools/lib/src/device.dart
```diff
@@ -272,6 +272,45 @@ class EmulatorDevice extends Device {
...
+      <String>['-s', serial!, 'root', 'on'],
+    );
+    if (result.exitCode != 0) {
+      print('Error: running "sdb root on" failed.');
+      return false;
+    }
+
+    result = _processRunner.runSync(
+      _tizenSdk.sdb.path,
+      <String>['-s', serial!, 'shell', 'touch', '/opt/share/askuser_disable'],
```
swift-kim, line 299: This cannot be done in driver scripts because the file (`askuser_disable`) must be created even before the app is installed.

.github/recipe.yaml
```diff
@@ -21,26 +23,25 @@ plugins:
...
   # Only testable with the drive command: https://github.com/flutter-tizen/plugins/issues/272
   tizen_app_control: []
 
   # Test frequently breaks due to memory issue: https://github.com/flutter-tizen/plugins/issues/135
   video_player: []
+
+  # Temporarily disabled due to test_api error. Testable with the drive command.
+  # https://github.com/flutter-tizen/plugins/pull/397#issuecomment-1139299838
+  webview_flutter: []
+  webview_flutter_lwe: []
```
swift-kim, line 47: I don't still understand why this error occurs. Any help?

Might be related to https://github.com/dart-lang/test/pull/1314?

.github/recipe.yaml
```diff
@@ -6,10 +6,12 @@ plugins:
   flutter_app_badger: ["wearable-5.5"]
   flutter_secure_storage: ["wearable-5.5", "tv-6.5"]
   flutter_tts: ["wearable-5.5", "tv-6.5"]
+  google_maps_flutter: ["wearable-5.5", "tv-6.5"]
   integration_test: ["wearable-5.5", "tv-6.5"]
   messageport: ["wearable-5.5", "tv-6.5"]
   package_info_plus: ["wearable-5.5", "tv-6.5"]
-  path_provider: ["tv-6.5"]
+  path_provider: ["wearable-5.5", "tv-6.5"]
+  permission_handler: ["wearable-5.5", "tv-6.5"]
```
JSUYA, line 14: I have a question.
permission_handler is not support tv profile.(Not applicable for TV)
Is it necessary to test?

swift-kim: permission_handler_tizen is "technically" supported on TV (⚠️) but no testing should be necessary. Let me fix.

JSUYA (approved):

PR #488, opened 2022-11-30
[webview_flutter] Prevent invalid access

JSUYA:
result_value parameter of Ewk_View_Script_Execute_Cb can be null.
So, add a null check to prevent crashes.

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -656,11 +656,13 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
 
 void WebView::OnEvaluateJavaScript(Evas_Object* obj, const char* result_value,
                                    void* user_data) {
+  FlMethodResult* result = static_cast<FlMethodResult*>(user_data);
   if (result_value) {
-    FlMethodResult* result = static_cast<FlMethodResult*>(user_data);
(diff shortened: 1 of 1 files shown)


WonyoungChoi (approved):

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -656,9 +656,11 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
...
 void WebView::OnEvaluateJavaScript(Evas_Object* obj, const char* result_value,
                                    void* user_data) {
-  FlMethodResult* result = static_cast<FlMethodResult*>(user_data);
-  result->Success(flutter::EncodableValue(result_value));
-  delete result;
+  if (result_value) {
+    FlMethodResult* result = static_cast<FlMethodResult*>(user_data);
+    result->Success(flutter::EncodableValue(result_value));
+    delete result;
+  }
```
swift-kim, line 663: Obviously this will cause a leak. Should be something like this:

```cpp
  FlMethodResult* result = static_cast<FlMethodResult*>(user_data);
  if (result_value) {
    result->Success(flutter::EncodableValue(result_value));
  } else {
    result->Success();
  }
  delete result;
```

swift-kim (changes requested):

swift-kim (approved):

PR #493, opened 2022-12-07
[image_picker] Use video_player_tizen plugin

JSUYA:
Use the tizen plugin package for the video_player of the example.


swift-kim:
https://github.com/flutter-tizen/plugins/tree/master/packages/image_picker#supported-apis

This plugin doesn't support picking video files so actually there's no need for adding the dependency (since it effectively increases the total compile time). Do you have any device on which you can pick video files?

JSUYA:
> https://github.com/flutter-tizen/plugins/tree/master/packages/image_picker#supported-apis
> 
> This plugin doesn't support picking video files so actually there's no need for adding the dependency (since it effectively increases the total compile time). Do you have any device on which you can pick video files?

I did not check that video is not supported. We don't need to have dependencies if we don't need them. so i will close this.
(I tested mobile device. I know the mobile target is not a consideration. I just thought that using the tizen package would be a priority in our examples.)

PR #497, opened 2022-12-12
Run integration test once again on temporary failure

swift-kim:
Tests may temporarily fail for unknown reason on a self-hosted runner. Try again by passing the verbose (`-v`) option so that the false alarm does not disrupt the subsequent (release) workflow.

### DIFF
--- a/tools/lib/src/device.dart
+++ b/tools/lib/src/device.dart
@@ -127,8 +127,9 @@ class Device {
   /// nor [PackageResult.exclude].
   Future<PackageResult> runIntegrationTest(
     Directory workingDir,
-    Duration timeout,
-  ) async {
+    Duration timeout, {
+    bool debug = false,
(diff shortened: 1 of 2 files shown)


tools/lib/src/device.dart
```diff
@@ -127,16 +127,23 @@ class Device {
...
     if (!isConnected) {
       return PackageResult.fail(
           <String>['Device $name ($profile) is not connected.']);
     }
 
     final io.Process process = await _processRunner.start(
       'flutter-tizen',
-      <String>['-d', serial!, 'test', 'integration_test'],
+      <String>[
+        if (debug) '-v',
```
swift-kim, line 141: Temporarily added for debugging purpose. May be removed in the future.

JSUYA (approved):

PR #498, opened 2022-12-14
[firebase_functions] Add Firebase Cloud Functions plugin

wiertel:


swift-kim:
Regarding the analyze workflow failure: https://github.com/flutter-tizen/plugins/pull/502#issuecomment-1354998485

PR #500, opened 2022-12-16
[webview_flutter] General improvements

swift-kim:
- Change the project type to staticLib.
- Enable back key navigation.
- Implement `WebSettings.javascriptMode` and `WebViewController.clearCache`.
- Redirect console messages to stdout/stderr for easy debugging.
- Minor code cleanups.

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -10,6 +10,8 @@
 #include <flutter_texture_registrar.h>
 #include <tbm_surface.h>
 
+#include <ostream>
+
 #include "buffer_pool.h"
 #include "ewk_internal_api_binding.h"
 #include "log.h"
@@ -24,8 +26,7 @@ typedef flutter::MethodChannel<flutter::EncodableValue> FlMethodChannel;
 constexpr size_t kBufferPoolSize = 5;
 constexpr char kEwkInstance[] = "ewk_instance";
 
-class NavigationRequestResult
-    : public flutter::MethodResult<flutter::EncodableValue> {
+class NavigationRequestResult : public FlMethodResult {
  public:
   NavigationRequestResult(WebView* webview) : webview_(webview) {}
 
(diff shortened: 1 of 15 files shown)


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -190,14 +197,18 @@ void WebView::ApplySettings(const flutter::EncodableMap& settings) {
           has_progress_tracking_ = std::get<bool>(value);
         }
       } else if (string_key == "debuggingEnabled") {
+        // Not supported on Tizen.
       } else if (string_key == "gestureNavigationEnabled") {
+        // Not implemented.
       } else if (string_key == "allowsInlineMediaPlayback") {
+        // Not applicable for Tizen (always allowed).
```
swift-kim, line 204: This is true for the lightweight web engine. Is this also true for EWK?

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -339,23 +356,20 @@ void WebView::InitWebView() {
...
-  if (manager) {
+  Ewk_Cookie_Manager* cookie_manager = ewk_context_cookie_manager_get(context);
+  if (cookie_manager) {
     ewk_cookie_manager_accept_policy_set(
-        manager, EWK_COOKIE_ACCEPT_POLICY_NO_THIRD_PARTY);
+        cookie_manager, EWK_COOKIE_ACCEPT_POLICY_NO_THIRD_PARTY);
   }
-  EwkInternalApiBinding::GetInstance().settings.ImePanelEnabledSet(settings,
-                                                                   true);
-  ewk_settings_javascript_enabled_set(settings, true);
```
JSUYA, line 352: Is javascript enabled still true after deleting this line?

JSUYA (commented):
Shouldn't `int color`(in 150 line) be changed to int32_t as well?

packages/webview_flutter/tizen/src/webview.cc
swift-kim: Yes, and actually it's not deleted. `ApplySettings` is always called during the webview initialization and `ewk_settings_javascript_enabled_set` is called there.

swift-kim:
@JSUYA Either way should be okay but `int32_t` seems to make more sense since the value is expected to be exactly 32-bits long.

JSUYA (approved):

packages/webview_flutter/tizen/src/webview.cc
bbrto21: I think this feature only for iOS. and as far as I know, most weview implementations are allowed inline media playback by default. so I think you don't have to care about it.

bbrto21 (approved):
LGTM

PR #502, opened 2022-12-16
[firebase_core] Minor cleanups

swift-kim:


wiertel (approved):

swift-kim:
I just noticed that there's a common [`analysis_options.yaml`](https://github.com/invertase/flutterfire_desktop/blob/main/analysis_options.yaml) for Firebase Desktop packages. Our policy when implementing 3rd-party packages is to respect "their" analysis options and copy their `analysis_options.yaml` to the package root ([example](https://github.com/flutter-tizen/plugins/blob/master/packages/flutter_secure_storage/analysis_options.yaml)). However, as we are eventually going to have multiple FlutterFire plugin implementations in this repo, we could also create a common analysis options file in the repo root just like the [plus plugins'](https://github.com/flutter-tizen/plugins/blob/master/analysis_options_plus.yaml) case. 

So there are three options:

1. No change ("their" code should follow "our" analysis options when being copied to this repo).
2. Let each FlutterFire package in this repo have its own `analysis_options.yaml` and `all_lint_rules.yaml` files.
3. Create a common `analysis_options_firebase.yaml` in the repo root and import it from each FlutterFire package.

I'll go with the first option if you guys don't have a specific idea.

cc @wiertel

PR #508, opened 2023-01-04
[camera] Treat permission request as a success on TV

bbrto21:
* TV does not support the privilege API, which causes a crash.
* This patch resolves this crash. but it still doesn't support TV.

### DIFF
--- a/packages/camera/tizen/src/permission_manager.cc
+++ b/packages/camera/tizen/src/permission_manager.cc
@@ -6,11 +6,9 @@
 
 #ifndef TV_PROFILE
 #include <privacy_privilege_manager.h>
-#endif
 
 #include "log.h"
 
-#ifndef TV_PROFILE
 const char* PermissionToString(Permission permission) {
   switch (permission) {
     case Permission::kCamera:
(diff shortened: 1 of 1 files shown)


packages/camera/tizen/src/permission_manager.cc
```diff
@@ -4,10 +4,13 @@
 
 #include "permission_manager.h"
 
+#ifndef TV_PROFILE
 #include <privacy_privilege_manager.h>
+#endif
 
 #include "log.h"
 
+#ifndef TV_PROFILE
```
swift-kim, line 13: `log.h` is only used by `PermissionToString` so maybe these `#endif` and `#ifndef` are redundant?

packages/camera/tizen/src/permission_manager.cc
```diff
@@ -78,4 +84,5 @@ void PermissionManager::RequestPermission(Permission permission,
     }
   }
   return;
+#endif  // TV_PROFILE
```
swift-kim, line 87: The return statements could be removed.

swift-kim (approved):
The change looks good to me, but please also consider refactoring the whole code in this file in accordance with the implementation of other plugins (geolocator_tizen, image_picker_tizen, and permission_handler_tizen). Their `permission_manager.cc` are almost identical to each other's where `RequestPermission` is a blocking call and takes no callback as an argument.

bbrto21:
@swift-kim Thanks you for review. I'll gradually refactor this plugin.

PR #509, opened 2023-01-05
[tizen_rpc_port] Support explicitly stopping a server

swift-kim:
Part of https://github.com/flutter-tizen/plugins/issues/504.

Calling `StubBase.close` will destroy the underlying native handle so no additional operation can be made to the stub afterwards.


upple (approved):
LGTM 👍

PR #513, opened 2023-01-16
[tizen_audio_manager] Refactor the C++ code

JSUYA:

- Hiding tizen api in plugin.cc
- Minimize use of output parameters
  - Use c++ exception
- Use == instead of .compare()
- Use GetValueFromEncodableMap() to extract arguments from an EncodableMap
- int -> int32_t
- Remove unnecessary LOG_INFO()


packages/tizen_audio_manager/tizen/src/audio_manager.h
```diff
@@ -0,0 +1,42 @@
...
+  void SetLevel(std::string type, int32_t volume);
+
+  int32_t GetLevel(std::string type);
+
+  static void OnVolumeChanged(sound_type_e type, unsigned int volume,
+                              void *user_data);
+
+  void AddVolumeChangedCallback(VolumeChangedListener volume_changed_listener);
+
+  void RemoveVolumeChangedCallback();
```
swift-kim, line 34: The name `Add...` implies that the method can be called multiple times but in reality it will be a memory leak. You might rename and merge these to a single method `SetVolumeChangedHandler` to avoid any misunderstanding. You can also remove the extra `registered_volume_changed_cb_` flag by moving the implementation of the methods to the constructor and destructor of the class if you want (please refer to similar patterns used by other plugins such as [tizen_app_manager.h](https://github.com/flutter-tizen/plugins/blob/master/packages/tizen_app_manager/tizen/src/tizen_app_manager.h)).

packages/tizen_audio_manager/tizen/src/audio_manager.h
```diff
@@ -0,0 +1,42 @@
...
+  std::string getCurrentPlaybackType();
+
+  int32_t getMaxLevel(std::string getMaxLevel);
+
+  void SetLevel(std::string type, int32_t volume);
+
+  int32_t GetLevel(std::string type);
+
+  static void OnVolumeChanged(sound_type_e type, unsigned int volume,
+                              void *user_data);
```
swift-kim, line 30: This could be a private method.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -10,73 +10,38 @@
...
-  for (; iter != kSoundTypes.end(); iter++) {
-    if (iter->first == str) {
-      type = iter->second;
-      return true;
-    }
-  }
-
-  return false;
-}
+typedef flutter::MethodCall<flutter::EncodableValue> FlMethodCall;
```
swift-kim, line 22: Please add a blank line at the beginning of the namespace block.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -10,73 +10,38 @@
...
   return false;
 }
-
 }  // namespace
 
 class AudioManagerTizenPlugin : public flutter::Plugin {
  public:
-  using MethodResultPtr =
-      std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>>;
+  using MethodResultPtr = std::unique_ptr<FlMethodResult>;
```
swift-kim, line 44: This alias might be removed.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -89,180 +54,123 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
-    if (method_call.method_name().compare("getCurrentPlaybackType") == 0) {
-      GetCurrentPlaybackType(std::move(result));
-    } else if (method_call.method_name().compare("getMaxLevel") == 0) {
-      GetMaxLevel(arguments, std::move(result));
-    } else if (method_call.method_name().compare("setLevel") == 0) {
-      SetLevel(arguments, std::move(result));
-    } else if (method_call.method_name().compare("getLevel") == 0) {
-      GetLevel(arguments, std::move(result));
+    if (result_) {
+      result_->Error("Already active", "Cancelled by a second request.");
```
swift-kim, line 62: cancelled → canceled

Also only synchronous functions are used in the implementation so I doubt if this check and the extra variable (`result_`) are necessary.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -89,180 +54,123 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
-    }
-
-    if (!ConvertStringToSoundType(type_str, type)) {
-      LOG_ERROR("Failed to convert string to sound type.");
-      result->Error("UnknownSoundType",
-                    "Failed to convert string to sound type.");
+  void GetMaxLevel(const flutter::EncodableMap *arguments) {
+    std::string type;
+    if (!GetValueFromEncodableMap(arguments, "type", type)) {
+      result_->Error("InvalidArguments", "Please check type");
```
swift-kim, line 121: nit: Not mandatory but I usually follow this convention.
```suggestion
      result_->Error("Invalid arguments", "No type provided.");
```

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -89,180 +54,123 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
+      result_->Success(flutter::EncodableValue(max));
+    } catch (const AudioManagerError &error) {
+      result_->Error(error.code(), error.message());
     }
   }
 
-  void SetLevel(const flutter::EncodableValue &arguments,
-                MethodResultPtr result) {
+  void SetLevel(const flutter::EncodableMap *arguments) {
     std::string type_str;
```
swift-kim, line 134:
```suggestion
    std::string type;
```

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -89,180 +54,123 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
-    if (ret == SOUND_MANAGER_ERROR_NONE) {
-      result->Success(flutter::EncodableValue(volume));
-    } else {
-      const char *message = get_error_message(ret);
-      LOG_ERROR("Failed to get volume: %s", message);
-      result->Error(std::to_string(ret), "Failed to get volume.",
-                    flutter::EncodableValue(std::string(message)));
+    try {
+      int32_t volume;
+      volume = audio_manager_.GetLevel(type_str);
```
swift-kim, line 159:
```suggestion
      int32_t volume = audio_manager_.GetLevel(type);
```

packages/tizen_audio_manager/tizen/src/audio_manager.h
```diff
@@ -0,0 +1,42 @@
...
+class AudioManager {
+ public:
+  AudioManager() {}
+  ~AudioManager() {}
+
+  std::string getCurrentPlaybackType();
+
+  int32_t getMaxLevel(std::string getMaxLevel);
+
+  void SetLevel(std::string type, int32_t volume);
```
swift-kim, line 25: nit: volume vs. level vs. volume level?

We should use consistent wording as much as possible.

FYI) This plugin has been named after the C# class [AudioManager](https://samsung.github.io/TizenFX/stable/api/Tizen.Multimedia.AudioManager.html) (and [AudioVolume](https://samsung.github.io/TizenFX/stable/api/Tizen.Multimedia.AudioVolume.html)) but this C++ `AudioManager` class is actually a wrapper around the native Sound Manager module. "volume" may be preferred over "level" if you want to follow the native module's naming convention. However please note that this has nothing to do with the names in the `AudioManagerTizenPlugin` class.

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -273,41 +181,21 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
-
-  std::unique_ptr<flutter::EventSink<flutter::EncodableValue>> events_;
-  bool volume_changed_cb_is_registered_ = false;
-  int volume_changed_cb_id_ = 0;
+  std::unique_ptr<FlMethodResult> result_;
+  std::unique_ptr<FlEventSink> events_;
+  AudioManager audio_manager_;
 };
 
 void AudioManagerTizenPluginRegisterWithRegistrar(
```
swift-kim, line 201: nit: Maybe `AudioManagerTizen` → `TizenAudioManager`

(`pubspec.yaml` should be updated as well if you rename the function and the class.)

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: done

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
```diff
@@ -273,41 +181,21 @@ class AudioManagerTizenPlugin : public flutter::Plugin {
...
               return nullptr;
             },
             [this](const flutter::EncodableValue *arguments)
                 -> std::unique_ptr<flutter::StreamHandlerError<>> {
-              LOG_INFO("OnCancel");
               this->UnregisterObserver();
               return nullptr;
             });
 
     event_channel->SetStreamHandler(std::move(event_channel_handler));
```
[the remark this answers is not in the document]
JSUYA: I added TizenAudioManagerStreamHandler class.

packages/tizen_audio_manager/tizen/src/audio_manager.h
```diff
@@ -0,0 +1,42 @@
...
+#ifndef FLUTTER_PLUGIN_AUDIO_MANAGER_H_
+#define FLUTTER_PLUGIN_AUDIO_MANAGER_H_
+
+#include <sound_manager.h>
+
+#include <functional>
+#include <string>
+
+using VolumeChangedListener =
+    std::function<void(std::string type_str, int32_t volume)>;
```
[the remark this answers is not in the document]
JSUYA: done

packages/tizen_audio_manager/tizen/src/audio_manager.h
```diff
@@ -0,0 +1,42 @@
...
+
+class AudioManager {
+ public:
+  AudioManager() {}
+  ~AudioManager() {}
+
+  std::string getCurrentPlaybackType();
+
+  int32_t getMaxLevel(std::string getMaxLevel);
```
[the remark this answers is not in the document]
JSUYA: done

packages/tizen_audio_manager/tizen/src/audio_manager.h
JSUYA: removed

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: removed

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: done

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: done

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: done

packages/tizen_audio_manager/tizen/src/tizen_audio_manager_plugin.cc
JSUYA: done

packages/tizen_audio_manager/tizen/src/audio_manager.h
JSUYA: done

packages/tizen_audio_manager/tizen/src/audio_manager.h
JSUYA: > You can also remove the extra registered_volume_changed_cb_ flag by moving the implementation of the methods to the constructor and destructor of the class if you want (please refer to similar patterns used by other plugins such as [tizen_app_manager.h](https://github.com/flutter-tizen/plugins/blob/master/packages/tizen_app_manager/tizen/src/tizen_app_manager.h)).

I removed registered_volume_changed_cb_ and changed method name(Register/Unregister).

JSUYA: I didnt move Register/Unregister method to the constructor/destructor.
Because, in the `TizenAppManager` structure, when the tizen api inside the constructor fails, it seems that the `ret`'s error is not handled. I thought it would be better to keep this structure rather than adding separate code for error handle.
(And we should check the error handling of `app_manager_set_app_context_event_cb` of `TizenAppManager`.)

swift-kim: Note that in that case `TizenAudioManager` doesn't have to be a singleton class (`TizenAudioManager::GetInstance`).

JSUYA: I removed singleton code

packages/tizen_audio_manager/tizen/src/audio_manager.h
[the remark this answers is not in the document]
JSUYA: oh thank you. I fixed code

PR #518, opened 2023-01-25
Resolve Flutter 3.7 analyze issues

swift-kim:
Resolve lint warnings and formatting error.

@upple A change has been made to `message_client.dart` which is an auto-generated file. You may need to update the code generator (TIDLC) implementation.

### DIFF
--- a/packages/tizen_rpc_port/example/client/lib/message_client.dart
+++ b/packages/tizen_rpc_port/example/client/lib/message_client.dart
@@ -99,17 +99,13 @@ class Message extends ProxyBase {
   }
 
   Parcel _consumeCommand(Port port) {
-    try {
-      final Parcel parcel = Parcel.fromPort(port);
-      final int cmd = parcel.readInt32();
-      if (cmd != _MethodId.result.id) {
-        print('Received parcel is invalid. $cmd');
-      }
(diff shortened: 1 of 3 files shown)


packages/tizen_rpc_port/example/client/lib/message_client.dart
```diff
@@ -107,7 +107,7 @@ class Message extends ProxyBase {
       }
       return parcel;
     } catch (error) {
-      print(error.toString());
+      print(error);
```
swift-kim, line 110: Can we remove the catch clause so that the parcel parsing error can be propagated to the user code (as an unhandled exception) properly?

upple: Ok, I'll prepare modified patch. :)

upple (approved):

packages/tizen_rpc_port/example/client/lib/message_client.dart
upple: https://review.tizen.org/gerrit/c/platform/core/appfw/tidl/+/287255

bbrto21 (approved):

PR #520, opened 2023-01-30
[tizen_package_manager] Add threading to prevent app freezing

swift-kim:
Calling `PackageManager.getPackagesInfo` makes the app become unresponsive (for up to 10 seconds) on wearable devices.

### DIFF
--- a/packages/tizen_package_manager/tizen/src/tizen_package_manager_plugin.cc
+++ b/packages/tizen_package_manager/tizen/src/tizen_package_manager_plugin.cc
@@ -4,6 +4,7 @@
 
 #include "tizen_package_manager_plugin.h"
 
+#include <Ecore.h>
 #include <flutter/encodable_value.h>
 #include <flutter/event_channel.h>
 #include <flutter/event_sink.h>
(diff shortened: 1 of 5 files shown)


packages/tizen_package_manager/tizen/src/tizen_package_manager_plugin.cc
```diff
@@ -218,20 +219,31 @@ class TizenPackageManagerPlugin : public flutter::Plugin {
...
+    // might cause unresponsiveness on low-end devices if run on the platform
+    // thread.
+    ecore_thread_run(
+        [](void *data, Ecore_Thread *thread) {
+          auto *result = static_cast<FlMethodResult *>(data);
+
+          TizenPackageManager &package_manager =
+              TizenPackageManager::GetInstance();
+          std::optional<std::vector<PackageInfo>> packages =
+              package_manager.GetAllPackagesInfo();
```
swift-kim, line 232: Would it be safe if the caller thread (main thread) exits before the operation finishes?

I don't think we need a lock here because the operation is read-only.

packages/tizen_package_manager/tizen/src/tizen_package_manager_plugin.cc
```diff
@@ -218,20 +219,31 @@ class TizenPackageManagerPlugin : public flutter::Plugin {
...
+          if (packages.has_value()) {
+            flutter::EncodableList list;
+            for (const PackageInfo &package : packages.value()) {
+              list.push_back(
+                  flutter::EncodableValue(PackageInfoToMap(package)));
+            }
+            result->Success(flutter::EncodableValue(list));
```
bbrto21, line 239: Is it safe calling `Success`(or `Error`) in worker thread?  I think it's safe to return to the main thread and reply back.
(main thread.  Why don't you consider using the second and third arguments of `ecore_thread_run`?

bbrto21: If there's no problem with this patch , I'm also fine :)
However, I think it is necessary to add handling in case the start of the worker thread fails or is cancelled.
so  I think you should delete pointer of `result`(passed as user data) in `func_cancel`.

swift-kim: The thread is never canceled, so the only case we need to consider is the thread failing to start (according to the documentation). I'll update the implementation. Thanks!

bbrto21 (approved):
LGTM

PR #521, opened 2023-01-30
[tizen_app_manager] Add threading to prevent app freezing

swift-kim:
Similar to https://github.com/flutter-tizen/plugins/pull/520.


bbrto21 (approved):
LGTM

PR #522, opened 2023-02-02
[webview_flutter_tizen] Update webview_flutter to 4.0.2(legacy)

JSUYA:
Update the versions of webview_flutter and webview_flutter_platform_interface.
and use legacy APIs, temporarily.


JSUYA:
The existing example is designed for earlier versions.(webview_flutter < v4).
Therefore, change to use legacy widget in example and change webview_flutter_tizen to use webview_flutter's legacy API.
(Of course, we can modify and apply this patch after preparing all the v4 versions of webview_flutter.)

PR #526, opened 2023-02-10
[messageport] Refactor the C++ code

bbrto21:
* Contribute #354

This PR includes:
* Split MessagePort into LocalPort and RemotePort.
* Apply typedef for readablity.
* Apply singleton pattern to MessagePortManager.
* Introduce ErrorOr<> and LocalPortStreamHandler.
* Simplify plugin implementation.
* Cleanup debug logs.


bbrto21:
I tried to reproduce the problem that occurs in CI through the following commands in my local environment, but I failed to reproduce it :( 
```
./tools/tools_runner.sh integration-test --recipe ./.github/recipe.yaml --generate-emulator // recipe has only message port
```
```
cd packages/messageport
flutter-tizen test integration_test
```

Looking at the logs shown in CI, it seems that all the tests have passed.
But at the end of the test process, something seems to be going wrong after deleting the TV emulator...

and even though CI was executed using [meaningless commit](https://github.com/flutter-tizen/plugins/pull/527), [it failed the same way](https://github.com/flutter-tizen/plugins/actions/runs/4143369017).

Does anyone know about this?

swift-kim:
@bbrto21 Reproducible by providing the `GITHUB_ACTIONS` environment variable as `true`.

https://github.com/dart-lang/test/tree/master/pkgs/test#selecting-a-test-reporter

I'll make a patch in the tools soon.

packages/messageport/tizen/src/messageport_manager.cc
```diff
@@ -0,0 +1,244 @@
+// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "messageport_manager.h"
```
swift-kim, line 5: You could have just renamed the class to `MessagePort` without renaming the file, although it doesn't really matter.

packages/messageport/tizen/src/messageport_manager.h
```diff
@@ -0,0 +1,117 @@
...
+                                bool trusted_remote_port, bundle* message,
+                                void* user_data);
+
+  ErrorOr<bundle*> PrepareBundle(flutter::EncodableValue& message);
+
+  ErrorOr<int> GetRegisteredLocalPort(const std::string& port_name,
+                                      bool is_trusted);
+
+  // map<port, event_sink>
+  std::map<int, std::unique_ptr<FlEventSink>> sinks_;
```
swift-kim, line 111: No Flutter-specific types should appear within the platform API wrapper, according to our plugin refactoring guide.

packages/messageport/tizen/src/messageport_manager.h
```diff
@@ -0,0 +1,117 @@
...
+
+  // map<pair<port_name, trused>, port>
+  std::map<std::pair<std::string, bool>, int> local_ports_;
```
swift-kim, line 114: This data structure is not very readable and there may be a better way to do a similar thing (e.g. using two maps). Can both trusted and untrusted ports can be registered with the same name?

packages/messageport/tizen/src/messageport_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+
+#include <cstdint>
+#include <vector>
+
+#include "log.h"
+
+MessagePortManager::MessagePortManager() {}
+
+MessagePortManager::~MessagePortManager() {
+  for (auto& pair : local_ports_) {
```
swift-kim, line 18: May be simplified to
```suggestion
  for (const auto& [???, port] : local_ports_) {
```

packages/messageport/tizen/src/messageport_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+    int port = pair.second;
+    bool is_trusted = pair.first.second;
+    int ret = MESSAGE_PORT_ERROR_NONE;
+    if (is_trusted) {
+      ret = message_port_unregister_trusted_local_port(port);
+    } else {
+      ret = message_port_unregister_local_port(port);
+    }
+
+    if (MESSAGE_PORT_ERROR_NONE != ret) {
```
swift-kim, line 28: The style is not consistent within the file.
```suggestion
    if (ret != MESSAGE_PORT_ERROR_NONE) {
```

packages/messageport/tizen/src/messageport_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+  }
+}
+
+static bool ConvertEncodableValueToBundle(flutter::EncodableValue& value,
+                                          bundle* bundle) {
+  if (!bundle) {
+    return false;
+  }
+  std::unique_ptr<std::vector<uint8_t>> encoded =
+      flutter::StandardMessageCodec::GetInstance().EncodeMessage(value);
```
swift-kim, line 40: If a non-Flutter app is to communicate with a Flutter app via the message port, should it also use the same codec (`StandardMessageCodec`)?

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -206,36 +198,27 @@ class MessageportTizenPlugin : public flutter::Plugin {
...
+    result->Success();
   }
 
   // < channel_name, is_trusted > -> native number >
-  std::map<std::pair<std::string, bool>, int> native_ports_;
-  std::set<std::unique_ptr<flutter::EventChannel<flutter::EncodableValue>>>
-      event_channels_;
-  MessagePortManager manager_;
+  std::set<std::unique_ptr<FlEventChannel>> event_channels_;
   flutter::PluginRegistrar *plugin_registrar_;
```
swift-kim, line 222: `pluginRegistrar`→`plugin_registrar` at line 82.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -206,36 +198,27 @@ class MessageportTizenPlugin : public flutter::Plugin {
```
swift-kim, line 198: The first argument of `result->Error` should be an error code, something like "Invalid arguments".

packages/messageport/tizen/src/messageport_manager.cc
bbrto21: I think so.since this is a kind of serialization, it necessarily needs a deserialization algorithm.

packages/messageport/tizen/src/messageport_manager.h
bbrto21: Yes they can be registered under the same name and can be used together.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -128,70 +161,29 @@ class MessageportTizenPlugin : public flutter::Plugin {
...
-              return nullptr;
-            });
-
-    event_channel->SetStreamHandler(std::move(event_channel_handler));
+    auto event_channel = std::make_unique<FlEventChannel>(
+        plugin_registrar_->messenger(), event_channel_name.str(),
+        &flutter::StandardMethodCodec::GetInstance());
+    event_channel->SetStreamHandler(
+        std::make_unique<LocalPortStreamHandler>(port_name, trusted));
     event_channels_.insert(std::move(event_channel));
```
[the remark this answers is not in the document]
bbrto21: Did you mean this plugin need Finalizer to close channel in the Dart side? I'm not the original author, so I don't know original design principle.

bbrto21:
@swift-kim I updated PR based on review. please review again.

packages/messageport/tizen/src/messageport_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+                                           bundle* message, void* user_data) {
+  MessagePortManager* manager = static_cast<MessagePortManager*>(user_data);
+  if (manager->sinks_.find(local_port_id) != manager->sinks_.end()) {
+    uint8_t* byte_array = nullptr;
+    size_t size = 0;
+    int ret = bundle_get_byte(message, "bytes",
+                              reinterpret_cast<void**>(&byte_array), &size);
+    if (ret != BUNDLE_ERROR_NONE) {
+      manager->sinks_[local_port_id]->Error("Failed to parse a response");
+    }
```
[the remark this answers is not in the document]
bbrto21: Thank you, good point!

packages/messageport/tizen/src/messageport_tizen_plugin.cc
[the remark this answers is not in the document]
bbrto21: Okay. As the title of this PR, I didn't really think carefully about the Dart side.

packages/messageport/tizen/src/messageport.cc
```diff
@@ -185,21 +171,31 @@ MessagePortResult MessagePortManager::Send(std::string& remote_app_id,
...
-  bundle* bundle = nullptr;
-  MessagePortResult result = PrepareBundle(message, bundle);
-  if (!result) {
-    return result;
-  }
-  int ret;
+std::optional<MessagePortError> MessagePort::Send(
+    std::string& remote_app_id, std::string& port_name,
+    std::unique_ptr<std::vector<uint8_t>> encoded_message, bool is_trusted,
+    const std::string& local_port_name, bool local_port_is_trusted) {
```
swift-kim, line 184: You may specify `remote` and `local` appropriately in the argument names (e.g. `remote_port_name`, `trusted_remote_port`) to avoid ambiguity.

As there are too many arguments to this method, a better way to refactor the code could be splitting the whole class into two (`LocalPort` and `RemotePort`), and let each class has `name` and `is_trusted` as members (as well as other relevant methods). In this way the code will become more object oriented and the singleton pattern will no longer be necessary.

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,126 @@
...
 
-struct MessagePortResult {
-  MessagePortResult() : error_code(MESSAGE_PORT_ERROR_NONE){};
-  MessagePortResult(int code) : error_code(code) {}
+ private:
+  ErrorOr() = default;
 
-  // Returns false on error
-  operator bool() const { return MESSAGE_PORT_ERROR_NONE == error_code; }
+  T TakeValue() && { return std::get<T>(std::move(vlaue_or_error_)); }
```
swift-kim, line 60: Never used...

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,127 @@
...
+  ErrorOr() = default;
+
+  std::variant<T, MessagePortError> vlaue_or_error_;
+};
 
-  int error_code;
+struct Message {
+  const std::string remote_app_id;
+  const std::string remote_port;
+  bool trusted_remote_port = false;
```
swift-kim, line 66: You may use the already existing `RemotePort` class to represent this information.

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,127 @@
...
-  MessagePortResult RegisterLocalPort(const std::string& port_name,
-                                      EventSink sink, bool is_trusted,
-                                      int* local_port);
-  MessagePortResult UnregisterLocalPort(int local_port_id);
-  MessagePortResult Send(std::string& remote_app_id, std::string& port_name,
-                         flutter::EncodableValue& message, bool is_trusted);
-  MessagePortResult Send(std::string& remote_app_id, std::string& port_name,
-                         flutter::EncodableValue& message, bool is_trusted,
-                         int local_port);
+  LocalPort(const std::string name, const bool is_trusted)
```
swift-kim, line 75:
```suggestion
  LocalPort(const std::string& name, bool is_trusted)
```
The second `const` doesn't seem useful.

The same applies to `RemotePort::RemotePort`.

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,127 @@
...
-  MessagePortResult CreateResult(int return_code);
-  MessagePortResult PrepareBundle(flutter::EncodableValue& message, bundle*& b);
+class RemotePort {
+ public:
+  RemotePort(const std::string app_id, const std::string name,
+             const bool is_trusted)
+      : app_id_(app_id), name_(name), is_trusted_(is_trusted) {}
+
+  ErrorOr<bool> CheckRemotePort();
+  std::optional<MessagePortError> Send(
```
swift-kim, line 107: You may add a blank line between the lines.

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,127 @@
...
+  bool is_trusted() { return is_trusted_; }
+
+  int port() { return port_; }
 
  private:
   static void OnMessageReceived(int local_port_id, const char* remote_app_id,
                                 const char* remote_port,
                                 bool trusted_remote_port, bundle* message,
                                 void* user_data);
+  const std::string name_;
```
swift-kim, line 94: nit: A blank line may be added before the data member declarations.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -9,25 +9,117 @@
...
+  if (iter != map->end() && !iter->second.IsNull()) {
+    if (auto *value = std::get_if<T>(&iter->second)) {
+      out = *value;
+      return true;
+    }
+  }
+  return false;
+}
+
+class LocalPortStreamHandlerHandlerError : public FlStreamHandlerError {
```
swift-kim, line 46: HandlerHandler

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -9,25 +9,117 @@
...
+                  *message.encoded_message.get());
+          map[flutter::EncodableValue("message")] = *(value.get());
+          if (message.remote_app_id.size()) {
+            map[flutter::EncodableValue("remoteAppId")] =
+                flutter::EncodableValue(message.remote_app_id);
+          }
+          if (message.remote_port.size()) {
+            map[flutter::EncodableValue("remotePort")] =
+                flutter::EncodableValue(message.remote_port);
+          }
```
swift-kim, line 87: `remoteAppId` and `remotePort` can't be optional. Non-null string values are expected by the Dart side.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -9,25 +9,117 @@
...
+    }
+
+    return nullptr;
+  }
+
+  std::unique_ptr<FlStreamHandlerError> OnCancelInternal(
+      const flutter::EncodableValue *arguments) override {
+    std::optional<MessagePortError> error = local_port_->Unregister();
+    if (error.has_value()) {
+      LOG_ERROR("Error OnCancel: %s", error.value().message().c_str());
```
swift-kim, line 107: You can return a `LocalPortStreamHandlerError`.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -9,25 +9,117 @@
...
+          if (message.error.size()) {
+            event_sink_->Error(message.error);
+            return;
+          }
+
+          flutter::EncodableMap map;
+          auto value =
+              flutter::StandardMessageCodec::GetInstance().DecodeMessage(
+                  *message.encoded_message.get());
+          map[flutter::EncodableValue("message")] = *(value.get());
```
swift-kim, line 79: How about
```suggestion
          std::unique_ptr<flutter::EncodableValue> decoded =
              flutter::StandardMessageCodec::GetInstance().DecodeMessage(
                  *message.encoded_message.get());
          map[flutter::EncodableValue("message")] = *decoded;
```

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -9,25 +9,117 @@
...
+          }
+          map[flutter::EncodableValue("trusted")] =
+              flutter::EncodableValue(message.trusted_remote_port);
+
+          event_sink_->Success(flutter::EncodableValue(map));
+        });
+
+    if (error.has_value()) {
+      return std::make_unique<LocalPortStreamHandlerHandlerError>(
+          std::to_string(error.value().code()), error.value().message().c_str(),
```
swift-kim, line 96:
```suggestion
          std::to_string(error.value().code()), error.value().message(),
```

packages/messageport/tizen/src/messageport.cc
```diff
@@ -5,234 +5,161 @@
...
-  bool result = ConvertEncodableValueToBundle(message, bundle);
-  if (!result) {
-    LOG_ERROR("Failed to parse EncodableValue");
+  int ret = bundle_add_byte(bundle, "bytes", encoded_message->data(),
+                            encoded_message->size());
+
+  if (ret != BUNDLE_ERROR_NONE) {
     bundle_free(bundle);
-    return CreateResult(MESSAGE_PORT_ERROR_INVALID_PARAMETER);
+    return MessagePortError(MESSAGE_PORT_ERROR_INVALID_PARAMETER);
```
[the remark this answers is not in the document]
bbrto21: I guess that original author want to represent a bundle error as a message port  error. 
I refrained from changing the results of the function, but I think it doesn't matter. Is it better to change?

packages/messageport/tizen/src/messageport.cc
```diff
@@ -5,234 +5,161 @@
...
-                                           bool is_trusted, int local_port) {
-  LOG_DEBUG("Send (%s, %s), port: %d, trusted: %s", remote_app_id.c_str(),
-            port_name.c_str(), local_port, is_trusted ? "yes" : "no");
-  bundle* bundle = nullptr;
-  MessagePortResult result = PrepareBundle(message, bundle);
-  if (!result) {
-    return result;
+std::optional<MessagePortError> RemotePort::Send(
+    std::unique_ptr<std::vector<uint8_t>> encoded_message,
+    LocalPort* local_port) {
```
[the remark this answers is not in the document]
bbrto21: > Also encoded_message shouldn't necessarily be a unique_ptr. It could be just const std::vector<uint8_t> & if you want to pass data without copying.

@swift-kim 
I think it is natural because the result of decoding with `StandardMessageCodec` is `unique_ptr`. or using raw pointer from `unique_ptr` is better for this case. also, it need to remove the `const` again to change `vector` to a `bundle`. how about this?
 
``` c++ 
std::unique_ptr<std::vector<uint8_t>> encoded_message =
        flutter::StandardMessageCodec::GetInstance().EncodeMessage(message);
// ...
remote_port.SendWithLocalPort(encoded_message.get()) // remote_port.SendWithLocalPort(*(encoded_message.get()))

```

packages/messageport/tizen/src/messageport.cc
```diff
@@ -5,234 +5,161 @@
...
+  int ret = bundle_get_byte(bundle, "bytes",
+                            reinterpret_cast<void**>(&byte_array), &size);
+  if (ret != BUNDLE_ERROR_NONE) {
+    Message message{error : "Failed to get data from bundle"};
+    self->message_callback_(message);
+  } else {
+    std::unique_ptr<std::vector<uint8_t>> encoded_message =
+        std::make_unique<std::vector<uint8_t>>(byte_array, byte_array + size);
+    Message message{remote_app_id ? remote_app_id : "",
+                    remote_port ? remote_port : "", trusted_remote_port,
```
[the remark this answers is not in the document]
bbrto21: Message has const members, they need to be initialized at creation.

packages/messageport/tizen/src/messageport.cc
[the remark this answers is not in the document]
swift-kim: This totally makes sense:
```patch
--- a/packages/messageport/tizen/src/messageport.cc
+++ b/packages/messageport/tizen/src/messageport.cc
@@ -121,9 +121,8 @@ std::optional<MessagePortError> RemotePort::Send(
 }

 std::optional<MessagePortError> RemotePort::Send(
-    std::unique_ptr<std::vector<uint8_t>> encoded_message,
-    LocalPort* local_port) {
-  ErrorOr<bundle*> maybe_bundle = PrepareBundle(encoded_message.get());
+    const std::vector<uint8_t>& encoded_message, LocalPort* local_port) {
+  ErrorOr<bundle*> maybe_bundle = PrepareBundle(encoded_message);
   if (maybe_bundle.has_error()) {
     return maybe_bundle.error();
   }
@@ -147,14 +146,14 @@ std::optional<MessagePortError> RemotePort::Send(
 }

 ErrorOr<bundle*> RemotePort::PrepareBundle(
-    std::vector<uint8_t>* encoded_message) {
+    const std::vector<uint8_t>& encoded_message) {
   bundle* bundle = bundle_create();
   if (!bundle) {
     return MessagePortError(MESSAGE_PORT_ERROR_OUT_OF_MEMORY);
   }

-  int ret = bundle_add_byte(bundle, "bytes", encoded_message->data(),
-                            encoded_message->size());
+  int ret = bundle_add_byte(bundle, "bytes", encoded_message.data(),
+                            encoded_message.size());

   if (ret != BUNDLE_ERROR_NONE) {
     bundle_free(bundle);

--- a/packages/messageport/tizen/src/messageport_tizen_plugin.cc
+++ b/packages/messageport/tizen/src/messageport_tizen_plugin.cc
@@ -249,9 +249,9 @@ class MessageportTizenPlugin : public flutter::Plugin {
         result->Error("Invalid arguments");
         return;
       }
-      error = remote_port.Send(std::move(encoded_message), local_port);
+      error = remote_port.Send(*encoded_message, local_port);
     } else {
-      error = remote_port.Send(std::move(encoded_message));
+      error = remote_port.Send(*encoded_message);
     }

     if (error.has_value()) {
```

The methods are synchronous and read-only (they do not modify the data or store it elsewhere), so transferring ownership is not very necessary in this case.

packages/messageport/tizen/src/messageport.cc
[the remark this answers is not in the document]
swift-kim: Hmm, but obviously `MESSAGE_PORT_ERROR_INVALID_PARAMETER` doesn't make sense. It would be better to change as suggested.

packages/messageport/tizen/src/messageport.cc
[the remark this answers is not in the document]
swift-kim: I don't think `const` is quite necessary for passive data objects but if you want to keep `const`, you may add some ` = ` and trailing comma for general consistency and readability:

```cpp
Message message = {
    remote_app_id ? remote_app_id : "",
    remote_port ? remote_port : "",
    trusted_remote_port,
    std::move(encoded_message),
};
```
You may not need the trailing comma if you apply https://github.com/flutter-tizen/plugins/pull/526#discussion_r1141807096.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
bbrto21: No. They are expected to be optional.

https://github.com/flutter-tizen/plugins/blob/882cf4339cfab3d6788c69cf8186b6505573487c/packages/messageport/lib/messageport_tizen.dart#L53

swift-kim: You're right. We shouldn't necessarily mark `remote_app_id` and `remote_port` as `std::optional` and can just use `""` if we're sure that they cannot be empty strings.

bbrto21: If we use "' for remote port, the final result of the Dart API will be different. As a result, we cannot pass tests expecting remort port to be null.

swift-kim: Uhm, what I meant was, the current C++ implementation which uses `std::string("")` to represent null strings should be okay.

packages/messageport/tizen/src/messageport.h
```diff
@@ -2,59 +2,129 @@
...
-  MessagePortResult RegisterLocalPort(const std::string& port_name,
-                                      EventSink sink, bool is_trusted,
-                                      int* local_port);
-  MessagePortResult UnregisterLocalPort(int local_port_id);
-  MessagePortResult Send(std::string& remote_app_id, std::string& port_name,
-                         flutter::EncodableValue& message, bool is_trusted);
-  MessagePortResult Send(std::string& remote_app_id, std::string& port_name,
-                         flutter::EncodableValue& message, bool is_trusted,
-                         int local_port);
+  RemotePort() = default;
```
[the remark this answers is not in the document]
bbrto21: I applied optional to Message but I couldn't remove `remote_port` presence check from C++ code totally.
Because this change brings about a change in API behavior.

packages/messageport/tizen/src/messageport_tizen_plugin.cc
```diff
@@ -39,204 +132,152 @@ class MessageportTizenPlugin : public flutter::Plugin {
...
-
-  void HandleMethodCall(
-      const flutter::MethodCall<flutter::EncodableValue> &method_call,
-      std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
+  void HandleMethodCall(const FlMethodCall &method_call,
+                        std::unique_ptr<FlMethodResult> result) {
     const auto &method_name = method_call.method_name();
-    const flutter::EncodableValue *args = method_call.arguments();
+    const auto *arguments =
+        std::get_if<flutter::EncodableMap>(method_call.arguments());
```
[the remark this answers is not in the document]
bbrto21: Thank you. I added a commit to check null

PR #541, opened 2023-03-09
Improve analysis_options alignment with flutter/packages

swift-kim:
There have been a number of non-trivial updates to the `analysis_options.yaml` file in the flutter/plugins (now it's flutter/packages) repo. I'm going to align our `analysis_options.yaml` with theirs gradually. This PR is part of the work.

Related commits:
- https://github.com/flutter/plugins/commit/9990a4b1d72158a14924d712e89d4efef3f178d2
- https://github.com/flutter/plugins/commit/3a093e49bbe956350106d869a5d6b346bed612ce

### DIFF
--- a/packages/google_maps_flutter/lib/src/convert.dart
+++ b/packages/google_maps_flutter/lib/src/convert.dart
@@ -171,7 +171,7 @@ LatLngBounds _convertToBounds(String value) {
           northeast: LatLng(bound['north'] as double, bound['east'] as double));
     }
   } catch (e) {
-    print('Javascript Error: $e');
+    debugPrint('Javascript Error: $e');
   }
   return util.nullLatLngBounds;
 }
@@ -184,7 +184,7 @@ LatLng _convertToLatLng(String value) {
       return LatLng((latlng['lat'] as num) + 0.0, (latlng['lng'] as num) + 0.0);
     }
   } catch (e) {
-    print('Javascript Error: $e');
+    debugPrint('Javascript Error: $e');
   }
   return util.nullLatLng;
 }
@@ -205,7 +205,7 @@ ScreenCoordinate _convertToPoint(String value) {
       return ScreenCoordinate(x: x, y: y);
     }
   } catch (e) {
-    print('Javascript Error: $e');
+    debugPrint('Javascript Error: $e');
   }
   return util.nullScreenCoordinate;
 }
--- a/packages/google_maps_flutter/lib/src/convert.dart
+++ b/packages/google_maps_flutter/lib/src/convert.dart
@@ -324,12 +324,10 @@ util.GPolygonOptions _polygonOptionsFromPolygon(Polygon polygon) {
     List<LatLng> holePath = polygon.holes[i];
     if (_isPolygonClockwise(holePath) == polygonDirection) {
       holePath = holePath.reversed.toList();
-      if (kDebugMode) {
-        print(
-            'Hole [$holeIndex] in Polygon [${polygon.polygonId.value}] has been reversed.'
-            ' Ensure holes in polygons are "wound in the opposite direction to the outer path."'
-            ' More info: https://github.com/flutter/flutter/issues/74096');
-      }
+      debugPrint(
+          'Hole [$holeIndex] in Polygon [${polygon.polygonId.value}] has been reversed.'
+          ' Ensure holes in polygons are "wound in the opposite direction to the outer path."'
+          ' More info: https://github.com/flutter/flutter/issues/74096');
     }
     paths.add(holePath);
     holeIndex++;
--- a/packages/tizen_rpc_port/example/client/lib/message_client.dart
+++ b/packages/tizen_rpc_port/example/client/lib/message_client.dart
@@ -102,9 +102,8 @@ class Message extends ProxyBase {
     final Parcel parcel = Parcel.fromPort(port);
     final int cmd = parcel.readInt32();
     if (cmd != _MethodId.result.id) {
-      print('Received parcel is invalid. $cmd');
+      throw Exception('The received parcel is invalid ($cmd).');
     }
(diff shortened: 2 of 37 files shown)


packages/google_maps_flutter/lib/src/convert.dart
```diff
@@ -171,7 +171,7 @@ LatLngBounds _convertToBounds(String value) {
           northeast: LatLng(bound['north'] as double, bound['east'] as double));
     }
   } catch (e) {
-    print('Javascript Error: $e');
+    debugPrint('Javascript Error: $e');
```
bbrto21, line 174: Why don't you fix `kDebugMode` and `print` in line 327?

bbrto21 (approved):

packages/google_maps_flutter/lib/src/convert.dart
swift-kim: Done?

I'm not sure if it's okay to change the code since I don't know where the code is originally from.

bbrto21: https://github.com/flutter/packages/commit/ab463da34bcf8ec80ba1992a8f6e7819f40a360d#diff-ad093591ababe86bff6ff2af9ce718a9b0ef32f011611922b7d245420c71cb27

It looks like a very old remnant...

packages/tizen_rpc_port/example/client/lib/message_client.dart
```diff
@@ -102,9 +102,8 @@ class Message extends ProxyBase {
     final Parcel parcel = Parcel.fromPort(port);
     final int cmd = parcel.readInt32();
     if (cmd != _MethodId.result.id) {
-      print('Received parcel is invalid. $cmd');
+      throw Exception('The received parcel is invalid ($cmd).');
     }
-
     return parcel;
```
swift-kim, line 107: @upple A code change has been made to auto-generated file. Please review and update the TIDL compiler.

packages/tizen_rpc_port/example/client/lib/message_client.dart
```diff
@@ -102,9 +102,8 @@ class Message extends ProxyBase {
     final Parcel parcel = Parcel.fromPort(port);
     final int cmd = parcel.readInt32();
     if (cmd != _MethodId.result.id) {
-      print('Received parcel is invalid. $cmd');
+      throw Exception('The received parcel is invalid ($cmd).');
     }
-
```
swift-kim, line 107: @upple A code change has been made to auto-generated file. Please review and update the TIDL compiler.

upple: https://review.tizen.org/gerrit/c/platform/core/appfw/tidl/+/289709
I'm preparing the patch.

swift-kim: Flutter exceptions are not like C# exceptions. When unhandled, they don't crash the process but are just logged to the console.

upple: Thank you.
As you said, it doesn't cause crash.

upple: @swift-kim 
Hello. I have a question.
When applying this patch, if a throw occurs in the `_coinsumeCommand()`, how does the subsequent operation proceed? We are proceeding with the while loop to receive invalid parcels. We are concerned that if a throw occurs, the subsequent operation will be damaged.

swift-kim: Unhandled exceptions are propagated to callers, aborting execution of the rest of the code.

upple: This will cause a series of errors in the data you receive afterwards.
Through while loop, we are organizing the wrong parcel that occurred before.
This was our intention.

swift-kim: It's an exceptional situation (invalid parcel), so an exception is thrown. I don't see any problem with this. Are you saying the invalid parcel is part of the normal execution flow? Then why did you add a `print` there and not just ignore the invalid (actually normal) parcel? Also how do you think of `Parcel.fromPort` and `parcel.readInt32`? They can both throw exceptions as well.

upple: A typical case in which the parcel is not valid is when the sender transmits late and the previous request is timeout.
In this case, an exception occurs in the fromPort and does not work afterwards, so when the parcel sent by the sender arrives later, it is received when the next request is processed. However, this should be ignored because it is a parcel for the previous request.
Printing log was simply put for debugging. If you think it's unnecessary, you can remove it.

swift-kim: Okay, I'll remove the logic in the currently open PR (#543).

upple: Thank you.
Then, we will prepare above gerrit patch without these changes.

swift-kim: No, you can't just remove the if statement from `_consumeCommand`. It will leave the `cmd` variable unused and the linter will complain again. You should check all the changes regarding the `tizen_rpc_port` package in my new PR and the TIDLC output should be exactly the same as the changed code. Note that EVERY change in the PR has a reason. If you don't understand why the change was made or if you disagree with the change, please leave a comment there.

upple: Ok. I'll apply these changes. 
Thanks for the preparing this patch!

swift-kim: Thank you!

PR #543, opened 2023-03-13
More analysis_options alignment

swift-kim:
This is a continuation of https://github.com/flutter-tizen/plugins/pull/541. The `analysis_options.yaml` in this repo will exactly match the `analysis_options.yaml` in [flutter/packages](https://github.com/flutter/packages) after this change.

Related commits:
- https://github.com/flutter/plugins/commit/9fdc899b72cac49a4ffade799563c4730c0279a8
- https://github.com/flutter/plugins/commit/783ce6d8f05ecb4da19b90bea063fff5278b551f
- https://github.com/flutter/plugins/commit/1e5efd144f93f3f5b5f804272c126af3185fd2f6

### DIFF
--- a/packages/tizen_rpc_port/example/client/lib/message_client.dart
+++ b/packages/tizen_rpc_port/example/client/lib/message_client.dart
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// ignore_for_file: public_member_api_docs
+// ignore_for_file: public_member_api_docs, omit_local_variable_types
 
 import 'package:tizen_rpc_port/tizen_rpc_port.dart';
 
@@ -104,6 +104,7 @@ class Message extends ProxyBase {
     if (cmd != _MethodId.result.id) {
       throw Exception('The received parcel is invalid ($cmd).');
     }
+
     return parcel;
   }
 
--- a/packages/tizen_rpc_port/example/client/lib/message_client.dart
+++ b/packages/tizen_rpc_port/example/client/lib/message_client.dart
@@ -127,6 +127,10 @@ class Message extends ProxyBase {
     late Parcel parcelReceived;
     while (true) {
       parcelReceived = Parcel.fromPort(port);
+      final int cmd = parcelReceived.readInt32();
+      if (cmd != _MethodId.result.id) {
+        continue;
+      }
       final ParcelHeader headerReceived = parcelReceived.header;
       if (headerReceived.tag.isEmpty) {
         break;
@@ -172,6 +176,10 @@ class Message extends ProxyBase {
     late Parcel parcelReceived;
     while (true) {
       parcelReceived = Parcel.fromPort(port);
+      final int cmd = parcelReceived.readInt32();
+      if (cmd != _MethodId.result.id) {
+        continue;
+      }
       final ParcelHeader headerReceived = parcelReceived.header;
       if (headerReceived.tag.isEmpty) {
         break;
--- a/packages/tizen_rpc_port/example/client/lib/message_client.dart
+++ b/packages/tizen_rpc_port/example/client/lib/message_client.dart
@@ -34,7 +34,7 @@ abstract class _Delegate extends Parcelable {
   int id = 0;
   bool once = false;
   int sequenceId = 0;
-  Function? callback;
+  Function callback;
   static int sequenceNum = 0;
 
   Future<void> onReceivedEvent(Parcel parcel);
@@ -65,7 +65,7 @@ class _NotifyCallback extends _Delegate {
     final String sender = parcel.readString();
     final String message = parcel.readString();
 
-    callback?.call(sender, message);
+    (callback as NotifyCallback)(sender, message);
   }
 }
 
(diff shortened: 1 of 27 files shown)


packages/tizen_rpc_port/example/client/lib/message_client.dart
```diff
@@ -65,7 +65,7 @@ class _NotifyCallback extends _Delegate {
     final String sender = parcel.readString();
     final String message = parcel.readString();
 
-    callback?.call(sender, message);
+    (callback as NotifyCallback)(sender, message);
```
swift-kim, line 68: @upple Modifications made to auto-generated `message_client.dart` and `message_server.dart`. Please review all and update the TIDL compiler accordingly.

upple: https://review.tizen.org/gerrit/c/platform/core/appfw/tidl/+/289709
I'm preparing the patch.

swift-kim: Check **ALL** the changes but not just this line.

https://github.com/flutter-tizen/plugins/pull/543/files

packages/tizen_rpc_port/example/client/lib/message_client.dart
```diff
@@ -135,7 +126,11 @@ class Message extends ProxyBase {
 
     late Parcel parcelReceived;
     while (true) {
-      parcelReceived = _consumeCommand(port);
+      parcelReceived = Parcel.fromPort(port);
+      final int cmd = parcel.readInt32();
+      if (cmd != _MethodId.result.id) {
+        continue;
+      }
```
swift-kim, line 133: This part has been added to suppress an `unused_field` warning regarding `_MethodId.result`.

upple: I forgot this.
I'll add it. Thanks :)

upple (approved):
tizen_rpc_port reviewed. 👍

packages/webview_flutter/lib/src/tizen_webview.dart
```diff
@@ -12,10 +12,6 @@ import 'package:webview_flutter_platform_interface/webview_flutter_platform_inte
 /// The channel name of [TizenWebView].
 const String kTizenWebViewChannelName = 'plugins.flutter.io/tizen_webview_';
 
-/// Signature for callbacks that report a method calls by navigation delegate method channel.
-typedef NavigationDelegateMethodChannelCallback = Future<dynamic> Function(
-    MethodCall method);
```
swift-kim, line 17: @JSUYA Removed this since it seemed to be unused.

packages/webview_flutter/lib/src/tizen_webview.dart
```diff
@@ -139,28 +137,29 @@ class TizenWebView {
...
         'y': y,
       });
 
-  /// Return the set scroll position of this view.
+  /// Returns the scroll position of this view set by [scrollTo].
   Future<Offset> getScrollPosition() async {
-    final dynamic position =
-        await _invokeChannelMethod<dynamic>('getScrollPosition');
+    final Map<Object?, Object?>? position =
+        await _invokeChannelMethod<Map<Object?, Object?>?>('getScrollPosition');
```
swift-kim, line 157: A better practice is to use `MethodChannel.invokeMapMethod`, but for now we keep using `_invokeChannelMethod`.

packages/webview_flutter/lib/src/tizen_webview.dart
```diff
@@ -33,8 +29,10 @@ class TizenWebView {
   Future<bool?> _onMethodCall(MethodCall call) async {
     switch (call.method) {
       case 'javaScriptChannelMessage':
-        final String channel = call.arguments['channel']! as String;
-        final String message = call.arguments['message']! as String;
+        final Map<String, Object?> arguments =
+            (call.arguments as Map<Object?, Object?>).cast<String, Object?>();
+        final String channel = arguments['channel']! as String;
+        final String message = arguments['message']! as String;
```
JSUYA, line 35: I have a question.
I understand it uses `Map` because of `avoid_dynamic_calls`.
Is cast even necessary in this situations that the type is predictable as String?
I just wonder if it is possible to do like below.
```dart
final Map<String, String?> arguments =
    call.arguments as Map<String, String?>;
final String channel = arguments['channel']!;
final String message = arguments['message']!;
```

swift-kim: No, that's not possible. `as` cannot replace `Map.cast()` since they are different things. An object of type `dynamic` or `Object?` must first be converted to either a `Map<dynamic, dynamic>` or `Map<Object?, Object?>` (the latter is preferred) using `as`, and then casted to a specific type (`Map<K, V>`) using either `Map.cast()` or `Map.from()` (the former is preferred because it involves less copying).

swift-kim: FYI) The code can be simplified to

```dart
final arguments = (call.arguments as Map).cast<String, Object?>();
```

if you disable the `always_specify_types` lint rule. By default it's disabled in the core lints, but we use strict typing with custom analysis options in this repo.

JSUYA: Oh.. OK. Thank you for detail reply. It helped me understand :) .

JSUYA (approved):
I checked the changes related to the webview.

PR #554, opened 2023-03-29
[webview_flutter] Clean up integration test and unused code

JSUYA:
- Remove unnecessary integration test.
  - "loadRequest with headers"
    - Currently, requests with headers are not supported.
  - "Video playback policy", "Audio playback policy"
    - Doesn't work normally. (When there are multimedia elements, onPageFinished is not called from the ewk.)
  - "target _blank opens in same window"
    - Tizen ewk is not supported.
  - "clearLocalStorage"
    - TizenWebViewController has no implementation.
- Remove unused code in webview.cc

### DIFF
--- a/packages/webview_flutter/example/integration_test/webview_flutter_test.dart
+++ b/packages/webview_flutter/example/integration_test/webview_flutter_test.dart
@@ -9,13 +9,9 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
-// TODO(a14n): remove this import once Flutter 3.1 or later reaches stable (including flutter/flutter#104231)
-// ignore: unnecessary_import
-import 'dart:typed_data';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:webview_flutter/webview_flutter.dart';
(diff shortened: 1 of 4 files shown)


packages/webview_flutter/example/integration_test/webview_flutter_test.dart
```diff
@@ -668,118 +392,6 @@ Future<void> main() async {
...
-        ))
-        ..loadRequest(Uri.parse(primaryUrl));
-
-      await tester.pumpWidget(WebViewWidget(controller: controller));
-
-      expect(controller.currentUrl(), completion(primaryUrl));
-      await pageLoaded.future;
-      pageLoaded = Completer<void>();
-
-      await controller.runJavaScript('window.open("$secondaryUrl")');
```
swift-kim, line 708: Do you know how this works on other platforms?

swift-kim (approved):

packages/webview_flutter/example/integration_test/webview_flutter_test.dart
swift-kim: You mean `window.open` doesn't actually open a new window on Android?

JSUYA: Yes, I know that if I do not specify a target, it will be output to the same Window.
In ewk, when I tested it, Window.open () returned NULL.
I couldn't find to the exact specifications of the ewk,so I think it as an unsupported.

PR #557, opened 2023-04-05
[google_maps_flutter] Update webview_flutter_lwe to 0.2.0

Swanseo0:
Apply the changed interface from webview_flutter_lwe v0.2.0


bbrto21:
The integration test for this plugin is skipped in CI, so you must check manually.
I've run the test but I got failures, I think you should check test related things before review.

bbrto21:
@Swanseo0 Could you please add a result of internal test you manually ran? It would be nice if you could simply leave it as a comment.

PR #562, opened 2023-04-23
[in_app_purchase] Add in_app_purchase package

hyue7:
The Tizen implementation of [in_app_purchase](https://github.com/flutter/plugins/tree/main/packages/in_app_purchase) based on [Samsung Checkout API](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/samsung-checkout.html) .
This plugin allows users to request products and make purchases within your application. It is only supported on Tizen TV devices.
Before start this plugin, please read 'README' file carefully, especially the [Preparation](https://github.com/flutter-tizen/plugins/pull/562/files#diff-c2cde410186d092cb87a071ad3eb21f5087d98c4766cdc153353a6490b737422R25).


packages/in_app_purchase_tizen/README.md
```diff
@@ -0,0 +1,18 @@
...
+a specialized package that includes platform-specific implementation code for
+Android and/or iOS.
+
+For help getting started with Flutter development, view the
+[online documentation](https://flutter.dev/docs), which offers tutorials,
+samples, guidance on mobile development, and a full API reference.
+
+The plugin project was generated without specifying the `--platforms` flag, no platforms are currently supported.
+To add platforms, run `flutter create -t plugin --platforms <platforms> .` in this directory.
+You can also find a detailed instruction on how to add platforms in the `pubspec.yaml` at https://flutter.dev/docs/development/packages-and-plugins/developing-packages#plugin-platforms.
```
xiaowei-guan, line 18: Please update README

packages/in_app_purchase_tizen/tizen/src/in_app_purchase_tizen_plugin.cc
```diff
@@ -0,0 +1,57 @@
...
+ public:
+  static void RegisterWithRegistrar(
+      FlutterDesktopPluginRegistrarRef registrar_ref,
+      flutter::PluginRegistrar *plugin_registrar);
+  InAppPurchaseTizenPlugin(FlutterDesktopPluginRegistrarRef registrar_ref,
+                           flutter::PluginRegistrar *plugin_registrar);
+  virtual ~InAppPurchaseTizenPlugin() { DisposeAllPlayers(); }
+
+ private:
+  void DisposeAllPlayers();
```
xiaowei-guan, line 23: Please modify this method name.

packages/in_app_purchase_tizen/tizen/src/in_app_purchase_tizen_plugin.cc
```diff
@@ -0,0 +1,57 @@
...
+
+#include "billing_manager.h"
+#include "log.h"
+
+class InAppPurchaseTizenPlugin : public flutter::Plugin {
+ public:
+  static void RegisterWithRegistrar(
+      FlutterDesktopPluginRegistrarRef registrar_ref,
+      flutter::PluginRegistrar *plugin_registrar);
+  InAppPurchaseTizenPlugin(FlutterDesktopPluginRegistrarRef registrar_ref,
```
xiaowei-guan, line 18: we can remove registrar_ref, because just need plugin_registrar when create method channel.

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,264 @@
...
+    LOG_ERROR(" Fail to init billing api");
+  }
+}
+
+bool BillingManager::GetProductList(const flutter::EncodableValue *args) {
+  LOG_INFO("start get product list");
+  std::string app_id;
+  std::string country_code;
+  std::string item_type;
+  int page_size;
```
xiaowei-guan, line 45: Suggest change int to int_64

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,160 @@
...
+
+const String kIAPSource = 'samsung_checkout';
+
+late String appId;
+
+late String serverType;
+
+late Set<String> identifiers_;
+
+/// An [InAppPurchasePlatform] that wraps Android BillingClient.
```
xuelian-bai, line 23: Remove android related comments

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,264 @@
...
+  } else if (strcasecmp("DUMMY", server_type_string) == 0) {
+    return SERVERTYPE_DUMMY;
+  } else
+    return SERVERTYPE_NONE;
+}
+
+void BillingManager::Init() {
+  billing_api_handle_ = OpenBillingApi();
+  if (billing_api_handle_ == nullptr) {
+    LOG_ERROR("Fail to open billing api");
```
xiaowei-guan, line 30: If fail to open billing api, we need return.

packages/in_app_purchase_tizen/tizen/src/billing_manager.h
```diff
@@ -0,0 +1,39 @@
+#include <flutter/method_channel.h>
```
xiaowei-guan, line 1: 1.All header file need Macro
2.All file need add copyright.

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,160 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
+import '../billing_client_wrappers.dart';
+
+const String kPurchaseErrorCode = 'purchase_error';
+
+const String kIAPSource = 'samsung_checkout';
+
+late String appId;
```
wanchao-xu, line 17: If you define 'appId' as global variable, please name it 'kAppId', and if it is private, name it '_kAppId'.
And why not define 'appId' in class 'InAppPurchaseTizenPlatform'?

packages/in_app_purchase_tizen/tizen/src/billing_service_proxy.cc
```diff
@@ -0,0 +1,65 @@
...
+      Dlsym(handle, "service_billing_buyitem"));
+  if (service_billing_buyitem == nullptr) {
+    return 0;
+  }
+
+  service_billing_set_buyitem_cb = reinterpret_cast<FuncBillingSetBuyItemCb>(
+      Dlsym(handle, "service_billing_set_buyitem_cb"));
+  if (service_billing_set_buyitem_cb == nullptr) {
+    return 0;
+  }
```
wanchao-xu, line 56: Please return sucessfull value here.

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,264 @@
...
+      (void *)this);
+  if (!bRet) {
+    LOG_ERROR("service_billing_get_products_list failed");
+    return false;
+  }
+  return true;
+}
+
+bool BillingManager::GetPurchaseList(const flutter::EncodableValue *args) {
+  LOG_INFO("start get purchase list");
```
wanchao-xu, line 84: I think it's better to set 'const flutter::EncodableMap args' as parameter. Otherwise args may be null or not EncodableMap.

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,264 @@
...
+  std::string app_id;
+  std::string custom_id;
+  std::string country_code;
+  int page_num;
+  std::string check_value;
+  std::string server_type;
+  flutter::EncodableMap encodables = std::get<flutter::EncodableMap>(*args);
+  flutter::EncodableValue &app_id_value =
```
wanchao-xu, line 92: 'encodables' maybe not contain 'appId'

```
  const auto* app_id =
      std::get_if<std::string>(ValueOrNull(encodables, "appId"));
  if (!app_id ) {
    return false;
  }
```

packages/in_app_purchase_tizen/tizen/src/in_app_purchase_tizen_plugin.cc
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/billing_service_proxy.cc
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/in_app_purchase_tizen_plugin.cc
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
hyue7: fixed

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
hyue7: fixed

packages/in_app_purchase_tizen/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -0,0 +1,433 @@
...
+            code: isAvailable.result, message: isAvailable.resultTitle);
+    }
+    return resultRet;
+  }
+
+  Future<ProductsListApiResult> requestProducts(
+      List<String> requestparameters) async {
+    final String? productResponse =
+        await channel.invokeMethod<String?>('getProductList', {
+      'appId': requestparameters.asMap()[0] as String,
```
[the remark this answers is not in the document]
hyue7: fixed

packages/in_app_purchase_tizen/example/lib/main.dart
```diff
@@ -0,0 +1,356 @@
...
+// import 'package:in_app_purchase_platform_interface/in_app_purchase_platform_interface.dart';
+
+void main() {
+  WidgetsFlutterBinding.ensureInitialized();
+
+  InAppPurchaseTizenPlatform.registerPlatform();
+  runApp(_MyApp());
+}
+
+const String appId = '3201504002021';
```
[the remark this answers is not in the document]
hyue7: fixed

packages/in_app_purchase_tizen/tizen/src/billing_manager.h
hyue7: fixed

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
hyue7: fixed

swift-kim:
Please read https://github.com/flutter-tizen/flutter-tizen/wiki/New-plugin-submission-guideline and also resolve the CI failure.

packages/in_app_purchase_tizen/tizen/src/billing_manager.cc
hyue7: fixed

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,165 @@
...
+
+      purchaseDetails.status = PurchaseStatus.restored;
+      return purchaseDetails;
+    }).toList();
+    _purchaseUpdatedController.add(pastPurchases);
+  }
+
+  @override
+  Future<bool> buyNonConsumable({required PurchaseParam purchaseParam}) async {
+    final BillingBuyData billingResultWrapper = await billingManager.buyItem(
```
swift-kim, line 143: According to the documentation, this call should add a new event to `purchaseStream` when the purchase is complete. As far as I can see, `purchaseStream` is currently only updated on `restorePurchases`.

packages/in_app_purchase_tizen/lib/src/billing_manager_wrappers/billing_manager_wrapper.g.dart
```diff
@@ -0,0 +1,172 @@
...
+      'payResult': instance.payResult,
+      'payDetails': instance.payDetails,
+    };
+
+GetUserPurchaseListAPIResult _$GetUserPurchaseListAPIResultFromJson(
+        Map<String, dynamic> json) =>
+    GetUserPurchaseListAPIResult(
+      cPResult: json['CPResult'] as String? ?? '',
+      cPStatus: json['CPStatus'] as String? ?? '',
+      invoiceDetails: (json['invoiceDetails'] as List<dynamic>?)
```
swift-kim, line 102: You may need to regenerate this file.
```suggestion
      invoiceDetails: (json['InvoiceDetails'] as List<dynamic>?)
```

packages/in_app_purchase_tizen/.metadata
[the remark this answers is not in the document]
hyue7: fixed.
[Remove .metadata](https://github.com/flutter-tizen/plugins/pull/562/commits/b58b63ece13a1eb49d6dc040f40d76a1bd8b311b)
[Rename package name](https://github.com/flutter-tizen/plugins/pull/562/commits/8746dd915f8616c5b642b512dea05a7bed807687)

packages/in_app_purchase_tizen/lib/src/billing_manager_wrappers/billing_manager_wrapper.g.dart
hyue7: fixed.

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
hyue7: Fixed.
It will add the new purchase into `purchaseStream` when the `payResult` of `buyItem` is 'SUCCESS'.
[PurchaseUpdate if buyitem successfully](https://github.com/flutter-tizen/plugins/pull/562/commits/48ed9da0788db764fd3cb1bf16ade54732bb161d)

packages/in_app_purchase_tizen/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,165 @@
...
+    } else {
+      return false;
+    }
+  }
+
+  @override
+  Future<bool> buyConsumable(
+      {required PurchaseParam purchaseParam, bool autoConsume = true}) {
+    if (autoConsume) {
+      _productIdsToConsume.add(purchaseParam.productDetails.id);
```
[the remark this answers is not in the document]
hyue7: I don't find apis to implement `consumePurchase` in [SamsungCheckout documentation](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/implementing-the-purchase-process.html#), so I temporarily set it as auto-consumed on Tizen.
[Set buyconsumable to auto consum](https://github.com/flutter-tizen/plugins/pull/562/commits/58013deeb7e226298a29b135cea269742d8109df)

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+// To try without auto-consume, change `true` to `false` here.
+const bool _kAutoConsume = true;
+
+/// The [_kConsumableType]'s value:
+/// "1": CONSUMABLE
+/// "2": NON-CONSUMABLE
+/// "3": LIMITED-PERIOD
+/// "4": SUBSCRIPTION
+const int _kConsumableType = 1;
+const String _kAppId = '3201504002021';
```
swift-kim, line 30: What does the app ID mean in this context? Isn't an "app ID" a string that looks like `org.tizen.my_app`?

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+/// The [_kPurchaseCheckValue] is generated with [_kAppId] and [_kCustumId] and [_kCountryCode] and [_kItemType] and [_kSecurityKey];
+/// var hmacSha256 = Hmac(sha256, utf8.encode(_kSecurityKey));
+/// String _kPurchaseCheckValue = base64Url.encode(hmacSha256.convert(utf8.encode(_kAppId + _kCustumId + _kCountryCode + _kItemType)).bytes);
+String _kPurchaseCheckValue = 'eAKHVR/FUfgEmDb3KmtgkozWAKyEb1OWXKdvkUV4km0=';
+
+// WARNING: Don't change order of `_kRequestParams`.
+List<String> _kRequestParams = <String>[
+  _kAppId,
+  _kCountryCode,
+  _kItemType,
```
swift-kim, line 57: What is the item type? The parameter is not documented in the [API references](https://developer.samsung.com/smarttv/develop/api-references/samsung-product-api-references/billing-api.html#BillingManager-getProductsList).

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+  _kPageSize,
+  _kPageNum,
+  _kServerType,
+  _kProductCheckValue,
+  _kPurchaseCheckValue,
```
swift-kim, line 62: Should all these values be provided by the user? Can't they be calculated automatically or have default values? The API will not be very usable if all this information has to be provided by the user. Also `queryProductDetails` basically expects a `Set` of identifiers, not a list of params as I mentioned above.

packages/in_app_purchase/example/lib/main.dart
hyue7: I explain `_kAppId` in [readme](https://github.com/hyue7/plugins/tree/in_app_purchase/packages/in_app_purchase#preparation) :
_kAppId: The APP ID in [APP Details Setting](https://dpi.samsungcheckout.com/settings/appdetails) when user  register and login to [Samsung Checkout DPI portal](https://dpi.samsungcheckout.com/)

packages/in_app_purchase/example/lib/main.dart
hyue7: Generate `_kProductCheckValue` and `_kPurchaseCheckValue` need `_kAppId` and `_kSecurityKey`, which are both provided by user(in [Samsung Checkout DPI portal](https://dpi.samsungcheckout.com/))

packages/in_app_purchase/example/lib/main.dart
swift-kim: You're passing the value when retrieving the product list but not the purchase list. That's not the right place. Also I still doubt that you need to pass the value since

- The value is literally unused. The value is used to calculate `checkValue` but you already hardcoded the `checkValue`.
- The value is a constant (2), so the plugin already knows the value. There's no need to require the user to provide the value.

packages/in_app_purchase/example/lib/main.dart
swift-kim: Are the check values expected to be calculated in advance and hard-coded into the code (for real world apps)? Should `kSecurityKey` not be present in the user code for security reasons? (I can't find guidance from the API references.)

If not, you could provide utility functions that accept the required parameters + `kSecurityKey` and return the check values. (Preferably in the plugin API itself or otherwise in the example app code.)

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,185 @@
...
+        orderCurrencyId: purchaseParam.productDetails.currencyCode);
+
+    if (billingResultWrapper.payResult == 'SUCCESS') {
+      final String invoiceId =
+          billingResultWrapper.payDetails['InvoiceID'] ?? '';
+      final GetUserPurchaseListAPIResult responses = await billingManager
+          .requestPurchases(_identifiers.toList()[8], _identifiers.toList());
+
+      for (int i = 0; i < responses.invoiceDetails.length; i++) {
+        if (responses.invoiceDetails[i].invoiceID != invoiceId) {
```
swift-kim, line 156: If I understand correctly,
```suggestion
        if (responses.invoiceDetails[i].invoiceID == invoiceId) {
```

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
```diff
@@ -0,0 +1,185 @@
...
+            verificationData: PurchaseVerificationData(
+                localVerificationData: responses.invoiceDetails[i].invoiceID,
+                serverVerificationData: responses.invoiceDetails[i].invoiceID,
+                source: kIAPSource),
+            transactionDate: responses.invoiceDetails[i].orderTime,
+            status: const PurchaseStateConverter()
+                .toPurchaseStatus(responses.invoiceDetails[i].appliedStatus),
+          ));
+
+          _purchaseUpdatedController.add(purchases);
```
swift-kim, line 170: `buyNonConsumable` is supposed to return immediately, regardless of the actual result of the purchase, according to the API description:

> This method does not return the result of the purchase. Instead, after triggering this method, purchase updates will be sent to [purchaseStream].

So I'd suggest replacing the `await billingManager.requestPurchases()` part with `Future.then()` as follows:

```dart
billingManager
    .requestPurchases(...)
    .then((GetUserPurchaseListAPIResult responses) {
  for (...) {

  }
});
```
In this way the `requestPurchases` body is executed asynchronously after `buyNonConsumable` returns. (Consider adding `catchError` as well if `requestPurchases` may throw.)

packages/in_app_purchase/example/lib/main.dart
hyue7: Ok, I know, I will change it.

packages/in_app_purchase/example/lib/main.dart
hyue7: OK, I find the [documentation](https://developer.samsung.com/smarttv/develop/tizen-net-tv/guides/samsung-checkout/implementing-the-purchase-process.html#Generating-Check-Values), said generate check value during runtime, and not expose security key. 
But how can I not expose `kSecurityKey`? It is provided from user.

swift-kim: I don't know. The document says "You can use a key management server" so if the user wants to generate the check value at runtime, they could retrieve the key online. Otherwise the key (or the check value) should be hard-coded into the code as far as I can see.

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+/// var hmacSha256 = Hmac(sha256, utf8.encode(_kSecurityKey));
+/// String _kProductCheckValue = base64Url.encode(hmacSha256.convert(utf8.encode(_kAppId + _kCountryCode)).bytes);
+String _kProductCheckValue = 'sbKzQJrFdeLkfXTQUdjjq4jt7ykx79QmJno9ZLam+SM=';
+
+/// The [_kPurchaseCheckValue] is generated with [_kAppId] and [_kCustumId] and [_kCountryCode] and [_kItemType] and [_kSecurityKey];
+/// var hmacSha256 = Hmac(sha256, utf8.encode(_kSecurityKey));
+/// String _kPurchaseCheckValue = base64Url.encode(hmacSha256.convert(utf8.encode(_kAppId + _kCustumId + _kCountryCode + _kItemType)).bytes);
+String _kPurchaseCheckValue = 'eAKHVR/FUfgEmDb3KmtgkozWAKyEb1OWXKdvkUV4km0=';
+
+// WARNING: Don't change order of `_kRequestParams`.
```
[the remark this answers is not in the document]
hyue7: > `queryProductDetails` expects an unordered set of identifiers

Yes, i konw, but in `billingManager.requestProducts(identifiers.toList())` , a `Set` of `identifiers` is converted to list, this must be Ordered.

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
hyue7: fixed.
[Use Future.then() in buyNonConsumable](https://github.com/flutter-tizen/plugins/pull/562/commits/ac657875a0678afdb13bf91f5bb01df8ff1e8075)

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
hyue7: fixed.
[Use Future.then() in buyNonConsumable](https://github.com/flutter-tizen/plugins/pull/562/commits/ac657875a0678afdb13bf91f5bb01df8ff1e8075)

swift-kim: I said `catchError` (`Future.then(...).catchError(...)`) but not try-catch. The try-catch statement added to the commit seems pointless. You should `addError` to the stream controller (`_purchaseUpdatedController`) if you want to report any error.

packages/in_app_purchase/example/lib/main.dart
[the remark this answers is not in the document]
swift-kim: So I'm saying it's not the case for other platforms implementing the same API. If you really need to pass multiple parameters in a particular order to the API, it would be nice to minimize the number of the parameters to improve usability and avoid confusion as much as possible.

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+
+// To try without auto-consume, change `true` to `false` here.
+const bool _kAutoConsume = true;
+
+/// The [_kConsumableType]'s value:
+/// "1": CONSUMABLE
+/// "2": NON-CONSUMABLE
+/// "3": LIMITED-PERIOD
+/// "4": SUBSCRIPTION
+const int _kConsumableType = 1;
```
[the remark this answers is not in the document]
hyue7: fixed.
[Change itemType to an enum](https://github.com/flutter-tizen/plugins/pull/562/commits/46ec6f3acf1df7d7375bde24c98684f368c35797)

packages/in_app_purchase/example/lib/main.dart
hyue7: fixed

packages/in_app_purchase/example/lib/main.dart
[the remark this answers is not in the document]
hyue7: I add a new api in addition file: setRequestParameters, this can pass all params that getproductlist, getpurchaselist and buyitem need. Just call it before call` queryProductDetails`, however, the set of `identifiers` will not be used in `billingManager.requestProducts`.
[Add tizen specific api:setRequestParameters](https://github.com/flutter-tizen/plugins/pull/562/commits/adf7c11f89b0f3705a7385e30b12cafd68fee506)

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,372 @@
...
+List<String> _kRequestParams = <String>[
+  _kAppId,
+  _kCountryCode,
+  _kItemType,
+  _kPageSize,
+  _kPageNum,
+  _kServerType,
+  _kProductCheckValue,
+  _kPurchaseCheckValue,
+  _kCustumId,
```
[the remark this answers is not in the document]
hyue7: fixed

packages/in_app_purchase/example/lib/main.dart
hyue7: fixed.
Check value will be generated at runtime.

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
hyue7: fixed.
[add catchError if requestPurchases throw](https://github.com/flutter-tizen/plugins/pull/562/commits/632ba7fdef9f8d3bc05f96c9a45a9c34870cbf25)

packages/in_app_purchase/example/lib/main.dart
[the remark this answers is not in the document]
swift-kim: `requestProducts`, `requestPurchases`, and `buyItem` are different APIs and they require different parameters. For example, `buyItem` only requires `appId` and `serverType` as parameters, but in your design the developer must pass parameters like `customId` and `itemType` even if they doesn't want to call `requestPurchases`. This doesn't make much sense and the developer will have a hard time guessing what `pageSize` means in that context since `pageSize` is not documented in the `BuyItem` documentation. Do you think this is still okay? Is the limitation intentional, because you couldn't find a better design?

Also you're passing `_kRequestParams` to `queryProductDetails` redundantly. You could pass an empty set if they are not used.

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform.dart
swift-kim: You shouldn't throw anything in the `catchError` clause. No one can handle the exception.

As I mentioned above, you can either `addError` to the stream controller or just ignore the error.

hyue7: ok,  just ignore error

packages/in_app_purchase/example/lib/main.dart
hyue7: Set `itemType` as a constant value '2', hardcoded into requestPurchases.

packages/in_app_purchase/example/lib/main.dart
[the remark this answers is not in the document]
hyue7: Have modified the api.
In `setRequestParameters`, only `appId` is required, users can pass the parameters they want. If miss some param, an error will be throw by native side.

packages/in_app_purchase/tizen/src/billing_service_proxy.h
```diff
@@ -0,0 +1,70 @@
...
+void CloseApi(void *handle);
+
+extern FuncGetProductslist service_billing_get_products_list;
+extern FuncGetpurchaselist service_billing_get_purchase_list;
+extern FuncBillingBuyItem service_billing_buyitem;
+extern FuncBillingSetBuyItemCb service_billing_set_buyitem_cb;
+extern FuncServiceBillingIsServiceAvailable
+    service_billing_is_service_available;
+extern FuncServiceBillingVerifyInvoice service_billing_verify_invoice;
+extern FuncSsoGetLoginInfo sso_get_login_info;
```
Swanseo0, line 68: Is there a specific reason why it needs to be used as a global variable? If not, how about encapsulation, like `EwkInternalApiBinding` class in the webview_flutter_tizen.

packages/in_app_purchase/pubspec.yaml
```diff
@@ -0,0 +1,31 @@
+name: in_app_purchase_tizen
+description: An implementation for the Tizen platform of the Flutter `in_app_purchase` plugin. This uses the Samsung Checkout APIs.
+homepage: https://github.com/flutter-tizen/plugins
+repository: https://github.com/flutter-tizen/plugins/tree/master/packages/in_app_purchase
+version: 0.1.0
+
+environment:
+  sdk: ">=2.17.0 <3.0.0"
+  flutter: ">=3.0.0"
```
Swanseo0, line 9: Please update.
```
environment:
  sdk: ">=2.18.0 <4.0.0"
  flutter: ">=3.3.0"
```

packages/in_app_purchase/example/lib/main.dart
```diff
@@ -0,0 +1,360 @@
...
+
+// To try without auto-consume, change `true` to `false` here.
+const bool _kAutoConsume = true;
+
+const String _kAppId = '3201504002021';
+const String _kServerType = 'DEV';
+const String _kCountryCode = 'US';
+const int _kPageSize = 20;
+const int _kPageNum = 1;
+const String _kSecurityKey = 'YxE757K+aDWHJXa0QMnL5AJmItefoEizvv8L7WPJAMs=';
```
Swanseo0, line 28: What kind of key is this? Is it an officially provided test key that can be used for testing purposes, or is it someone's personal security key? If it is someone's personal key, should we only indicate the need to input the security as a comment?

swift-kim:
Just rebased the commits.

packages/in_app_purchase/example/lib/main.dart
hyue7: > Is it an officially provided test key that can be used for testing purposes, or is it someone's personal security key?

`_kSecurityKey` is provided by users, it is a personal key. As the [documentation ](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/implementing-the-purchase-process.html#Generating-Check-Values)said: 
The DPI security key is used only by Samsung Smart TV applications for open API calls. Do not expose this key. You can use a key management server for greater security.
Users can retrieve the key on their own.

packages/in_app_purchase/example/tizen/tizen-manifest.xml
```diff
@@ -0,0 +1,15 @@
...
+    <profile name="tv"/>
+    <ui-application appid="org.tizen.in_app_purchase_tizen_example" exec="Runner.dll" type="dotnet" multiple="false" nodisplay="false" taskmanage="true">
+        <label>in_app_purchase_tizen_example</label>
+        <icon>ic_launcher.png</icon>
+        <metadata key="http://tizen.org/metadata/prefer_dotnet_aot" value="true"/>
+    </ui-application>
+    <feature name="http://tizen.org/feature/screen.size.all"/>
+    <privileges>
+        <privilege>http://developer.samsung.com/privilege/billing</privilege>
+        <privilege>http://developer.samsung.com/privilege/sso.partner</privilege>
```
swift-kim, line 12: Is this privilege (`sso.partner`) mandatory? It's not listed in the privilege list in README.md.

If this is a partner-level privilege and requires a partner-level certificate, it should be documented properly (see flutter_webrtc_tizen or video_player_videohole).

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform_addition.dart
```diff
@@ -0,0 +1,58 @@
...
+  final BillingManager _billingManager;
+
+  /// Set all request parameters that SamsungCheckout DPI Portal needed.
+  void setRequestParameters({
+    required String appId,
+    String? countryCode,
+    int? pageSize,
+    int? pageNum,
+    String? serverType,
+    String? securityKey,
```
swift-kim, line 25: Please provide some descriptions of the arguments (including how to get each value and which values are required by which API) in the doc comments or README.

https://dart.dev/effective-dart/documentation#doc-comments

packages/in_app_purchase/example/lib/main.dart
hyue7: Fixed.

packages/in_app_purchase/tizen/src/billing_service_proxy.h
hyue7: Fixed.

packages/in_app_purchase/example/tizen/tizen-manifest.xml
hyue7: `sso.partner` is used for calling `sso_get_login_info` to get customId.
Fixed.

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform_addition.dart
hyue7: Fixed.

packages/in_app_purchase/example/tizen/tizen-manifest.xml
swift-kim: Thanks. I added some additional information and removed redundant information.

packages/in_app_purchase/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Then you can import `in_app_purchase` and `in_app_purchase_tizen` in your Dart code:
+
+```dart
+import 'package:in_app_purchase/in_app_purchase.dart';
+import 'package:in_app_purchase_tizen/in_app_purchase_tizen.dart';
+```
+
+You must call `setRequestParameters` to set required parameters before making any plugin API call.
+You can find your `countryCode` in [Samsung Developers: Implementing the Purchase Process](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/implementing-the-purchase-process.html#Country-and-Currency-Codes) and `serverType` in [Samsung Developers: Billing API References](https://developer.samsung.com/smarttv/develop/api-references/samsung-product-api-references/billing-api.html#1.-Type-Definitions)
```
swift-kim, line 49: Is the country and currency code the user's or the app developer's? What if I added multiple countries/regions in the product detail settings?

Also how can I determine the server type (how should I configure my product for each server in the DPI portal)?

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform_addition.dart
```diff
@@ -0,0 +1,76 @@
...
+class InAppPurchaseTizenPlatformAddition extends InAppPurchasePlatformAddition {
+  /// Creates a [InAppPurchaseTizenPlatformAddition] which uses the supplied
+  /// `BillingManager` to provide Tizen specific features.
+  InAppPurchaseTizenPlatformAddition(this._billingManager);
+
+  final BillingManager _billingManager;
+
+  /// Set all request parameters that SamsungCheckout DPI Portal needed.
+  ///
+  /// The `appId` is your application id, must be required.
```
swift-kim, line 20: "must (be set)" and "required" are synonymous.

packages/in_app_purchase/lib/src/in_app_purchase_tizen_platform_addition.dart
```diff
@@ -0,0 +1,76 @@
...
+  ///
+  /// The `serverType` is Billing server type, must be required.
+  ///
+  /// The `countryCode` is your device country code.
+  /// Use it when call `queryProductDetails`, `restorePurchases`, `buyConsumable` and `buyNonConsumable`.
+  ///
+  /// The `pageSize` is the number of products retrieved per page, more than 1 and less than 100.
+  /// Use it when call `queryProductDetails`.
+  ///
+  /// The `pageNum` is the requested page number, more than 1.
```
swift-kim, line 30: "more than 1" means a number larger than 1, that is, 2 or higher.

How about `(>= 1)`, `(1-N)`, or `(min 1)`?

packages/in_app_purchase/README.md
swift-kim: I noticed this information in the Billing Info page of the Seller Office:

> The app must check checkout-available country/location before calling the billing API. Ensure that the 'Billing Service Available Country/location Check API’ is integrated to your app to check checkout-available country/location.

> Please ensure that Checkout is supported in your service country/location by using the Billing Service Available Country/location Check API. Once confirmed, your application can call the billing API at [Checking Billing Service Available Country/Location.](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/implementing-the-purchase-process.html#Checking-Billing-Service-Available-Country%2FLocation)

so "your service country" probably refers to where the end user (consumer) lives. How do I know where the user's region is? Also do we have an API for checking the service availability for the region in Dart?

swift-kim:
The example app works with your account info but when I replace the app ID and security key in the code with my own, the app shows the "Unable to connect to the payments processor" error message as follows:



I enabled "Samsung Checkout" in the Seller Office and changed the package ID in `tizen-manifest.xml` to my registered package ID. Can you guess what I am missing?

hyue7:
> The example app works with your account info but when I replace the app ID and security key in the code with my own, the app shows the "Unable to connect to the payments processor" error message as follows:
> 
> 
> 
> I enabled "Samsung Checkout" in the Seller Office and changed the package ID in `tizen-manifest.xml` to my registered package ID. Can you guess what I am missing?


Can you catch the logs? 
log tags:
```
dlogutil InAppPurchaseTizenPlugin&
dlogutil BILLING_NATIVE_PLUGIN&
dlogutil BILLING_MANAGED_PLUGIN&
dlogutil BLClient&
dlogutil IPCProxy&
dlogutil CSBilling&
```
It seems not login Samsung account(Although I know you may have logged in).
I test my own appid and key, it works. Or you can send the appid and key to me, i will test.

packages/in_app_purchase/README.md
hyue7: > so "your service country" probably refers to where the end user (consumer) lives. How do I know where the user's region is?
As I can see the [prerequisites ](https://developer.samsung.com/smarttv/develop/guides/samsung-checkout/implementing-the-purchase-process.html#Prerequisites),  the countryCode is (user's) TV countryCode. 
Web use `getSystemConfig` to get the countryCode.

> Also do we have an API for checking the service availability for the region in Dart?
I will find a capi of checking service avaiable country/location.

swift-kim:
@hyue7 I tried both "KR" and "US" but it didn't help. I don't think it's even possible to be "not logged in" since I can't access the Apps page on TV without setting up a Samsung account as far as I know (so the error message is a bit misleading).

My TV is a consumer model so I can't see the logs. I'll send you my app ID and key.

swift-kim (approved):
I'm waiting for a response from the Checkout team regarding the above issue. You can merge this PR if there are no more reviews from other maintainers. I don't quite have time to fully review the implementation.

packages/in_app_purchase/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,365 @@
...
+      std::get_if<flutter::EncodableMap>(method_call.arguments());
+  try {
+    const auto &method_name = method_call.method_name();
+    if (method_name == "getProductList") {
+      if (!encodables) {
+        result->Error(kInvalidArgument, "No arguments provided");
+        return;
+      }
+      if (GetProductList(encodables)) {
+        method_result_ = std::move(result);
```
wanchao-xu, line 256: How to make sure the 'method_result_' is not changed by other function?

packages/in_app_purchase/tizen/src/billing_manager.h
```diff
@@ -0,0 +1,183 @@
...
+  SYSTEM_INFO_ERROR_OUT_OF_MEMORY =
+      TIZEN_ERROR_OUT_OF_MEMORY, /**< Out of memory */
+  SYSTEM_INFO_ERROR_IO_ERROR =
+      TIZEN_ERROR_IO_ERROR, /**< An input/output error occurred when reading
+                               value from system */
+  SYSTEM_INFO_ERROR_PERMISSION_DENIED =
+      TIZEN_ERROR_PERMISSION_DENIED, /**< No permission to use the API */
+  SYSTEM_INFO_ERROR_NOT_SUPPORTED =
+      TIZEN_ERROR_NOT_SUPPORTED, /**< Not supported parameter (Since 3.0) */
+} system_info_error_e;
```
wanchao-xu, line 48: 'system_info_error_e' is inclued in head file 'system_info.h', why not use system_info.h'?

packages/in_app_purchase/tizen/src/billing_manager.cc
```diff
@@ -0,0 +1,365 @@
...
+            billing_server_type_ = SERVERTYPE_DEV;
+            break;
+          default:
+            billing_server_type_ = SERVERTYPE_NONE;
+            break;
+        }
+        LOG_INFO("[BillingManager] Billing_Server_Type is %d",
+                 billing_server_type_);
+      } else {
+        LOG_ERROR("[BillingManager] Fail to get TV server type.");
```
wanchao-xu, line 134: Please call 'dlclose(handle)' before return.

packages/in_app_purchase/tizen/src/billing_service_proxy.h
```diff
@@ -0,0 +1,71 @@
...
+    return instance;
+  }
+
+  ~BillingWrapper();
+
+  BillingWrapper(const BillingWrapper &) = delete;
+  BillingWrapper &operator=(const BillingWrapper &) = delete;
+
+  bool Initialize();
+  FuncGetProductslist service_billing_get_products_list = nullptr;
```
wanchao-xu, line 57: I think it's better not to set public for member variable.

hyue7: These variables are used in billing_manager.cc, must set public.

hyue7: Fixed!

packages/in_app_purchase/tizen/src/billing_manager.cc
hyue7: Fixed!

packages/in_app_purchase/tizen/src/billing_manager.cc
hyue7: Fixed!

packages/in_app_purchase/tizen/src/in_app_purchase_tizen_plugin.cc
```diff
@@ -0,0 +1,228 @@
...
+  const auto *encodables =
+      std::get_if<flutter::EncodableMap>(method_call.arguments());
+  const auto &method_name = method_call.method_name();
+
+  try {
+    if (method_name == "getProductList") {
+      if (!encodables) {
+        result->Error(kInvalidArgument, "No arguments provided");
+        return;
+      }
```
xiaowei-guan, line 118: we can move these code to beginning of function, and remove repeat code.

packages/in_app_purchase/tizen/src/in_app_purchase_tizen_plugin.cc
```diff
@@ -0,0 +1,228 @@
...
+      GetProductList(encodables, std::move(result));
+    } else if (method_name == "getPurchaseList") {
```
xiaowei-guan, line 120: we can remove all if ...  else...  by map<key, function>

xuelian-bai (approved):

xiaowei-guan (approved):

PR #550, opened 2023-03-24
[video_player_videohole] Add video_player_videohole package

hyue7:
”video_player_videohole“ is a new video player plugin ,used to play videos of DRM (playready & widevine).
It's based on MM player and only supports on Tizen TV devices.

PR #564, opened 2023-04-26
[webview_flutter_tizen] Fix YouTube player on TV

Swanseo0:


packages/webview_flutter/tizen/src/ewk_internal_api_binding.cc
```diff
@@ -35,6 +35,8 @@ bool EwkInternalApiBinding::Initialize() {
       dlsym(handle_, "ewk_view_ime_window_set"));
   view.KeyEventsEnabledSet = reinterpret_cast<EwkViewKeyEventsEnabledSetFnPtr>(
       dlsym(handle_, "ewk_view_key_events_enabled_set"));
+  view.SupportVideoHoleSet = reinterpret_cast<EwkViewSupportVideoHoleSetFnPtr>(
+      dlsym(handle_, "ewk_view_set_support_video_hole"));
```
bbrto21, line 39: Doesn't this require a profile distinction?

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -19,6 +19,12 @@
 
 namespace {
 
+#ifdef TV_PROFILE
+constexpr double kProfileFactor = 1.6;
+#else
+constexpr double kProfileFactor = 1.0;
+#endif
```
bbrto21, line 26: At the current stage, using these constants is fine to me. 
but IMO, this should be passed by the embedder to the platform view.

swift-kim: FYI) The terms "scale factor" and "device pixel ratio" are interchangeable, but "profile factor" is not. The profile factor is predefined as a constant for each profile.

https://docs.tizen.org/application/native/guides/ui/efl/multiple-screens/#calculating-the-base-scale

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -169,6 +175,14 @@ void WebView::Dispose() {
   }
 }
 
+void WebView::Offset(double left, double top) {
+  left_ = left * kProfileFactor;
+  top_ = top * kProfileFactor;
+
+  evas_object_move(webview_instance_, static_cast<int32_t>(left_),
+                   static_cast<int32_t>(top_));
```
JSUYA, line 183: The argument type of evas_object_move is int.

JSUYA (commented):
Please update the README and CHANGELOG.
and webview_flutter_lwe also needs update.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -283,6 +297,11 @@ void WebView::InitWebView() {
   EwkInternalApiBinding::GetInstance().view.KeyEventsEnabledSet(
       webview_instance_, true);
 
+#ifdef TV_PROFILE
+  EwkInternalApiBinding::GetInstance().view.SupportVideoHoleSet(
+      webview_instance_, window_, true, false);
```
JSUYA, line 302: In order to check that this API actually works, the example needs to be modified.
We should to choose one of two ways.
1. Modify the example(TODO) and write the related contents in the readme.
```diff
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
   @override
   Widget build(BuildContext context) {
     return Scaffold(
+       backgroundColor: Color.fromARGB(0, 0, 0, 0),
+       // backgroundColor: Colors.green,
       appBar: AppBar(
         title: const Text('Flutter WebView example'),

```
2. Referring to the [video hole pr](https://github.com/flutter-tizen/plugins/pull/550/files#diff-38e843fda10b05b7020e435141400d2b62ec8311e913ecb815f48e6e7cd1c830), I think the webview widget can be blended. (I don't know if that's possible.... But as of now, I don't think there is a way...)

packages/webview_flutter/tizen/src/ewk_internal_api_binding.cc
Swanseo0: Although it is not used in `TV_PROFILE`, I thought it was not necessary to distinguish registering in `EwkViewProcTable`. Do you think it is necessary?

PR #568, opened 2023-05-08
[video_player_videohole] Update pigeon to 10.0.0

swift-kim:
Also rename `TizenVideoPlayerApi` to `VideoPlayerVideoholeApi` to avoid any possible conflict with `video_player_tizen`.


hyue7 (approved):
Looks good to me.

PR #571, opened 2023-05-11
[video_player] Fix memory leak issue

xiaowei-guan:
OnVideoFrameDecoded callback may be received during destory MM player. So check is_initialized_ value when receive frame decoded callback.

### DIFF
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.6
+
+* Fix memory leak issue.
+
 ## 2.4.5
 
 * Update README with supported devices information.
--- a/packages/video_player/CHANGELOG.md
+++ b/packages/video_player/CHANGELOG.md
@@ -1,9 +1,6 @@
 ## 2.4.6
 
 * Fix memory leak issue.
-
-## NEXT
-
 * Increase the minimum Flutter version to 3.3.
 
 ## 2.4.5
--- a/packages/video_player/tizen/src/video_player.cc
+++ b/packages/video_player/tizen/src/video_player.cc
@@ -158,7 +158,19 @@ VideoPlayer::VideoPlayer(flutter::PluginRegistrar *plugin_registrar,
   SetUpEventChannel(plugin_registrar->messenger());
 }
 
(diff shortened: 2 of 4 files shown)


wanchao-xu (approved):
Looks good to me

packages/video_player/tizen/src/video_player.cc
```diff
@@ -158,7 +158,19 @@ VideoPlayer::VideoPlayer(flutter::PluginRegistrar *plugin_registrar,
   SetUpEventChannel(plugin_registrar->messenger());
 }
 
-VideoPlayer::~VideoPlayer() { Dispose(); }
+VideoPlayer::~VideoPlayer() {
```
swift-kim, line 161: Is it no longer necessary to call `Dispose` from the destructor? Should a similar thing be applied to video_player_videohole for consistency?

xiaowei-guan:
```cpp
std::optional<FlutterError> VideoPlayerTizenPlugin::Dispose(
    const TextureMessage &msg) {
  LOG_DEBUG("[VideoPlayerTizenPlugin] textureId: %ld", msg.texture_id());

  auto iter = players_.find(msg.texture_id());
  if (iter != players_.end()) {
    iter->second->Dispose();
    players_.erase(iter);
  }

  return std::nullopt;
}
```
we don't need call dispose int destructor,  will cause call dispose() twice.

haesik:
Can this PR solve the robot video player issue?

packages/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.4.6
+
+* Fix memory leak issue.
+
 ## NEXT
 
 * Increase the minimum Flutter version to 3.3.
```
swift-kim, line 7:
```suggestion
## 2.4.6

* Fix a memory leak issue.
* Increase the minimum Flutter version to 3.3.
```

xiaowei-guan: ok

swift-kim (approved):

swift-kim:
Was @haesik's question answered?

xuelian-bai:
memory leak issue has been solve

PR #584, opened 2023-05-26
[tool] Use new pub cache location for the publish command

swift-kim:
The pub cache location has changed in Dart 3.0.

This change should land as soon as possible to fix our [broken CI](https://github.com/flutter-tizen/plugins/actions/runs/5091477987).

Reference: https://github.com/flutter/packages/commit/5b7d732939860d38aeb6d255aa23713788c9833d

### DIFF
--- a/tools/lib/src/publish_command.dart
+++ b/tools/lib/src/publish_command.dart
@@ -73,10 +73,10 @@ class PublishCommand extends PackageLoopingCommand {
   static const String _pubCredentialName = 'PUB_CREDENTIALS';
 
   @override
-  String get name => 'publish';
+  final String name = 'publish';
 
   @override
-  String get description => 'Attempts to publish the given packages to pub.\n'
+  final String description = 'Attempts to publish the given packages to pub.\n'
       'If running this on CI, an environment variable named $_pubCredentialName must be set to a String that represents the pub credential JSON.\n'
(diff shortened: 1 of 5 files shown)


tools/lib/src/publish_command.dart
```diff
@@ -256,51 +256,52 @@ Safe to ignore if the package is deleted in this commit.
...
     }
   }
+  if (configHome != null) {
+    configDir = path.join(configHome, 'dart');
+  }
 
-  if (cacheDir == null) {
-    printError('Unable to determine pub cache location');
+  if (configDir == null) {
+    printError('Unable to determine pub con location');
```
bbrto21, line 302: I think `config` or `conf` is helpful for understanding.

```suggestion
    printError('Unable to determine pub config location');
```

bbrto21 (approved):

tools/lib/src/publish_command.dart
swift-kim: I noticed but I wanted to keep the original code as it is :)

https://github.com/flutter/packages/blob/4a5cc3e6947b2b9840d7d810fe310e00726803fb/script/tool/lib/src/publish_command.dart#L459

bbrto21: OK :)

PR #590, opened 2023-06-15
[device_info_plus] Add TizenDeviceInfo.data property

swift-kim:
- Add `TizenDeviceInfo.data` which represents the device info as a map. ([source](https://github.com/fluttercommunity/plus_plugins/blob/a246de0323c454b0a89dfa0fc1ba6e7aa58ef477/packages/device_info_plus/device_info_plus_platform_interface/lib/model/base_device_info.dart#L7))
- Disambiguate the method channel name to avoid a possible conflict with the original device_info_plus.

### DIFF
--- a/packages/device_info_plus/lib/device_info_plus_tizen.dart
+++ b/packages/device_info_plus/lib/device_info_plus_tizen.dart
@@ -86,7 +86,7 @@ class TizenDeviceInfo {
   /// http://tizen.org/system/tizenid
   final String? tizenId;
 
-  /// Deserializes from the message received from [_kChannel].
+  /// Creates a [TizenDeviceInfo] from the [map].
   static TizenDeviceInfo fromMap(Map<String, dynamic> map) {
     return TizenDeviceInfo(
       modelName: map['modelName'],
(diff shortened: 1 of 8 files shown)


packages/device_info_plus/lib/device_info_plus_tizen.dart
```diff
@@ -109,12 +109,33 @@ class TizenDeviceInfo {
...
+        'buildTime': buildTime,
+        'buildType': buildType,
+        'buildVariant': buildVariant,
+        'buildRelease': buildRelease,
+        'deviceType': deviceType,
+        'manufacturer': manufacturer,
+        'platformName': platformName,
+        'platformProcessor': platformProcessor,
+        'tizenId': tizenId,
+      };
```
JSUYA, line 133: Can you explain where this data getter is used? I couldn't find out why this is needed.

JSUYA: Thank you for explain. I missed PR comment.
I think we need to add the toString() API too.

https://github.com/fluttercommunity/plus_plugins/blob/a246de0323c454b0a89dfa0fc1ba6e7aa58ef477/packages/device_info_plus/device_info_plus_platform_interface/lib/model/base_device_info.dart#L13-L16

swift-kim: Okay, let me clean up a bit.

packages/device_info_plus/lib/device_info_plus_tizen.dart
```diff
@@ -10,8 +10,8 @@ import 'package:flutter/services.dart';
 ///
 /// See: https://docs.tizen.org/application/native/guides/device/system
 class TizenDeviceInfo {
-  /// Tizen device info class.
-  TizenDeviceInfo({
+  TizenDeviceInfo._({
```
swift-kim, line 13: Technically this is a breaking change but I don't think there's any use case.

packages/device_info_plus/lib/device_info_plus_tizen.dart
swift-kim: Done. Thanks!

packages/device_info_plus/lib/device_info_plus_tizen.dart
JSUYA: Just curious! is there a reason why a named constructor is necessary here technically? kind of pattern?

JSUYA (approved):

packages/device_info_plus/lib/device_info_plus_tizen.dart
swift-kim: This just changes the visibility of the constructor. Any member that starts with `_` is considered private in Dart.

PR #601, opened 2023-07-20
[shared_preferences] Update to support shared_preferences 2.2.0

kazakan:
- Update shared_preferences to 2.2.0.
- Update shared_preferences_platform_interface to 2.3.0.
- Implement `clearWithParameters` and `getAllWithParameters`
- Update integration_test.

### DIFF
--- a/packages/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/CHANGELOG.md
@@ -1,6 +1,9 @@
-## NEXT
+## 2.1.1
 
 * Increase the minimum Flutter version to 3.3.
+* Update shared_preferences to 2.2.0.
+* Update shared_preferences_interface to 2.3.0.
+* Update integration_test.
 
 ## 2.1.0
 
--- a/packages/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Increase the minimum Flutter version to 3.3.
 * Update shared_preferences to 2.2.0.
 * Update shared_preferences_interface to 2.3.0.
+* Add `clearWithParameters` and `getAllWithParameters`.
 * Update integration_test.
 
 ## 2.1.0
--- a/packages/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Update shared_preferences to 2.2.0.
 * Update shared_preferences_interface to 2.3.0.
 * Add `clearWithParameters` and `getAllWithParameters`.
+* Update `clear` to use `clearWithParameters`
+* Update `getAll` to use `getAllWithParameters`
 * Update integration_test.
 
 ## 2.1.0
--- a/packages/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.1.1
+## 2.2.0
 
 * Increase the minimum Flutter version to 3.3.
 * Update shared_preferences to 2.2.0.
--- a/packages/shared_preferences/lib/shared_preferences_tizen.dart
+++ b/packages/shared_preferences/lib/shared_preferences_tizen.dart
@@ -94,7 +94,7 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
         failed |= !(await remove(key));
       }
     }
-    return failed;
+    return !failed;
   }
 
   @override
--- a/packages/shared_preferences/lib/shared_preferences_tizen.dart
+++ b/packages/shared_preferences/lib/shared_preferences_tizen.dart
@@ -7,6 +7,7 @@ import 'dart:ffi';
 
 import 'package:ffi/ffi.dart';
 import 'package:shared_preferences_platform_interface/shared_preferences_platform_interface.dart';
+import 'package:shared_preferences_platform_interface/types.dart';
 import 'package:tizen_interop/4.0/tizen.dart';
 
 /// The Tizen implementation of [SharedPreferencesStorePlatform].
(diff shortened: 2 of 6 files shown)


packages/shared_preferences/lib/shared_preferences_tizen.dart
```diff
@@ -82,9 +83,34 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
...
 
+  @override
+  Future<bool> clearWithParameters(ClearParameters parameters) async {
+    final PreferencesFilter filter = parameters.filter;
+    final List<String> keys = List<String>.of(_preferences.keys);
+    bool failed = false;
+    for (final String key in keys) {
+      if (key.startsWith(filter.prefix) &&
+          (filter.allowList == null || filter.allowList!.contains(key))) {
+        failed |= !(await remove(key));
```
JSUYA, line 94: `!(await remove(key))` This code returns false even if remove succeeds.

packages/shared_preferences/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
-## NEXT
+## 2.2.0
 
 * Increase the minimum Flutter version to 3.3.
+* Update shared_preferences to 2.2.0.
+* Update shared_preferences_interface to 2.3.0.
+* Update integration_test.
```
JSUYA, line 6: The point of this update is the addition of two APIs. It would be good to write this down as well. (clearWithParameters, getAllWithParameters)

JSUYA (commented):
Have you tested the newly added API? There is no this api's unit test in shared_preferences's [integration_test](https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart). (Tested on each platform implementation.)


In this case, it is better to write the results in the comment after testing.

packages/shared_preferences/lib/shared_preferences_tizen.dart
kazakan: About integration_test, `getAllWithParameters` and `clearWithParameters` are used by other methods in original shared_preferences package. It seems they omit test about two api's because if they don't work properly, other methods won't work properly.

kazakan:
> Have you tested the newly added API? There is no this api's unit test in shared_preferences's [integration_test](https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart). (Tested on each platform implementation.)

`getAllWithParameters` and `clearWithParameters` are used by other methods in original shared_preferences package. It seems they omit test about two api's because if they don't work properly, other methods won't work properly.

packages/shared_preferences/lib/shared_preferences_tizen.dart
kazakan: Fix by returning `!failed` at last of this method. This makes return `false` when `remove` of any key failed.

packages/shared_preferences/CHANGELOG.md
kazakan: Added it!

packages/shared_preferences/lib/shared_preferences_tizen.dart
JSUYA: OK but I don't know if we should stop or continue the for loop if remove fails. What do you think?
The Linux implementation writes `preference` and returns false if an exception occurs. when an exception occurs, further writing stops.

(If you will the for loop to stop when remove fails, you can replace `failed` with `ret` (like other code).)

JSUYA:
> > Have you tested the newly added API? There is no this api's unit test in shared_preferences's [integration_test](https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart). (Tested on each platform implementation.)
> 
> `getAllWithParameters` and `clearWithParameters` are used by other methods in original shared_preferences package. It seems they omit test about two api's because if they don't work properly, other methods won't work properly.

However, even before clearWithParameters's failed was fixed, integration_test was still all pass.

packages/shared_preferences/lib/shared_preferences_tizen.dart
kazakan: Then it would be good to stop on exception like other implementations for unity. I'll fix it.

kazakan:
> > > Have you tested the newly added API? There is no this api's unit test in shared_preferences's [integration_test](https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart). (Tested on each platform implementation.)
> > 
> > 
> > `getAllWithParameters` and `clearWithParameters` are used by other methods in original shared_preferences package. It seems they omit test about two api's because if they don't work properly, other methods won't work properly.
> 
> However, even before clearWithParameters's failed was fixed, integration_test was still all pass.

Yes. I think I was wrong. It seems the test of original package missed this point. For more accurate test, we need to add more test code.

packages/shared_preferences/lib/shared_preferences_tizen.dart
kazakan: Fixed to return false right after remove fail.

JSUYA (approved):
LGTM

packages/shared_preferences/lib/shared_preferences_tizen.dart
```diff
@@ -82,9 +83,36 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
     return tizen.preference_remove_all() == 0;
   }
```
kazakan, line 84: Other implementation uses `clearWithParameters` in `clear` but this implement doesn't. Wouldn't this be a big problem in future?

JSUYA: You are right. it should modified before merged.

kazakan: I modified it.

Swanseo0 (approved):
LGTM!

PR #602, opened 2023-07-24
[video_player_videohole] Fix web build

MasterSon05:


swift-kim:
`video_player_videohole` has not been tested for web. Consider using `video_player` (1st-party plugin) instead of `video_player_videohole` when building for web (see the note in the [plugin usage](https://pub.dev/packages/video_player_videohole#usage)).

MasterSon05:
> `video_player_videohole` has not been tested for web. Consider using `video_player` (1st-party plugin) instead of `video_player_videohole` when building for web (see the note in the [plugin usage](https://pub.dev/packages/video_player_videohole#usage)).

Yes, you are right, but if a cross platform application is developed, adding the package to the dependencies completely prevents the web build. Logic can be added in the code according to the platform, but even importing the old package prevents web build from being imported. In this way, we fix the build for the web platform. Because it may need to be used in different scenarios with the web version of the same application.

swift-kim:
https://github.com/swift-kim/plugins/blob/videohole-test/packages/video_player_videohole/example/lib/main.dart#L12

I managed to build the example app for both Tizen and web, although there's a huge amount of duplicated code.

MasterSon05:
**Error during web build without changes in PR:**
```
Error: Dart library 'dart:ffi' is not available on this platform.
import 'dart:ffi' hide Size;
       ^
Context: The unavailable library 'dart:ffi' is imported through these packages:
```

swift-kim:
I have provided a working example in my previous comment. Please show me your code if you encountered an error.

MasterSon05:
> I have provided a working example in my previous comment. Please show me your code if you encountered an error.

In my usage scenario, I added a DRM-supporting web player myself instead of the video_player package for the web. So in this case I need to create a noop version within the application for everything I use from the video_player_hole package. So I think it shouldn't prevent getting web build independently of the package. Therefore, it would be a healthier approach to make this change in the package.

swift-kim:
Did you see the example? You can use the conditional import (`if (dart.library.html)`) to prevent importing source files when building for web.

MasterSon05:
> Did you see the example? You can use the conditional import (`if (dart.library.html)`) to prevent importing source files when building for web.

Yes, I saw the example, but if we do not use the video_player package for the web in return for the video_player_videohole package, we have to create the noop version for each class and method we import from the video_player_videohole package. However, I think it is a healthier approach to handle this situation in the video_player_videohole package.

swift-kim:
Your app structure should look like this:

- `my_player_tizen.dart`: Provides an `Mp4RemoteVideo` widget implemented via the `video_player_videohole` package.
- `my_player_web.dart`: Provides an `Mp4RemoteVideo` widget implemented via the DRM-enabled web video player.
- `main.dart`: Adds `Mp4RemoteVideo` to the app's widget tree by conditionally importing `my_player_tizen.dart` or `my_player_web.dart` depending on the current platform.

You can even create a placeholder `Mp4RemoteVideo` implementation (e.g. empty container) for non-Tizen and non-web platforms.

Is there anything wrong with this design? Please show your code if you think it doesn't make sense.

MasterSon05:
> 

I created a common interface with better_player and I use it that way. 
I also created such a widget structure in the widget section. 
`  @override
  Widget build(BuildContext context) {
    final playerController = widget.controller;
    if (playerController is BetterPlayerController) {
      return BetterPlayerControllerProvider(
        controller: playerController,
        child: _buildPlayer(),
      );
    } else if (playerController is TizenPlayerController) {
      return ValueListenableBuilder<VideoPlayerController?>(
          valueListenable: playerController.videoPlayerControllerNotifier,
          builder: (context, controller, _) {
            if (controller != null) {
              return TizenPlayer(controller: controller);
            }
            return const SizedBox();
          });
    }

    return const SizedBox();
  }`
Since I use the package at more than one point, I need to create a placeholder for each class or method.
It works fine with the fork I created. All I do is `registerDrmCallback(drmConfigs, _playerId);` By separating the function, I imported it only in cases where dart ffi can be used.
In this way, I think that the code is more readable and that it can improve the developer's experience in many different scenarios.
I don't understand why you think the structure I added is illogical. I don't think it will damage any part or create a bug.
In my opinion, it would be more accurate to throw an unsupported platform exception in method calls from the application instead of completely breaking the build of a platform (web) when a plugin is added.
I'll add an exception for the noop method and update it if it's ok for you. Or if you do not find this approach right, you can reject the Pull request and close it.

packages/video_player_videohole/lib/video_player.dart
```diff
@@ -446,38 +418,8 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
       }
     }
 
-    if (closedCaptionFile != null) {
-      _closedCaptionFile ??= await closedCaptionFile;
-      value = value.copyWith(caption: _getCaptionAt(value.position));
-    }
```
swift-kim, line 452: Accidentally removed?

packages/video_player_videohole/pubspec.yaml
```diff
@@ -2,13 +2,13 @@ name: video_player_videohole
...
 repository: https://github.com/flutter-tizen/plugins/tree/master/packages/video_player_videohole
-version: 0.1.1
+version: 0.1.2
 
 environment:
   sdk: ">=2.18.0 <4.0.0"
   flutter: ">=3.3.0"
 
-flutter:
+flutter:
```
swift-kim, line 11: nit: redundant space added

packages/video_player_videohole/lib/video_player.dart
MasterSon05: yes sorry

swift-kim (approved):

Issue #348, opened 2022-04-05
Avoid use of shared method channels


swift-kim:
Shared method channels (default impl. of platform interfaces) are not going to be actively maintained in flutter/plugins (although they are not being deprecated immediately), so we need to stop using them in our packages and instead, add explicit Dart implementations.

Context:
- https://docs.flutter.dev/go/platform-channels-in-federated-plugins
- https://github.com/flutter/flutter/issues/68498
- https://github.com/flutter/flutter/issues/94224

Packages that need migration:
- [x] video_player_tizen
  - This might ease migrating the package to the plusplayer-based implementation which will require more Tizen-specific functions to be added on top of the existing platform interface. (@xuelian-bai Please provide some details if you have any.)
  - May depend on https://github.com/flutter-tizen/plugins/issues/347.
- [x] image_picker_tizen
- [x] camera_tizen

Not applicable (already migrated):
- google_maps_flutter_tizen
- google_sign_in_tizen
- path_provider_tizen
- shared_preferences_tizen
- url_launcher_tizen
- webview_flutter_tizen

stuartmorgan:
> * May depend on [[pigeon] Add  C++ support #347](https://github.com/flutter-tizen/plugins/issues/347).

This is only the case if you specifically want to use Pigeon. Per-package method channels means that everything about the communication channel—including whether or not to use Pigeon—is up to each implementation package to decide, regardless of what other platform implementations do.

PR #603, opened 2023-07-26
[connectivity_plus] Update package dependency

Swanseo0:
Update connectivity_plus to 4.0.1.
Update connectivity_plus_platform_interface to 1.2.4.
Update example app.

### DIFF
--- a/packages/connectivity_plus/CHANGELOG.md
+++ b/packages/connectivity_plus/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.1.4
+
+* Update connectivity_plus to 4.0.1.
+* Update connectivity_plus_platform_interface to 1.2.4.
+* Update the example app.
+
 ## 1.1.3
 
 * Remove unnecessary `StreamHandlerError` implementation.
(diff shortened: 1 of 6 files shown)


JSUYA (approved):

JSUYA (commented):
Could you please check why we can continue to use [MethodChannelConnectivity](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus_platform_interface/lib/method_channel_connectivity.dart) without implementing [ConnectivityPlatform](https://github.com/fluttercommunity/plus_plugins/blob/main/packages/connectivity_plus/connectivity_plus_platform_interface/lib/connectivity_plus_platform_interface.dart)?
I haven't yet found a reason to use platform_interface in the example.
I may be missing something.

swift-kim:
@JSUYA Plus plugins fall back to shared method channels if no platform implementation is found.

c.f. For 1P plugins we have our own platform implementations because the [shared method channels](https://github.com/flutter-tizen/plugins/issues/348) are deprecated (or they [don't even exist](https://github.com/flutter/plugins/commit/3c7209408af83a89943d2074862f070d1665023f)). Note that this only applies to 1P plugins.

PR #606, opened 2023-07-28
[permission_handler] Update package dependency

kazakan:
* Update permission_handler to 10.4.3.
* Update permission_handler_platform_interface to 3.11.0.
* Update the example app.

### DIFF
--- a/packages/permission_handler/example/lib/main.dart
+++ b/packages/permission_handler/example/lib/main.dart
@@ -35,7 +35,7 @@ class _PermissionHandlerWidgetState extends State<PermissionHandlerWidget> {
       child: ListView(
           children: Permission.values
               .where((permission) {
-                /// Permissions not applicable for Tizen
+                // Permissions not applicable for Tizen.
                 return permission != Permission.unknown &&
                     permission != Permission.photos &&
                     permission != Permission.photosAddOnly &&
--- a/packages/permission_handler/example/lib/main.dart
+++ b/packages/permission_handler/example/lib/main.dart
@@ -1,5 +1,3 @@
-import 'dart:io';
-
 import 'package:baseflow_plugin_template/baseflow_plugin_template.dart';
 import 'package:flutter/material.dart';
 import 'package:permission_handler/permission_handler.dart';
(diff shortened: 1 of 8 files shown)


packages/permission_handler/example/lib/main.dart
```diff
@@ -50,18 +50,23 @@ class _PermissionHandlerWidgetState extends State<PermissionHandlerWidget> {
...
+                      permission != Permission.bluetoothConnect &&
+                      permission != Permission.nearbyWifiDevices &&
+                      permission != Permission.videos &&
+                      permission != Permission.audio &&
+                      permission != Permission.scheduleExactAlarm &&
+                      permission != Permission.sensorsAlways;
                 } else {
                   return permission != Permission.unknown &&
                       permission != Permission.mediaLibrary &&
-                      permission != Permission.photos &&
```
JSUYA, line 57: Could you please check this?
In README, photos is iOS-only permission. is this available in tizen?
Please check the permissions allowed by tizen platform, README, and this if condition together.(android-only also..)

kazakan: Permission.photos is for android (with version >= TIRAMISU) and IOS. [permission handler code](https://github.com/Baseflow/flutter-permission-handler/blob/dfed5d2114e4fe1ce2d51711e0a2074b612cb943/permission_handler_platform_interface/lib/src/permissions.dart#L86C1-L101C41)

But it seems there's no privilege related with `photos` in Tizen.
I'll remove `photos` in ios-only list in README, undo removal of that line, and check for other values again.

kazakan:
By the way, can we convert android-only permission `Permission.scheduleExactAlarm` to tizen privilege `http://tizen.org/privilege/alarm.set`?

- scheduleExactAlarm [Related docs](https://developer.android.com/about/versions/14/changes/schedule-exact-alarms)
>[Calendar or alarm clock apps](https://support.google.com/googleplay/android-developer/answer/12253906#exact_alarm_preview) need to send calendar reminders, wake-up alarms, or alerts when the app is no longer running. These apps can request the [USE_EXACT_ALARM](https://developer.android.com/reference/android/Manifest.permission#USE_EXACT_ALARM) normal permission. The USE_EXACT_ALARM permission will be granted on install, and apps holding this permission will be able to schedule exact alarms just like apps with the [SCHEDULE_EXACT_ALARM](https://developer.android.com/reference/android/Manifest.permission#SCHEDULE_EXACT_ALARM) permission.

- `http://tizen.org/privilege/alarm.set` [Related docs](https://www.tizen.org/privilege)
>  Allows the application to set alarms and wake the device up at scheduled times.
Type: Native / Dotnet
Level: Public
Privacy: No


But when I converted it and run example app, example app says `PermissionStatus.permanentlyDenied`

packages/permission_handler/example/lib/main.dart
kazakan: This Tizen plugin returns `kGranted` when given Permission is not convertible into tizen privilege in mobile emulator. 

[Code that converts permission into string](https://github.com/flutter-tizen/plugins/blob/b33b193d5439246ff19437436c6b89b495fae34c/packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc#L21C1-L49C2)
[Code that ignores empty string when requestPermission](https://github.com/flutter-tizen/plugins/blob/b33b193d5439246ff19437436c6b89b495fae34c/packages/permission_handler/tizen/src/permission_handler_tizen_plugin.cc#L116C1-L122C12)

So, I think, to test permission supported by test in example app, just not to put the supported permissions into that return statement is enough just like before.

But I'll keep checking there's permissions that convertible into tizen privilege.

kazakan:
> By the way, can we convert android-only permission `Permission.scheduleExactAlarm` to tizen privilege `http://tizen.org/privilege/alarm.set`?
> 
>     * scheduleExactAlarm [Related docs](https://developer.android.com/about/versions/14/changes/schedule-exact-alarms)
> 
> 
> > [Calendar or alarm clock apps](https://support.google.com/googleplay/android-developer/answer/12253906#exact_alarm_preview) need to send calendar reminders, wake-up alarms, or alerts when the app is no longer running. These apps can request the [USE_EXACT_ALARM](https://developer.android.com/reference/android/Manifest.permission#USE_EXACT_ALARM) normal permission. The USE_EXACT_ALARM permission will be granted on install, and apps holding this permission will be able to schedule exact alarms just like apps with the [SCHEDULE_EXACT_ALARM](https://developer.android.com/reference/android/Manifest.permission#SCHEDULE_EXACT_ALARM) permission.
> 
>     * `http://tizen.org/privilege/alarm.set` [Related docs](https://www.tizen.org/privilege)
> 
> 
> > Allows the application to set alarms and wake the device up at scheduled times.
> > Type: Native / Dotnet
> > Level: Public
> > Privacy: No
> 
> But when I converted it and run example app in emulator, example app says `PermissionStatus.permanentlyDenied`

After add permission into manifest file, app says `Permission.Granted` without explicit permission acquirement.

All the permissions supported in this plugin before needed explicit acquire of permission. But then we couldn't find out permissions that doesn't need explicit acquire is acquired or not. Is this intended behavior?

swift-kim:
Only privacy-related privileges are associated with specific permissions. The alarm privilege doesn't require any permission on Tizen. Please see the Tizen doc linked in the package README.

packages/permission_handler/example/lib/main.dart
kazakan: Fixed it!

kazakan: + They are ordered by enum value.

JSUYA (approved):
LGTM

PR #622, opened 2023-10-24
Fix non-platform thread issue.

hyue7:
fix issue in video_player and video_player_videohole:
```
[E] [ERROR:flutter/shell/common/shell.cc(1004)] The 'flutter.io/videoPlayer/videoEvents2' channel sent a message from native to Flutter on a non-platform thread. Platform channel messages must be sent on the platform thread. Failure to do so may result in data  loss or crashes, and must be fixed in the plugin or application code creating that channel.
See https://docs.flutter.dev/platform-integration/platform-channels#channels-and-platform-threading for  more information. [E]
```


packages/video_player/tizen/src/video_player.cc
```diff
@@ -205,6 +212,45 @@ VideoPlayer::~VideoPlayer() {
   }
 }
 
+void VideoPlayer::ExecuteSinkEvents() {
```
swift-kim, line 215: How about
```suggestion
void VideoPlayer::SendPendingEvents() {
```

packages/video_player/tizen/src/video_player.cc
```diff
@@ -205,6 +212,45 @@ VideoPlayer::~VideoPlayer() {
...
+  while (!error_event_queue_.empty()) {
+    if (event_sink_) {
+      event_sink_->Error(error_event_queue_.front().first,
+                         error_event_queue_.front().second);
+    }
+    error_event_queue_.pop();
+  }
+}
+
+void VideoPlayer::PushEvent(flutter::EncodableValue encodable_value) {
```
swift-kim, line 233: Will `std::queue::push` involve additional copying?
```suggestion
void VideoPlayer::PushEvent(const flutter::EncodableValue& encodable_value) {
```

packages/video_player/tizen/src/video_player.cc
```diff
@@ -205,6 +212,45 @@ VideoPlayer::~VideoPlayer() {
...
+  if (event_sink_ == nullptr) {
```
swift-kim, line 234: Just for consistency
```suggestion
  if (!event_sink_) {
```

swift-kim (commented):
Otherwise looks good to me. I'm afraid other plugins may have similar issues.

packages/video_player/tizen/src/video_player.cc
hyue7: Fixed!

packages/video_player/tizen/src/video_player.cc
hyue7: Will std::queue::push involve additional copying? >>>>> I think will not.
Fixed!

packages/video_player/tizen/src/video_player.cc
hyue7: Fixed!

swift-kim (approved):

xiaowei-guan (approved):

PR #624, opened 2023-10-25
[sqflite] Downgrade flutter_lints to resolve analyze issues

swift-kim:
Fixes the current [CI issue](https://github.com/flutter-tizen/plugins/actions/runs/6624690231/job/17994188145?pr=622).

- Downgrade `flutter_lints` from any (3.0.0) to ^2.0.0.
- Remove unnecessary dependency on `flutter_lints` and remove `analysis_options.yaml`.
- Sync `example/analysis_options.yaml`.


packages/sqflite/analysis_options.yaml
```diff
@@ -1,101 +0,0 @@
-include: package:flutter_lints/flutter.yaml
```
JSUYA, line 1: Frontend package maintains this file. Is there a reason why it must be deleted?
https://github.com/tekartik/sqflite/blob/master/sqflite/analysis_options.yaml

swift-kim: The file is for the [Dart code](https://github.com/tekartik/sqflite/tree/master/sqflite/lib) included in the package. There's no Dart code in our package (except `sqflite_tizen.dart`).

JSUYA (approved):

Issue #621, opened 2023-10-16
[video_player] httpHeaders support


JMaroz:
Hello,

I'm currently working on a TV app, and we need to set an HTTP header cookie when playing a video. I noticed that the [video_player_tizen](https://github.com/flutter-tizen/plugins/tree/master/packages/video_player#video_player_tizen) does not support this option.

Are you working on this feature? Do you have any ETA?

Thanks,
Andrea

swift-kim:
@xuelian-bai Do you know anything about this? This feature doesn't seem to be supported by any of the TV player APIs as far as I can see.

hyue7:
Hi, @JMaroz  
[video_player_tizen](https://github.com/flutter-tizen/plugins/tree/master/packages/video_player#video_player_tizen) and [video_player_videohole](https://github.com/flutter-tizen/plugins/tree/master/packages/video_player_videohole) not support set http header yet. Since we are now focusing on [avplyer plugin](https://github.com/flutter-tizen/plugins/pull/620), I will try to support it with this plugin. Please follow this, if there is any update on it.

By the way, i find these TV player Apis:
`player_set_streaming_cookie` with mmplayer and `SetStreamingProperty` with plusplayer
But I'm not sure if they're feasible to implement this feature, i will try.

PR #628, opened 2023-11-17
[video_player_videohole] Implement 'httpHeaders' of 'VideoPlayerController.network'.

hyue7:
Close https://github.com/flutter-tizen/plugins/issues/621


packages/video_player_videohole/tizen/src/video_player.cc
```diff
@@ -105,6 +106,34 @@ int64_t VideoPlayer::Create(const std::string &uri, int drm_type,
...
+        std::string user_agent = std::get<std::string>(iter->second);
+        ret = player_set_streaming_user_agent(player_, user_agent.c_str(),
+                                              user_agent.size());
+        if (ret != PLAYER_ERROR_NONE) {
+          LOG_ERROR("[MediaPlayer] player_set_streaming_user_agent failed: %s.",
+                    get_error_message(ret));
+        }
+      }
+    }
+  }
```
xiaowei-guan, line 181: suggest extract a new method to get value from http headers, for example:
```cpp
template <typename T>
void ExtractValueFromMap(const flutter::EncodableValue &arguments,
                         const char *key, T &outValue) {
  if (std::holds_alternative<flutter::EncodableMap>(arguments)) {
    flutter::EncodableMap values = std::get<flutter::EncodableMap>(arguments);
    flutter::EncodableValue value = values[flutter::EncodableValue(key)];
    if (std::holds_alternative<T>(value)) outValue = std::get<T>(value);
  }
}
```

packages/video_player_videohole/tizen/src/video_player_tizen_plugin.cc
```diff
@@ -145,11 +146,17 @@ ErrorOr<PlayerMessage> VideoPlayerTizenPlugin::Create(
         }
       }
     }
+
+    const flutter::EncodableMap *http_headers_map = msg.http_headers();
+    if (http_headers_map) {
+      http_headers = *http_headers_map;
```
xiaowei-guan, line 152: Copying the http_header_map is not recommended.

packages/video_player_videohole/tizen/src/video_player.cc
hyue7: fixed

packages/video_player_videohole/tizen/src/video_player_tizen_plugin.cc
hyue7: fixed

xiaowei-guan (approved):

PR #634, opened 2023-11-28
[sensor_plus] Implement set*SamplingPeriod method call

JSUYA:
* Implement set*SamplingPeriod method call.
* Update deprecated API in integration_test.
* Update sensors_plus to 4.0.1.
* Update sensors_plus_platform_interface to 1.2.0.

### DIFF
--- a/packages/sensors_plus/tizen/src/device_sensor.cc
+++ b/packages/sensors_plus/tizen/src/device_sensor.cc
@@ -24,8 +24,21 @@ sensor_type_e ToTizenSensorType(const SensorType &sensor_type) {
 
 }  // namespace
 
-DeviceSensor::DeviceSensor(SensorType sensor_type)
-    : sensor_type_(sensor_type) {}
+DeviceSensor::DeviceSensor(SensorType sensor_type) : sensor_type_(sensor_type) {
+  sensor_h sensor;
+  int ret = sensor_get_default_sensor(ToTizenSensorType(sensor_type_), &sensor);
+  if (ret != SENSOR_ERROR_NONE) {
+    LOG_ERROR("Failed to get default sensor: %s", get_error_message(ret));
+    last_error_ = ret;
+    return;
+  }
+
+  ret = sensor_create_listener(sensor, &listener_);
+  if (ret != SENSOR_ERROR_NONE) {
+    LOG_ERROR("Failed to create listener: %s", get_error_message(ret));
+    last_error_ = ret;
+  }
+}
 
 DeviceSensor::~DeviceSensor() {
   if (is_listening_) {
@@ -51,23 +64,8 @@ bool DeviceSensor::StartListen(SensorEventCallback callback) {
     return false;
   }
 
-  sensor_h sensor;
-  int ret = sensor_get_default_sensor(ToTizenSensorType(sensor_type_), &sensor);
-  if (ret != SENSOR_ERROR_NONE) {
-    LOG_ERROR("Failed to get default sensor: %s", get_error_message(ret));
-    last_error_ = ret;
-    return false;
-  }
-
-  ret = sensor_create_listener(sensor, &listener_);
-  if (ret != SENSOR_ERROR_NONE) {
-    LOG_ERROR("Failed to create listener: %s", get_error_message(ret));
-    last_error_ = ret;
-    return false;
-  }
-
-  ret = sensor_listener_set_event_cb(
-      listener_, 60,
+  int ret = sensor_listener_set_event_cb(
+      listener_, interval_ms_,
       [](sensor_h sensor, sensor_event_s *event, void *user_data) {
         auto *self = static_cast<DeviceSensor *>(user_data);
         SensorEvent sensor_event;
@@ -116,5 +114,13 @@ void DeviceSensor::StopListen() {
     LOG_ERROR("Failed to unset event callback: %s", get_error_message(ret));
     last_error_ = ret;
     return;
+  }
+}
+
+void DeviceSensor::SetInterval(int interval_ms) {
+  interval_ms_ = interval_ms;
+
+  if (listener_) {
+    sensor_listener_set_interval(listener_, interval_ms_);
   }
 }
--- a/packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
+++ b/packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
@@ -7,22 +7,27 @@
 #include <flutter/event_channel.h>
 #include <flutter/event_sink.h>
 #include <flutter/event_stream_handler_functions.h>
+#include <flutter/method_channel.h>
 #include <flutter/plugin_registrar.h>
 #include <flutter/standard_method_codec.h>
 
 #include <memory>
 
 #include "device_sensor.h"
+#include "log.h"
 
 typedef flutter::EventChannel<flutter::EncodableValue> FlEventChannel;
 typedef flutter::EventSink<flutter::EncodableValue> FlEventSink;
+typedef flutter::MethodCall<flutter::EncodableValue> FlMethodCall;
+typedef flutter::MethodChannel<flutter::EncodableValue> FlMethodChannel;
+typedef flutter::MethodResult<flutter::EncodableValue> FlMethodResult;
 typedef flutter::StreamHandler<flutter::EncodableValue> FlStreamHandler;
 typedef flutter::StreamHandlerError<flutter::EncodableValue>
     FlStreamHandlerError;
 
 class DeviceSensorStreamHandler : public FlStreamHandler {
  public:
-  DeviceSensorStreamHandler(SensorType type) : sensor_(type) {}
+  DeviceSensorStreamHandler(DeviceSensor *sensor) : sensor_(sensor) {}
 
  protected:
   std::unique_ptr<FlStreamHandlerError> OnListenInternal(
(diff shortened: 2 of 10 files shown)


packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +71,87 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+    const flutter::EncodableValue *arguments = method_call.arguments();
+
+    if (method_name == "setAccelerationSamplingPeriod" ||
+        method_name == "setGyroscopeSamplingPeriod" ||
+        method_name == "setUserAccelerometerSamplingPeriod" ||
+        method_name == "setMagnetometerSamplingPeriod") {
+      const auto *sampling_period = std::get_if<int32_t>(arguments);
+      if (sampling_period) {
+        // TODO(jsuya): In the sensor_plus_platform_interface, the
+        // set***SamplingPeriod method call is called only before creating an
```
swift-kim, line 136: Why don't you unset the event callback and set the callback again?

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +71,87 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+        // Event channel. Tizen has sensor_listener_set_interval, but it does
+        // not support setting the sampling period after creating an event
+        // channel because there is no use case. This may be supported in the
+        // future.
+        interval_ = *sampling_period / 1000;  // sampling_period is microsecond
```
swift-kim, line 141: To clarify, how about
```suggestion
        interval_ms_ = *sampling_period / 1000;  // sampling_period is in microseconds.
```

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +71,87 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+        return;
+      }
+    } else {
+      result->NotImplemented();
+      return;
+    }
+    result->Success();
+  }
+
+  int interval_ = 0;
```
swift-kim, line 154: According to the documentation, 0 means the system default interval (typically 100 ms). Can we distinguish between the default interval and `SensorInterval.fastestInterval`, which has a value of 0?

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -58,7 +62,7 @@ class SensorsPlusPlugin : public flutter::Plugin {
  public:
   static void RegisterWithRegistrar(flutter::PluginRegistrar *registrar) {
     auto plugin = std::make_unique<SensorsPlusPlugin>();
-    plugin->SetupEventChannels(registrar);
+    plugin->SetupEventChannels(registrar, plugin.get());
```
swift-kim, line 65: nit:
```suggestion
    plugin->SetUpEventChannels(registrar, plugin.get());
```

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: I didn't check the example code before writing these comments.
The example code calls each `set*SamplingPeriod` to be set separately from the event stream creation. So I added `SetInterval()`, which calls `sensor_listener_set_interval()`.

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: done

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: done

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: done

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +73,129 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+    std::unique_ptr<DeviceSensor> accelerometer_sensor =
+        std::make_unique<DeviceSensor>(SensorType::kAccelerometer,
+                                       interval_ms_);
+    accelerometer_sensor_ = accelerometer_sensor.get();
     accelerometer_channel->SetStreamHandler(
         std::make_unique<DeviceSensorStreamHandler>(
-            SensorType::kAccelerometer));
+            std::move(accelerometer_sensor)));
 
     std::unique_ptr<FlEventChannel> gyroscope_channel =
```
swift-kim, line 101: FYI) [For historical reasons](https://github.com/flutter-tizen/plugins/pull/134), event channel instances are typically owned by plugin classes (see our other plugins that use event channels). However, according to https://github.com/flutter-tizen/plugins/issues/131, this issue has already been resolved and the workaround seems no longer necessary.

That said, it's still good practice to keep the event channel instances alive during the plugin lifecycle. You can update the code accordingly if you think it's better.

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +73,129 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+    } else {
+      interval_ms_ =
+          *sampling_period / 1000;  // sampling_period is in microsecond.
+    }
+
+    return true;
   }
+
+  int interval_ms_ = 0;
+  DeviceSensor *accelerometer_sensor_ = nullptr;
```
swift-kim, line 195: I'd recommend making the sensor owned by the plugin class and passing its pointer when creating a stream handler.

```suggestion
  std::unique_ptr<DeviceSensor> accelerometer_sensor_;
```

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +73,129 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
+
+  bool SetInterval(const FlMethodCall &method_call, FlMethodResult *result) {
+    const auto *sampling_period = std::get_if<int32_t>(method_call.arguments());
+    if (!sampling_period) {
+      result->Error("Invalid argument", "No sampling period provided.");
+      return false;
+    }
+
+    if (*sampling_period == 0) {
+      interval_ms_ = 1000;
```
swift-kim, line 185: What's the reason for this? 1000 is neither the default interval nor the fastest interval.

packages/sensors_plus/tizen/src/device_sensor.cc
```diff
@@ -118,3 +118,11 @@ void DeviceSensor::StopListen() {
...
   }
 }
+
+void DeviceSensor::SetInterval(int interval_ms) {
+  interval_ms_ = interval_ms;
+
+  if (listener_) {
+    sensor_listener_set_interval(listener_, interval_ms_);
+  }
+}
```
swift-kim, line 128: Missing newline.

packages/sensors_plus/tizen/src/device_sensor.cc
```diff
@@ -67,7 +67,7 @@ bool DeviceSensor::StartListen(SensorEventCallback callback) {
   }
```
swift-kim, line 67: For consistency, we should either move the `sensor_create_listener` call (and also `sensor_get_default_sensor`) to the constructor, or destroy the listener in `DeviceSensor::StopListen`. (Personally I prefer the former.)

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: oh It was my mistake.
According to the [documentation](https://developer.tizen.org/dev-guide/tizen-iot-headed/4.0/group__CAPI__SYSTEM__SENSOR__LISTENER__MODULE.html#ga32f88b4d08ae4c373dd83648a05613f6), bound is 10 to 1000, although it may vary for each sensor.
I will fix it to 10.

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: thanks done

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: done

packages/sensors_plus/tizen/src/device_sensor.cc
JSUYA: done

packages/sensors_plus/tizen/src/device_sensor.cc
JSUYA: done

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
```diff
@@ -67,40 +72,113 @@ class SensorsPlusPlugin : public flutter::Plugin {
...
-  void SetupEventChannels(flutter::PluginRegistrar *registrar) {
-    std::unique_ptr<FlEventChannel> accelerometer_channel =
-        std::make_unique<FlEventChannel>(
-            registrar->messenger(),
-            "dev.fluttercommunity.plus/sensors/accelerometer",
+  void SetUpEventChannels(flutter::PluginRegistrar *registrar,
+                          SensorsPlusPlugin *plugin) {
+    std::unique_ptr<FlMethodChannel> method_channel =
+        std::make_unique<FlMethodChannel>(
+            registrar->messenger(), "dev.fluttercommunity.plus/sensors/method",
```
swift-kim, line 79: nit: This method (`SetUpEventChannels`) is a place for initializing event channels. Initializing the method channel could be done outside (in `RegisterWithRegistrar`).

swift-kim (approved):
Otherwise looks good to me. Thanks!

packages/sensors_plus/tizen/src/sensors_plus_plugin.cc
JSUYA: Thanks !

PR #636, opened 2023-12-05
[video_player_avplay] Replace surface id with resource id when SetDisplay

xiaowei-guan:
SetDisplay
```bash
   * @param     [in] type : display type
   * @param     [in] serface_id : resource id of window.
   * @param     [in] x : x param of display window
   * @param     [in] y : y param of display window
   * @param     [in] w : width of display window. This value should be greater than 0.
   * @param     [in] h : height of display window. This value should be greater than 0.
```
Actually the surface id is the resource id of window.

### DIFF
--- a/packages/video_player_avplay/tizen/src/video_player_tizen_plugin.cc
+++ b/packages/video_player_avplay/tizen/src/video_player_tizen_plugin.cc
@@ -104,17 +104,6 @@ std::optional<FlutterError> VideoPlayerTizenPlugin::Initialize() {
 
 ErrorOr<PlayerMessage> VideoPlayerTizenPlugin::Create(
     const CreateMessage &msg) {
-  FlutterDesktopViewRef flutter_view =
-      FlutterDesktopPluginRegistrarGetView(registrar_ref_);
-  if (!flutter_view) {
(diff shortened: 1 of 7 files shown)


packages/video_player_avplay/tizen/src/video_player_tizen_plugin.cc
```diff
@@ -104,17 +104,6 @@ std::optional<FlutterError> VideoPlayerTizenPlugin::Initialize() {
 
 ErrorOr<PlayerMessage> VideoPlayerTizenPlugin::Create(
     const CreateMessage &msg) {
-  FlutterDesktopViewRef flutter_view =
-      FlutterDesktopPluginRegistrarGetView(registrar_ref_);
-  if (!flutter_view) {
-    return FlutterError("Operation failed", "Could not get a Flutter view.");
-  }
```
JSUYA, line 111: nit: This null check is missing.

JSUYA (approved):

swift-kim (approved):

packages/video_player_avplay/tizen/src/video_player_tizen_plugin.cc
xiaowei-guan: Fixed

Issue #637, opened 2023-12-11
[tizen_bundle] Bundle creation issue


JSUYA:
When creating a bundle, an assertion is occured in calling _finaliser.attach().
(https://github.com/flutter-tizen/plugins/actions/runs/7139685481/job/19497841519)
```
2023-12-11T02:19:35.3050863Z ##[group]❌ can create a bundle (failed)
2023-12-11T02:19:35.3052247Z ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
2023-12-11T02:19:35.3053538Z The following assertion was thrown running a test:
2023-12-11T02:19:35.3054501Z The token should not be the value being attached
2023-12-11T02:19:35.3055404Z 'dart:core-patch/finalizer_patch.dart':
2023-12-11T02:19:35.3056592Z Failed assertion: line 41 pos 12: '<optimized out>'
2023-12-11T02:19:35.3057138Z 
2023-12-11T02:19:35.3057449Z When the exception was thrown, this was the stack:
2023-12-11T02:19:35.3058845Z #2      _FinalizerImpl.attach (dart:core-patch/finalizer_patch.dart:41:12)
2023-12-11T02:19:35.3059766Z #3      new Bundle (package:tizen_bundle/tizen_bundle.dart:23:16)
2023-12-11T02:19:35.3061366Z #4      main.<anonymous closure> (file:///home/tizen/runners/runner1/_work/plugins/plugins/packages/tizen_bundle/example/integration_test/bundle_test.dart:13:27)
2023-12-11T02:19:35.3063714Z #5      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:168:29)
2023-12-11T02:19:35.3064803Z <asynchronous suspension>
2023-12-11T02:19:35.3065340Z #6      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1013:5)
2023-12-11T02:19:35.3066356Z <asynchronous suspension>
2023-12-11T02:19:35.3066645Z <asynchronous suspension>
2023-12-11T02:19:35.3067269Z (elided 3 frames from class _AssertionError and package:stack_trace)
```

I guess it's probably related to this issue.

https://github.com/dart-lang/sdk/commit/ea8485e967ecd1ac796832161ba853aee4eb085e

swift-kim:
[...]

PR #638, opened 2023-12-11
[tizen_bundle, tizen_rpc_port] Fix token in finalizer to be different from value

JSUYA:
related issue : https://github.com/flutter-tizen/plugins/issues/637

### DIFF
--- a/packages/tizen_rpc_port/lib/src/stub_base.dart
+++ b/packages/tizen_rpc_port/lib/src/stub_base.dart
@@ -33,14 +33,13 @@ abstract class StubBase {
       return pStub.value;
     });
 
-    _finalizer.attach(this, this);
+    _finalizer.attach(this, _handle);
   }
 
   late final rpc_port_stub_h _handle;
 
-  final Finalizer<StubBase> _finalizer = Finalizer<StubBase>((StubBase stub) {
-    stub.close();
-  });
+  final Finalizer<rpc_port_stub_h> _finalizer = Finalizer<rpc_port_stub_h>(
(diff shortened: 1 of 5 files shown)


packages/tizen_rpc_port/lib/src/stub_base.dart
```diff
@@ -33,13 +33,15 @@ abstract class StubBase {
       return pStub.value;
     });
 
-    _finalizer.attach(this, this);
+    _finalizer.attach(this, <String, StubBase>{'stub': this});
```
swift-kim, line 36: I think the stream subscription should already be closed when when the attachment (`StubBase`) is destroyed, so this should just be safe:

```patch
@@ -33,14 +33,13 @@ abstract class StubBase {
       return pStub.value;
     });

-    _finalizer.attach(this, this);
+    _finalizer.attach(this, _handle);
   }

   late final rpc_port_stub_h _handle;

-  final Finalizer<StubBase> _finalizer = Finalizer<StubBase>((StubBase stub) {
-    stub.close();
-  });
+  final Finalizer<rpc_port_stub_h> _finalizer = Finalizer<rpc_port_stub_h>(
+      (rpc_port_stub_h handle) => tizen.rpc_port_stub_destroy(handle));

   /// A port name to use when listening for connections.
```

JSUYA: Thank you for the advice. I thought it was the same case for ProxyBase, so I fixed both.

swift-kim (approved):

Issue #640, opened 2023-12-21
[ci] The release workflow fails


swift-kim:
https://github.com/flutter-tizen/plugins/actions/runs/7260865311/job/19780899340
```
▶ Extracting Ruby
  /usr/bin/tar -xz -C /opt/hostedtoolcache/Ruby/2.7.4 -f /home/runner/work/_temp/ef458c6f-a467-48b1-bdae-38624bc4f00d
  Took   0.27 seconds
▶ Installing Bundler
  /opt/hostedtoolcache/Ruby/2.7.4/x64/bin/gem install bundler -v ~> 2
  ERROR:  Error installing bundler:
  	The last version of bundler (~> 2) to support your Ruby & RubyGems was 2.4.22. Try installing it with `gem install bundler -v 2.4.22`
  	bundler requires Ruby version >= 3.0.0. The current ruby version is 2.7.4.191.
  Took  12.96 seconds
Error: The process '/opt/hostedtoolcache/Ruby/2.7.4/x64/bin/gem' failed with exit code 1
```

Updating `lewagon/wait-on-check-action` to the latest (1.3.1) doesn't seem to solve the issue. See https://github.com/lewagon/wait-on-check-action/issues/98.

JSUYA:
Thank you for check. I just published it manually. and I will check it too.

PR #641, opened 2023-12-21
[ci] Update wait-on-check-action version to 1.3.3

JSUYA:
bundler install issue that occurred in wait-on-check-action has been fixed.

https://github.com/flutter-tizen/plugins/issues/640

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
             --base-sha=HEAD~
       - name: Wait on all tests
         if: ${{ github.event_name == 'push' }}
-        uses: lewagon/wait-on-check-action@v1.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             --base-sha=HEAD~
       - name: Wait on all tests
         if: ${{ github.event_name == 'push' }}
-        uses: lewagon/wait-on-check-action@v1.1.1
+        uses: lewagon/wait-on-check-action@v1.3.2
```
swift-kim, line 29: The latest version is 1.3.3.

JSUYA: I updated it

swift-kim (approved):

PR #643, opened 2023-12-22
[video_player_videohole] Add get duration API for live stream

xiaowei-guan:
Live stream duration is different from common video duration.

1. Live stream duration is a range, so we change duration variable from Duration to DurationRange, the  DurationRange which include a start time and a end time.
2. The duration of the live stream changes during playback, so need add a timer to get duration.


packages/video_player_videohole/tizen/src/video_player_tizen_plugin.cc
```diff
@@ -177,6 +178,21 @@ ErrorOr<PlayerMessage> VideoPlayerTizenPlugin::Create(
   return result;
 }
 
+ErrorOr<DurationMessage> VideoPlayerTizenPlugin::Duration(
+    const PlayerMessage &msg) {
+  VideoPlayer *player = FindPlayerById(msg.player_id());
+  if (!player) {
+    return FlutterError("Invalid argument", "Player not found");
```
JSUYA, line 185:
```suggestion
    return FlutterError("Invalid argument", "Player not found.");
```

packages/video_player_videohole/lib/video_player_platform_interface.dart
```diff
@@ -271,7 +276,7 @@ class VideoEvent {
   /// Duration of the video.
   ///
   /// Only used if [eventType] is [VideoEventType.initialized].
-  final Duration? duration;
+  final DurationRange? duration;
```
JSUYA, line 279: https://github.com/flutter/packages/blob/main/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart#L229
Can't we just keep `Duration` type and use the value of `duration.second` returned by mediaplayer?
The interface that returns `DurationRange` for stream play seems to be `buffered`. (https://github.com/flutter/packages/blob/main/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart#L244)
what do you think?

xiaowei-guan: For live stream, the duration is a range which includes start time and end time.
```C++
 player_get_adaptive_streaming_info(player_, (void *)&live_duration_buff, PLAYER_ADAPTIVE_INFO_LIVE_DURATION);
```
When call seekTo, the time must be in the range， the start time is only used to judge. If remove start time, we might have seekTo issue.
I have tested seekTo for live stream, if we get live stream duration is [300, 500] and then call seekTo(200),  the player just start with 300.

packages/video_player_videohole/tizen/src/video_player_tizen_plugin.cc
xiaowei-guan: Fixed

packages/video_player_videohole/lib/video_player_platform_interface.dart
JSUYA: My worry is the change in interface. (I know, we have a separate interface from video_player_platform_interface.)

In my opinion..
When a user needs a seekTo point in valid range of streaming content, they can check the valid seekTo by taking the value from `buffered` rather than the duration value.
https://github.com/flutter/packages/blob/36a7b99381f85e86914e82c75fc7d9038ed96cca/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart#L241-L244
```dart
  /// Buffered parts of the video.
  ///
  /// Only used if [eventType] is [VideoEventType.bufferingUpdate].
  final List<DurationRange>? buffered;
```
Our `int buffered` is progress value of buffering. so we can change this API.
Of course, we don't necessarily have to follow the video_player_platform_interface, so this can be reconsidered later.

JSUYA (approved):

hyue7 (approved):

PR #644, opened 2023-12-22
[video_player_avplay] Fix not display video issue when start play

hyue7:
Issue: there is no video and audio after start play.
1. player should call `SetDisplayVisible` to display one fram after prepare is done.
2. plusplayer don't need to set playback speed when start play, otherwise it will clear buffer and restart buffer.
3. plusplayer can set playback speed when player is pause. Remove code:
```
if (!value.isPlaying) {
       return;
    }
```


xiaowei-guan (approved):
LGTM

swift-kim (approved):

PR #646, opened 2023-12-27
[video_player_videohole] Implement message queue for native license request

xiaowei-guan:
1. Create a new thread for native license request.
2. Spearate channel and native code from drm manager.


xiaowei-guan:
For now, we focus on video_player_avplay, we will move the code to video_player_avplay.

PR #651, opened 2024-01-04
[video_player_videohole] Free response buffer after install key

xiaowei-guan:


packages/video_player_videohole/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: video_player_videohole
 description: Flutter plugin for displaying inline video on Tizen TV devices.
 homepage: https://github.com/flutter-tizen/plugins
 repository: https://github.com/flutter-tizen/plugins/tree/master/packages/video_player_videohole
-version: 0.5.0
+version: 0.6.0
```
swift-kim, line 5: You can just increase the patch version if no new API has been added.

xiaowei-guan: Ok

swift-kim (approved):

PR #654, opened 2024-01-08
[video_player_avplay] Replace g_idle_add with ecore_pipe

xiaowei-guan:
1. Use the same type of interface when send event to main loop. 
2. Destory event channel first when destory player.


packages/video_player_avplay/tizen/src/drm_manager.cc
```diff
@@ -34,11 +34,19 @@ DrmManager::DrmManager() : drm_type_(DM_TYPE_NONE) {
...
+      },
+      this);
 }
 
 DrmManager::~DrmManager() {
   ReleaseDrmSession();
-
+  if (license_request_pipe_) {
+    ecore_pipe_del(license_request_pipe_);
+  }
```
hyue7, line 49:
```suggestion
    ecore_pipe_del(license_request_pipe_);
    license_request_pipe_ = nullptr;
  }
```

packages/video_player_avplay/tizen/src/media_player.cc
```diff
@@ -52,7 +53,22 @@ MediaPlayer::MediaPlayer(flutter::BinaryMessenger *messenger,
...
+    player_stop(player_);
+    player_unprepare(player_);
+    player_unset_buffering_cb(player_);
+    player_unset_completed_cb(player_);
+    player_unset_interrupted_cb(player_);
+    player_unset_error_cb(player_);
+    player_unset_subtitle_updated_cb(player_);
+    player_destroy(player_);
+    player_ = nullptr;
+  }
```
JSUYA, line 67: How about making `PlusPlayer`'s destructor code style and error log the same?
(`drm_manager_` and `player_` seem to have the same life cycle, but the processing results vary if player_ is null in `MediaPlayer` class.)

JSUYA (approved):

packages/video_player_avplay/tizen/src/media_player.cc
xiaowei-guan: Fixed

packages/video_player_avplay/tizen/src/drm_manager.cc
xiaowei-guan: Fixed

PR #655, opened 2024-01-12
[webview_flutter] Add ewk_set_version_policy() call

JSUYA:
If EWK, the web engine of webview flutter, has an upgrade as a UWE feature,
`ewk_set_version_policy(1)` must be called before set arguments.

 +) Add missing ewk_shutdown().


JSUYA:
Hi @swift-kim Could you please review this?

swift-kim (approved):
Thanks!

PR #659, opened 2024-01-17

xiaowei-guan:
Update plusplayer to latest :
* [dash] DASH unique event including index tag.
* [hls] Fix network set time bug(SSL Verification).
* [hls] Fix audio language to set in caps even in case of hls having mp3 audio. also done for other audio codecs.
* [http] Improved mmhttpsrc logging.
* [libav] Support VVC in ffmpeg side.
* [plusplayer] Stop feeder before renderer stop.
* [plusplayer] Update max resolution selection logic for multiview (portrait mode).
* [plusplayer] Fix set playing task crash.
* [plusplayer] Send Eos on Stop if trackrenderer prepare stuck.
* [plusplayer] Handling race condition between onmultiview cb and changesource.
* [plusplayer] Tracksource Seek only when Pause return success.
* [VVC/H.266] Add framerate force correction.
* [VVC/H.266] Bitrate change support (no drm).


packages/video_player_avplay/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.3
+
+* Update plusplayer to latest.
```
JSUYA, line 3: I don't know about the plusplayer package, but
Can you write a little more detail about the latest version?
version? latest commit information?

JSUYA (approved):

packages/video_player_avplay/CHANGELOG.md
xiaowei-guan: I have added dedail in changelog.

hyue7 (commented):
LGTM

PR #660, opened 2024-01-22
[video_player_avplay] Resolve drm manager proxy doesn't support multiple instances issue.

xiaowei-guan:
https://github.com/flutter-tizen/plugins/blob/master/packages/video_player_avplay/tizen/src/drm_manager_proxy.cc#L9-L14
```cpp
FuncDMGRSetData DMGRSetData = nullptr;
FuncDMGRGetData DMGRGetData = nullptr;
FuncDMGRSetDRMLocalMode DMGRSetDRMLocalMode = nullptr;
FuncDMGRCreateDRMSession DMGRCreateDRMSession = nullptr;
FuncDMGRSecurityInitCompleteCB DMGRSecurityInitCompleteCB = nullptr;
FuncDMGRReleaseDRMSession DMGRReleaseDRMSession = nullptr;
```
- If we create multiple Drm Manager, the new function pointer will overwrite the old ones, so we can extract a new class for drm manager proxy.
- Do not need dclose drm manager handler right away, which can cause coredump issue.


xiaowei-guan:
Maybe we can change the drm manager proxy to a singleton.

packages/video_player_avplay/tizen/src/drm_manager_proxy.cc
```diff
@@ -6,59 +6,103 @@
...
-
-void* OpenDrmManagerProxy() { return dlopen("libdrmmanager.so.0", RTLD_LAZY); }
-
-int InitDrmManagerProxy(void* handle) {
-  if (!handle) {
-    return DM_ERROR_INVALID_PARAM;
+#include "log.h"
+
+DrmManagerProxy::DrmManagerProxy()
+    : drm_manager_handle_(::dlopen("libdrmmanager.so.0", RTLD_LAZY)) {
```
JSUYA, line 12: :: doesn't seem to be necessary.
```suggestion
    : drm_manager_handle_(dlopen("libdrmmanager.so.0", RTLD_LAZY)) {
```

packages/video_player_avplay/tizen/src/drm_manager_proxy.cc
```diff
@@ -6,59 +6,103 @@
...
   }
 
-  DMGRSetDRMLocalMode = reinterpret_cast<FuncDMGRSetDRMLocalMode>(
-      dlsym(handle, "DMGRSetDRMLocalMode"));
-  if (!DMGRSetDRMLocalMode) {
-    return DM_ERROR_DL;
+  dmgr_set_drm_local_mode_ = reinterpret_cast<FuncDMGRSetDRMLocalMode>(
+      dlsym(drm_manager_handle_, "DMGRSetDRMLocalMode"));
+  if (dmgr_set_drm_local_mode_ == nullptr) {
+    LOG_ERROR("Fail to find DMGRSetDRMLocalMode.");
```
JSUYA, line 33: Since all of these symbols are necessary, I think it is better to return immediately if any of them fail to load.

JSUYA (approved):

packages/video_player_avplay/tizen/src/drm_manager_proxy.cc
xiaowei-guan: Ok

packages/video_player_avplay/tizen/src/drm_manager_proxy.cc
xiaowei-guan: Yes, good suggestion. I add a new method to obtaion address of drm manager functions.

packages/video_player_avplay/tizen/src/drm_manager_proxy.h
```diff
@@ -107,15 +107,39 @@ typedef bool (*FuncDMGRSecurityInitCompleteCB)(int* drm_handle,
...
+  ~DrmManagerProxy();
+  bool Initialize();
+  void* drm_manager_handle_ = nullptr;
+  FuncDMGRSetData dmgr_set_data_ = nullptr;
+  FuncDMGRGetData dmgr_get_data_ = nullptr;
+  FuncDMGRSetDRMLocalMode dmgr_set_drm_local_mode_ = nullptr;
+  FuncDMGRCreateDRMSession dmgr_create_drm_session_ = nullptr;
+  FuncDMGRSecurityInitCompleteCB dmgr_security_init_complete_cb_ = nullptr;
+  FuncDMGRReleaseDRMSession dmgr_release_drm_session_ = nullptr;
+  bool isValid = false;
```
JSUYA, line 142: typo :)
```suggestion
  bool is_valid_ = false;
```

JSUYA (approved):

packages/video_player_avplay/tizen/src/drm_manager_proxy.h
xiaowei-guan: ok

PR #662, opened 2024-02-01
[video_player_avplay] Fix state check in SetVolume

JSUYA:
SetVolume()'s state check returns fail even when the state is kPaused.
And _applyVolume called in VideoPlayerController's initialize() is called when the state is kReady, so modify the state check.

### DIFF
--- a/packages/video_player_avplay/tizen/src/plus_player.cc
+++ b/packages/video_player_avplay/tizen/src/plus_player.cc
@@ -267,8 +267,7 @@ bool PlusPlayer::SetLooping(bool is_looping) {
 bool PlusPlayer::SetVolume(double volume) {
   LOG_INFO("[PlusPlayer] Volume: %f", volume);
 
-  if (GetState(player_) != plusplayer::State::kPlaying ||
-      GetState(player_) != plusplayer::State::kPaused) {
+  if (GetState(player_) < plusplayer::State::kPlaying) {
     LOG_ERROR("[PlusPlayer] Player is in invalid state");
     return false;
   }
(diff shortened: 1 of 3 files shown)


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -267,8 +267,7 @@ bool PlusPlayer::SetLooping(bool is_looping) {
 bool PlusPlayer::SetVolume(double volume) {
   LOG_INFO("[PlusPlayer] Volume: %f", volume);
 
-  if (GetState(player_) != plusplayer::State::kPlaying ||
-      GetState(player_) != plusplayer::State::kPaused) {
+  if (GetState(player_) < plusplayer::State::kReady) {
```
xiaowei-guan, line 270: I have checked the API doc:
```c++
  /** 
  * @brief  Set audio volume level
  * @pre 	  The player state can be  #State::kPlaying or #State::kPaused
  * @param [in] volume  range  [0,100]
  * @return If set action excuted sucessfully 
  */
```
I think we can remove this code when player initizlized:
```dart
        case VideoEventType.initialized:
          value = value.copyWith(
            duration: event.duration,
            size: event.size,
            isInitialized: event.duration != null,
            errorDescription: null,
          );
          initializingCompleter.complete(null);
          _applyLooping();
          //_applyVolume();
          _applyPlayPause();
          _durationTimer?.cancel();
          _durationTimer = _createDurationTimer();
```

JSUYA: Thank you for check.
I modified code

xiaowei-guan (approved):

PR #666, opened 2024-02-01
[video_player_avplay] Convert volume range to [0,100] for plusplayer

xiaowei-guan:
Dart api volume range[0,1], plusplaer volume range[0,100]


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -266,14 +266,10 @@ bool PlusPlayer::SetLooping(bool is_looping) {
 }
 
 bool PlusPlayer::SetVolume(double volume) {
-  LOG_INFO("[PlusPlayer] Volume: %f", volume);
-
-  if (GetState(player_) < plusplayer::State::kPlaying) {
-    LOG_ERROR("[PlusPlayer] Player is in invalid state");
-    return false;
-  }
```
JSUYA, line 274: Is there a reason you removed the state check?

JSUYA (commented):
Even after applying this PR, I still get `[PlusPlayer] Fail to set volume.` in my environment.
Is your test okay?

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: I think we don't need state check, because if state is not right, it will return false when call setVolume.

xiaowei-guan: oh, I check the plusplayer source code, only dashplayer support set volume.
you can set vlume by below dash stream:
```dart
+        'https://dash.akamaized.net/dash264/TestCasesUHD/2b/11/MultiRate.mpd',
+        formatHint: VideoFormat.dash);
```

JSUYA: > I think we don't need state check, because if state is not right, it will return false when call setVolume.

However, since we can't know the reason for setVolume failure and I think it is good to check the state according to the document because it looks consistent with other API implementations.

> oh, I check the plusplayer source code, only dashplayer support set volume.

Thank you for check. I checked with dash video.

xiaowei-guan: Ok, I have rolled back the state check code.
And I have implemented the code to set the volume of all streams.

JSUYA (approved):

Issue #647, opened 2023-12-28
[webview_flutter_tizen] crash when run on real TV device


Stupidman22:
I'm running the example https://github.com/flutter-tizen/plugins/blob/master/packages/webview_flutter/example/lib/main.dart 
in TV emulator then it work normal.
When using the code to run in real samsung TV device (tizen 6.0)
then it crash and can not load webview:
[E] [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(Invalid operation, The webview instance has not been initialized., null, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:651:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:322:18)
Any one help me figure out this issue pls

Stupidman22:
If run in emulator then can not play video in any page... Any one face this issue :(

JSUYA:
Hi

> [E] [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(Invalid operation, The webview instance has not been initialized., null, null)

It's not common case to fail to create an internal WebView instance. Does this happen every time?
We need more information.

> If run in emulator then can not play video in any page... Any one face this issue :(

What kind of video are you playing? The example is blocking youtube.com (for testing). Also, YouTube video cannot be played on the emulator.
https://github.com/flutter-tizen/plugins/blob/0d363f0e89df764f57ec3f06f7fad32492e5a191/packages/webview_flutter/example/lib/main.dart#L145

If you need to play YouTube on your device, please change the `backgroundColor` of the example.
https://github.com/flutter-tizen/plugins/blob/master/packages/webview_flutter/README.md#note

JSUYA:
Hi
webview_flutter_tizen was written based on Tizen's ewk webview.
ewk webview does not support youtube play in the emulator.
Here's a similar case. youtube_player_iframe (researched it before...) is using webview internally and has the same problem.
https://github.com/flutter-tizen/plugins/issues/517

Stupidman22:
Hi bro, Now i tried to work with real TV devices:
1. I'm already connect and install test app with this device
![sdbdevice](https://github.com/flutter-tizen/plugins/assets/8415479/203274a6-8d84-4bb5-ad5d-f2e41beb0c56)
2. My devices capability is:
[...]

PR #670, opened 2024-02-16
[webview_flutter_tizen] Add tizenEnginePolicy extension api temporary

JSUYA:
There was a case where the ewk engine failed to load on some devices using Tizen 6.0.
As a result, ewk init failed and an engine instance could not be created. We have not found an exact cause for this.
Calling ewk_set_version_policy(1) will solve the problem,
but using UWE(upgrade web engine) in 3rd party application is not recommended.
(The webview works may not be guaranteed to normally.)
Therefore, add a temporary API and have user application call it. 
This API may be deleted in the future depending on whether tizen version 6.0 is supported.

related issue: https://github.com/flutter-tizen/plugins/issues/647

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,8 +1,9 @@
-## NEXT
+## 0.9.1
 
 * Add ewk_set_version_policy() call.
 * Fix new lint warnings.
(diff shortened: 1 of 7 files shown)


bbrto21 (approved):
Looks good to me.

xiaowei-guan (approved):

PR #674, opened 2024-03-04
[video_player_avplay] Remove duplicate code for checking player

xiaowei-guan:
1. Extract a new Macro Definition to check  if the player exists.
```C++
#define CHECK_PLAYER(statement)                                    \
  {                                                                \
    VideoPlayer *player = (statement);                             \
    if (!player) {                                                 \
      return FlutterError("Invalid argument", "Player not found"); \
    }                                                              \
  }

```
2. Extract a new macro definition to check the result of player
```C++
#define CHECK_PLAYER_RESULT(statement, return_value)                           \
  {                                                                            \
    const auto result = (statement);                                           \
    if (result != PLAYER_ERROR_NONE) {                                         \
      LOG_ERROR("[MediaPlayer] Player error : %s", get_error_message(result)); \
      return return_value;                                                     \
    }                                                                          \
  }

#define CHECK_PLAYER_RESULT_NO_RETURN(statement)                               \
  {                                                                            \
    const auto result = (statement);                                           \
    if (result != PLAYER_ERROR_NONE) {                                         \
      LOG_ERROR("[MediaPlayer] Player error : %s", get_error_message(result)); \
    }                                                                          \
  }

```

### DIFF
--- a/packages/video_player_avplay/CHANGELOG.md
+++ b/packages/video_player_avplay/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## NEXT
 
 * Minor refactor.
+* Remove duplicate code for checking player.
 
 ## 0.3.3
 
--- a/packages/video_player_avplay/tizen/src/media_player.cc
+++ b/packages/video_player_avplay/tizen/src/media_player.cc
@@ -10,6 +10,23 @@
 
 #include "log.h"
 
+#define CHECK_PLAYER_RESULT(statement, return_value)                           \
+  {                                                                            \
+    const auto result = (statement);                                           \
+    if (result != PLAYER_ERROR_NONE) {                                         \
+      LOG_ERROR("[MediaPlayer] Player error : %s", get_error_message(result)); \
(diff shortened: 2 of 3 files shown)


JSUYA (commented):
Hi I agree with your intention to reduce duplicate code.
However, we don't generally use the macro method.(we follow Google c++ style guide.)
https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros
If there is no suitable way, you can rename the macro (add project name + avoid conflict) and add `#undef`.
But I recommend changing it to an appropriate function.

xiaowei-guan:
> Hi I agree with your intention to reduce duplicate code. However, we don't generally use the macro method.(we follow Google c++ style guide.) https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros If there is no suitable way, you can rename the macro (add project name + avoid conflict) and add `#undef`. But I recommend changing it to an appropriate function.

Ok, I'll find a batter solution.

PR #675, opened 2024-03-06
Sync with analysis_options.yaml in flutter/packages

JSUYA:
Some packages require package upgrade rather than simple modification (ex: google_maps_flutter)
Therefore, the rules below are temporarily ignored.
Any necessary modifications can be applied in a later PR.

https://github.com/flutter/packages/pull/4060
- unnecessary_null_comparison

https://github.com/flutter/packages/pull/4067
- unawaited_futures

https://github.com/flutter/packages/pull/5717
- dangling_library_doc_comments
- no_literal_bool_comparisons
- unnecessary_library_directive
- use_colored_box
- use_enums
- use_string_in_part_of_directives

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,16 +1,22 @@
-# This file is a copy of analysis_options.yaml from flutter/packages repo
-# as of 2023-03-07.
+# Specify analysis options.
+#
+# This file is a copy of analysis_options.yaml from flutter repo
+# as of 2023-12-18, but with some modifications marked with
+# "DIFFERENT FROM FLUTTER/FLUTTER" below. The file is expected to
+# be kept in sync with the master file from the flutter repo.
 
 analyzer:
   language:
     strict-casts: true
+    strict-inference: true
     strict-raw-types: true
   errors:
-    # allow self-reference to deprecated members (we do this because otherwise we have
-    # to annotate every member in every test, assert, etc, when we deprecate something)
+    # allow deprecated members (we do this because otherwise we have to annotate
+    # every member in every test, assert, etc, when we or the Dart SDK deprecates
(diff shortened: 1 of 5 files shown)


packages/messageport/example/integration_test/messageport_test.dart
```diff
@@ -26,7 +26,7 @@ void main() {
 
   testWidgets('Create remote port', (WidgetTester tester) async {
     final LocalPort localPort = await LocalPort.create(kTestPort);
-    localPort.register((dynamic message, [RemotePort? remotePort]) => null);
+    localPort.register((dynamic message, [RemotePort? remotePort]) => () {});
```
swift-kim, line 29:
```suggestion
    localPort.register((dynamic message, [RemotePort? remotePort]) {});
```

analysis_options.yaml
```diff
@@ -1,16 +1,22 @@
-# This file is a copy of analysis_options.yaml from flutter/packages repo
-# as of 2023-03-07.
+# Specify analysis options.
+#
+# This file is a copy of analysis_options.yaml from flutter repo
+# as of 2023-12-18, but with some modifications marked with
+# "DIFFERENT FROM FLUTTER/FLUTTER" below. The file is expected to
+# be kept in sync with the master file from the flutter repo.
```
swift-kim, line 6: You may revert this change and just replace `2023-03-07` with `2024-03-07`.

JSUYA: Thank you for your review. I modified it.

packages/messageport/example/integration_test/messageport_test.dart
JSUYA: done

swift-kim (approved):

PR #679, opened 2024-04-26

xiaowei-guan:
Main changes:
- [Dash] Fix no EOS event of dashplayer.


JSUYA (approved):
Please support 8.0 libs in the next update.

PR #683, opened 2024-05-22

xiaowei-guan:
Main changes:
-   [libav-common] Fixing libav-common SVACE issue.
-   [Dash] Fix when app freeze , cpu usage is high.
-   [Dash] Fix issue that report two eos msg to app.
-   [Dash] Fix no EOS event of dashplayer.
-   [Dash] Fix app freeze issue when calling seek method.
-   [Dash] Fix issue that live stream start at the beginning not at live postion.


JSUYA (approved):
I tested it, the freezing problem looks like resolved.

xiaowei-guan:
> I tested it, the freezing problem looks like resolved.

Yes, this issue has been fixed in this PR

PR #686, opened 2024-05-23
Change the API version of example apps to 6.0

JSUYA:
Since the minimum supported version of flutter-tizen has changed from 5.5 to 6.0,
change the tizen version of the example to 6.0.
And, except for wearable_rotary, change the profile_name of packages whose it was `wearable` to `common`.

PR #688, opened 2024-05-25
[ci] Modify tizen package installation

JSUYA:
tizen_rpc_port requires tizen 6.5, and TV emulator uses Mobile 6.5 rootstrap.
Therefore, MOBILE package, not IOT-Headed, is required.

Part of the fixes missing in https://github.com/flutter-tizen/plugins/pull/686.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
           $HOME/tizen-studio/package-manager/package-manager-cli.bin install \
             NativeToolchain-Gcc-9.2 \
             IOT-Headed-6.0-NativeAppDevelopment-CLI \
-            IOT-Headed-6.5-NativeAppDevelopment-CLI
+            MOBILE-6.5-NativeAppDevelopment-CLI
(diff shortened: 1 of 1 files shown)


swift-kim:
We should either update the `api-version` of wearable_rotary to 6.0, or exclude wearable_rotary from the build workflow. The `api-version` of the package is still 5.5.

https://github.com/flutter-tizen/plugins/blob/master/packages/wearable_rotary/example/tizen/tizen-manifest.xml

JSUYA:
> We should either update the `api-version` of wearable_rotary to 6.0, or exclude wearable_rotary from the build workflow. The `api-version` of the package is still 5.5.
> 
> https://github.com/flutter-tizen/plugins/blob/master/packages/wearable_rotary/example/tizen/tizen-manifest.xml

Yes I didn't change the api-version of wearable_rotary to 6.0 because it would be confusing.
Since the build will not be processed unless the wearable_rotary package is modified, I think it is okay for now. (It was also excluded from the recipe.)

> exclude wearable_rotary from the build workflow.

I will check about this method.

PR #693, opened 2024-06-18
[webview_flutter_lwe] Supports multiple JavaScriptChannel method call

JSUYA:
Previously, javascript channel methods that are called before the webview is created are stored in the pending list.(based on name) Because of that, there was a problem with channels being missing when adding multiple channels.

When the javascript channel method is called,
it is stored differently depending on the name to prevent it from being missed.

```dart
controller
..addJavaScriptChannel(
        'method1',...)
..addJavaScriptChannel(
        'method2',...);
```

### DIFF
--- a/packages/webview_flutter_lwe/lib/src/lwe_webview.dart
+++ b/packages/webview_flutter_lwe/lib/src/lwe_webview.dart
@@ -48,7 +48,13 @@ class LweWebView {
 
   Future<T?> _invokeChannelMethod<T>(String method, [dynamic arguments]) async {
     if (!_isCreated) {
-      _pendingMethodCalls[method] = arguments;
+      if (method == 'addJavaScriptChannel' ||
+          method == 'runJavaScript' ||
+          method == 'runJavaScriptReturningResult') {
+        _pendingMethodCalls['${method}_$arguments'] = arguments;
+      } else {
+        _pendingMethodCalls[method] = arguments;
+      }
       return null;
     }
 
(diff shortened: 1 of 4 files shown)


packages/webview_flutter_lwe/lib/src/lwe_webview.dart
```diff
@@ -48,7 +48,13 @@ class LweWebView {
 
   Future<T?> _invokeChannelMethod<T>(String method, [dynamic arguments]) async {
     if (!_isCreated) {
-      _pendingMethodCalls[method] = arguments;
+      if (method == 'addJavaScriptChannel' ||
+          method == 'runJavaScript' ||
+          method == 'runJavaScriptReturningResult') {
+        _pendingMethodCalls['${method}_$arguments'] = arguments;
```
swift-kim, line 54: `Map` is not well suited for storing this type of data. It would be nice if Dart had a data structure like `multimap`, but there are many alternatives like `List<List>` (you can use a hash set if you don't care about ordering).

JSUYA: I changed to keep existing key(first) value.
The second value is dynamic, so the method's arguments related to javascript use to List<String> type.

JSUYA:
rebased

packages/webview_flutter_lwe/lib/src/lwe_webview.dart
swift-kim: I mean something like this (using `List<List>`).

```dart
// Definition
final List<List<dynamic>> _pendingMethodCalls = <List<dynamic>>[];

// Add
_pendingMethodCalls.add(<dynamic>[method, arguments]);

// Call
for (final List<dynamic> methodCall in _pendingMethodCalls) {
  final String method = methodCall[0] as String;
  final dynamic arguments = methodCall[1];
  await _lweWebViewChannel.invokeMethod<void>(method, arguments);
}
_pendingMethodCalls.clear();
```

The data structure and `await` ensure that the pending calls are invoked in order.

swift-kim: FYI) You can use a tuple ([Record](https://dart.dev/language/records)) if you increase the min SDK version to 3.0.0.

```dart
final List<(String, dynamic)> _pendingMethodCalls = <(String, dynamic)>[];

_pendingMethodCalls.add((method, arguments));

for (final (String method, dynamic arguments) in _pendingMethodCalls) {
  _lweWebViewChannel.invokeMethod<void>(method, arguments);
}
_pendingMethodCalls.clear();
```

JSUYA: thank you for your suggestion. I updated code.

swift-kim (approved):

swift-kim (approved):

xiaowei-guan (approved):

PR #695, opened 2024-07-01
[webview_flutter_lwe] Update lwe(1.2.2)

bwikbs:
Update lightweight web engine(1.2.2).


JSUYA:
I checked that Google Maps plugin works normally. It does not work on some webpages (etc.. pub.dev), but since lwe webview is only used in googlemaps, problems that occur while browsing webpages can be ignored.

JSUYA (approved):

PR #698, opened 2024-07-05
[webview_flutter_tizen] Prevent duplicate Dispose() calls

JSUYA:
Dispose() is called redundantly on destructor and PlatformView.
Modify it to only operate once.

+) Minor fix - Do `setEnginePolicy` method call and returns immediately. It prevents duplicate calls to `result`.

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -171,34 +171,36 @@ std::string WebView::GetNavigationDelegateChannelName() {
 }
 
 void WebView::Dispose() {
-  if (!disposed_) {
-    texture_registrar_->UnregisterTexture(GetTextureId(), nullptr);
-
-    if (webview_instance_) {
-      evas_object_smart_callback_del(webview_instance_,
-                                     "offscreen,frame,rendered",
-                                     &WebView::OnFrameRendered);
-      evas_object_smart_callback_del(webview_instance_, "load,started",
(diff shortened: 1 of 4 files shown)


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -171,31 +171,34 @@ std::string WebView::GetNavigationDelegateChannelName() {
...
-    evas_object_smart_callback_del(webview_instance_, "console,message",
-                                   &WebView::OnConsoleMessage);
-    evas_object_smart_callback_del(webview_instance_,
-                                   "policy,navigation,decide",
-                                   &WebView::OnNavigationPolicy);
-    evas_object_smart_callback_del(webview_instance_, "url,changed",
-                                   &WebView::OnUrlChange);
-    evas_object_del(webview_instance_);
-  }
+  if (!disposed_) {
```
swift-kim, line 174: Or simply
```suggestion
  if (disposed_) {
    return;
  }
```
But is there any case where `Dispose` is called twice?

swift-kim (approved):

packages/webview_flutter/tizen/src/webview.cc
JSUYA: Call `dispose()` on the `Webview` destructor, and also called through `OnDispose()` of `PlatformViewChannel`.
I can also delete `dispose()` of the destructor, but I think this way frees memory more clearly.

PR #700, opened 2024-07-11
[shared_preferences] Create a list from iterable collection

Senne021:
The shared_preferences tizen package should return a list and not an iterable when using getStringList, because the result is casted to a list (see [here](https://github.com/flutter/packages/blob/0ae6dda5d133bebe3b7eb511fea3ecb2c69321ec/packages/shared_preferences/shared_preferences/lib/shared_preferences.dart#L130))  and if the list is empty, a list is also returned (see [here](https://github.com/flutter-tizen/plugins/blob/9bd3d3e14304a0ac184e68c300b112fbeb48df29/packages/shared_preferences/lib/shared_preferences_tizen.dart#L52)). [getRange](https://api.flutter.dev/flutter/dart-core/List/getRange.html) returns an iterable and not a list.

If we return an iterable, the error `type SubListIterable<String>' is not a subtype of type 'List<dynamic>? in type cast` is thrown.


swift-kim (commented):
Thanks for the fix!

The package version in `README.md`, `pubspec.yaml`, and `CHANGELOG.md` needs to be updated if you want this change to be published on pub.dev. We can make the change by pushing a commit to your PR branch if you don't mind.

swift-kim:
Note: This error only occurs when retrieving a string list value after the app is restarted. This is why the integration test could not detect the error.

swift-kim (approved):

Issue #701, opened 2024-07-11
video_player_avplay crash


AMarozzi:
Hello,

Sometimes when i try to play an mp4 file the video_player_avplay crash.

Here the log:

> [E] [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
#0      VideoPlayerController.initialize.errorListener (package:video_player_avplay/video_player.dart:480:51)
#1      _RootZone.runUnaryGuarded (dart:async/zone.dart:1594:10)
#2      _BufferingStreamSubscription._sendError.sendError (dart:async/stream_impl.dart:386:15)
#3      _BufferingStreamSubscription._sendError (dart:async/stream_impl.dart:402:7)
#4      _BufferingStreamSubscription._addError (dart:async/stream_impl.dart:306:7)
#5      _ForwardingStreamSubscription._addError (dart:async/stream_pipe.dart:128:11)
#6      _ForwardingStream._handleError (dart:async/stream_pipe.dart:95:10)
#7      _ForwardingStreamSubscription._handleError (dart:async/stream_pipe.dart:157:13)
#8      _RootZone.runBinaryGuarded (dart:async/zone.dart:1606:10)
#9      _BufferingStreamSubscription._sendError.sendError (dart:async/stream_impl.dart:384:15)
#10     _BufferingStreamSubscription._sendError (dart:async/stream_impl.dart:402:7)
#11     _DelayedError.perform (dart:async/stream_impl.dart:552:14)
#12     _PendingEvents.handleNext (dart:async/stream_impl.dart:646:11)
#13     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:617:7)
#14     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
#15     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
[E] [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
#0      VideoPlayerController.initialize.errorListener (package:video_player_avplay/video_player.dart:480:51)
#1      _RootZone.runUnaryGuarded (dart:async/zone.dart:1594:10)
#2      _BufferingStreamSubscription._sendError.sendError (dart:async/stream_impl.dart:386:15)
#3      _BufferingStreamSubscription._sendError (dart:async/stream_impl.dart:402:7)
[...]

PR #703, opened 2024-07-12
[video_player_avplay]No return empty error message

xiaowei-guan:
Fix https://github.com/flutter-tizen/plugins/issues/701

### DIFF
--- a/packages/video_player_avplay/tizen/src/plus_player.cc
+++ b/packages/video_player_avplay/tizen/src/plus_player.cc
@@ -659,84 +659,58 @@ void PlusPlayer::OnResourceConflicted(void *user_data) {
 
 std::string GetErrorMessage(plusplayer::ErrorType error_code) {
   switch (error_code) {
-    case plusplayer::ErrorType::kNone: {
+    case plusplayer::ErrorType::kNone:
       return "Successful";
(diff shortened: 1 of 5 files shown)


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -657,20 +657,106 @@ void PlusPlayer::OnResourceConflicted(void *user_data) {
   self->SendError("PlusPlayer error", "Resource conflicted");
 }
 
+std::string GetErrorMessage(plusplayer::ErrorType error_code) {
+  switch (error_code) {
+    case plusplayer::ErrorType::kNone: {
+      return "Successful";
+    }
```
swift-kim, line 664: The braces seem redundant.

swift-kim (approved):
Otherwise looks good to me.

You could also increase the package version.

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: Done

xiaowei-guan:
> Otherwise looks good to me.
> 
> You could also increase the package version.

Done

Issue #697, opened 2024-07-04
[webview_flutter_tizen] Crash occurs when creating webview again


JSUYA:
```diff
--- a/packages/webview_flutter/example/lib/main.dart
+++ b/packages/webview_flutter/example/lib/main.dart
@@ -13,7 +13,7 @@ import 'package:flutter/material.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:webview_flutter/webview_flutter.dart';

-void main() => runApp(const MaterialApp(home: WebViewExample()));
+void main() => runApp(const MaterialApp(home: TestPage()));

 const String kNavigationExamplePage = '''
 <!DOCTYPE html><html>
@@ -104,6 +104,36 @@ const String kLogExamplePage = '''
 </html>
 ''';

+class TestPage extends StatefulWidget {
+  const TestPage({super.key});
+
+  @override
+  State<TestPage> createState() => _TestPageState();
+}
+
+class _TestPageState extends State<TestPage> {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('Flutter Demo'),
+      ),
+      body: Center(
+        child: ElevatedButton(
+          onPressed: () {
+            Navigator.push(
+              context,
+              MaterialPageRoute(
+                  builder: (BuildContext context) => const WebViewExample()),
+            );
+          },
+          child: const Text('Go to webview flutter'),
+        ),
+      ),
+    );
+  }
+}
```
I applied this code to webview_flutter_example and tested it.
A crash occurs in ewk_view_add() when create a webview by pressing 'Go to webview flutter' button, close the page by pressing the back button, and create a webview  again.

Test environment : tizen 8.0 + TV

JSUYA:
I checked that there is a problem with the chromium-efl library built into TizenTV. I will make a patch to bypass(?) this.

JSUYA:
Temporarily skipped ewk init(), ewk shutdown().

PR #709, opened 2024-07-17
[webview_flutter] Disable unnecessary web-engine initialization

JSUYA:
ewk_init() and ewk_shutdown() are designed to be called only once in a process.
(If ewk_init() is called after ewk_shutdown() is called, SIGTRAP is called internally.)
ewk_init() initializes the efl module and web engine's arguments data.
The efl module is initialized by default in OS, and Arguments data is also initialized
through SetArguments() API, so calling ewk_init() is not necessary.
Therefore, temporarily comment out ewk_init() and ewk_shutdown().
It can be reverted depending on updates to chromium-efl. ewk_init();

related issue: https://github.com/flutter-tizen/plugins/issues/697

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Disable unnecessary web-engine initialization.
+
 ## 0.9.3
 
 * Prevent duplicate Dispose() calls.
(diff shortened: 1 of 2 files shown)


JSUYA:
there is another option https://github.com/JSUYA/plugins/commit/ae223746a225004647dfd994b631f1daa1a17f77

xiaowei-guan (approved):

PR #713, opened 2024-07-26
[google_maps_flutter] Support longpress event

JSUYA:
GoogleMaps interface has a longpress event, but Google Javascript API does not provide a longpress event.
The existing code implemented longpress using the rightclick event,
but the rightclick event was deprecated when the Javascript API of Google Maps was updated. (https://developers.google.com/maps/documentation/javascript/reference/map#Map.rightclick)
Therefore, Implement longpress event using mousedown/up events.

### DIFF
--- a/packages/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.8
+
+* Support longpress event.
+
 ## 0.1.7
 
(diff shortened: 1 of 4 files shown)


JSUYA:
> LGTM, but I don't understand the context for this PR. Someone please review additionally.

I think, there is nothing wrong with the added javascript code. This change can be tested in the example's 'Map click'. Since it is confirmed to work properly, I will merge this PR.

PR #715, opened 2024-08-07
[flutter_secure_storage] Fix Crashes on Encrypting/Decrypting Empty Strings

MasterSon05:
### Summary

This PR addresses an issue that caused the application to crash when attempting to encrypt or decrypt an empty string.

### Changes

- **Encrypt Method:**
  - Added a check to return a vector containing a single null byte when the input string is empty. This ensures that the function handles empty inputs without errors.

- **Decrypt Method:**
  - Added a check to return an empty string when the input vector size is less than or equal to the initialization vector size. This prevents errors during decryption when the input is insufficient.

### Testing

- Verified that attempting to encrypt or decrypt an empty string no longer causes the application to crash.
- To test with the example app, edit the value in the text field, remove all content to make it an empty string, and then save it. The application should handle this gracefully without crashing.

### Notes

These changes improve the robustness of the encryption and decryption functions by ensuring they handle edge cases gracefully.

### DIFF
--- a/packages/flutter_secure_storage/CHANGELOG.md
+++ b/packages/flutter_secure_storage/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## NEXT
 
+* Fixed an issue causing crashes when encrypting or decrypting empty strings by adding appropriate checks and handling in `Encrypt` and `Decrypt` methods.
 * Fix new lint warnings.
 * Update minimum Flutter and Dart version to 3.13 and 3.1.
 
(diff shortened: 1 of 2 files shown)


JSUYA (approved):
The dart_analyze failed issue requires updates to these packages (battery_plus, flutter_webrtc, permission_handler, sensors_plus).

JSUYA:
I tested it and checked it works fine.

PR #725, opened 2024-08-30
[video_player_avplay] Call the open before calling the SetStreamingProperty

xiaowei-guan:
From SetStreamingProperty API guide:
```
  /**
   * @brief     Set streamingengine property
   * @param     [in] type : attribute type. \n
   *             type can be \n
   *               "COOKIE" \n
   *               "USER_AGENT" \n
   *               "RESUME_TIME" \n
   *               "ADAPTIVE_INFO"
   * @param     [in] value : value of attribute type
   * @pre       The player state can be all of #State except #State::kNone
   */
``` 
We need call open first, then call SetStreamingProperty.

### DIFF
--- a/packages/video_player_avplay/lib/src/video_player_tizen.dart
+++ b/packages/video_player_avplay/lib/src/video_player_tizen.dart
@@ -200,11 +200,9 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
   @override
   Future<String> getStreamingProperty(
       int playerId, StreamingPropertyType type) async {
-    final StreamingPropertyMessage streamingPropertyMessage =
-        await _api.getStreamingProperty(StreamingPropertyTypeMessage(
-            playerId: playerId,
-            streamingPropertyType: _streamingPropertyType[type]!));
-    return streamingPropertyMessage.streamingProperty;
+    return _api.getStreamingProperty(StreamingPropertyTypeMessage(
+        playerId: playerId,
+        streamingPropertyType: _streamingPropertyType[type]!));
   }
 
   @override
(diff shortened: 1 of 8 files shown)


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -345,7 +345,6 @@ std::pair<int64_t, int64_t> PlusPlayer::GetDuration() {
       LOG_ERROR("[PlusPlayer] Player fail to get the duration.");
       return std::make_pair(0, 0);
     }
-    LOG_INFO("[PlusPlayer] Video duration: %lld.", duration);
```
JSUYA, line 348: Is there any reason for removing it?

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -200,11 +200,9 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
   Future<String> getStreamingProperty(
       int playerId, StreamingPropertyType type) async {
-    final StreamingPropertyMessage streamingPropertyMessage =
-        await _api.getStreamingProperty(StreamingPropertyTypeMessage(
-            playerId: playerId,
-            streamingPropertyType: _streamingPropertyType[type]!));
-    return streamingPropertyMessage.streamingProperty;
+    return _api.getStreamingProperty(StreamingPropertyTypeMessage(
+        playerId: playerId,
+        streamingPropertyType: _streamingPropertyType[type]!));
```
JSUYA, line 205: Could you please write the changing that the type of getStreamingProperty in CHANGELOG.md or PR message?

JSUYA (approved):

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: Because this API will be called once one second,this will result in a lot of logs, affecting debugging.

PR #739, opened 2024-09-09
[video_player] Update pigeon to 22.3.0

JSUYA:
- Update pigeon from 10.0.0 to 22.3.0.
- Temporarily set duration to 1, if duration is 0.


JSUYA:
Could you please write the changes in lwe version in Changelog?

JSUYA:
> Temporarily set duration to 1, if duration is 0.

This change bypasses exceptions in some streaming content. (https://live-ch2.cjonstyle.net/cjosplus/live2/playlist.m3u8)

packages/video_player/tizen/src/video_player.cc
```diff
@@ -465,6 +465,15 @@ void VideoPlayer::SendInitialized() {
       }
     }
 
+    // TODO(jsuya):Some streaming resources may have a duration of 0 during the
+    // initialization step. If the duration is 0, it may affect the progress of
+    // video_player and cause unnecessary errors. Therefore, set it to 1
+    // temporarily. the future. In the future, It can be updated depending on
```
seungsoo47, line 471:
```suggestion
    // temporarily. In the future, It can be updated depending on
```

seungsoo47 (approved):

PR #741, opened 2024-09-10
[video_player_avplay] Fix DashEngine stream property "STARTBITRATE" can't be set correctly

xiaowei-guan:


JSUYA:
How about updating along with [this PR](https://github.com/flutter-tizen/plugins/pull/731)? Is it okay to just update libdash?

xiaowei-guan:
> How about updating along with [this PR](https://github.com/flutter-tizen/plugins/pull/731)? Is it okay to just update libdash?



> How about updating along with [this PR](https://github.com/flutter-tizen/plugins/pull/731)? Is it okay to just update libdash?

https://github.com/flutter-tizen/plugins/pull/731 has two crash issues, we are checking it now.
The 'STARTBITRATE' not works issue from libertyglobal, this is urgent issue for them, so I create a separate PR.

JSUYA (approved):

PR #742, opened 2024-09-11
[video_player] Live streaming content starts playing immediately when seekTo is called

JSUYA:
According to Tizen media player's member, Tizen player does not support `seekTo` for live streaming content.

Therefore, `duration` is returned as 0. When `duration` is 0, an error occurs due to a zero divide error(`NaN` value) in the progress bar of `video_player`. And the resume behavior(`play()`) of `video_player` calls `seekTo(0)` when `duration` and `position` are the same(when the video has been played to the end), which causes the video to not be played again in the case of live streaming content.

Therefore, in the case of live, if `set_play_position()` fails, call `play()` immediately to playback.


JSUYA:
You can test it using this(https://live-ch2.cjonstyle.net/cjosplus/live2/playlist.m3u8).

packages/video_player/tizen/src/video_player.cc
```diff
@@ -465,21 +474,23 @@ void VideoPlayer::SendInitialized() {
...
-    // the loading status.
-    if (width != 0 && height != 0 && duration == 0) {
-      duration = 1;
+    // TODO(jsuya):Tizen's player API returns duration as 0 for live
+    // streaming resources. If the duration is 0, it may affect the progress of
+    // video_player and cause unnecessary errors. Therefore, set it to 1.
+    if (uri_.substr(0, 4) == "http" && width != 0 && height != 0 &&
+        duration_ == 0) {
+      duration_ = 1;
+      is_live_ = true;
```
xiaowei-guan, line 483: Actually, media player has API for streaming info, we can get the is_live info:
```C
  int player_get_adaptive_streaming_info(player_h player, void* adaptive_info,
                                         int adaptive_type);
  bool MediaPlayer::IsLive() {
  int is_live = 0;
  int ret = media_player_proxy_->player_get_adaptive_streaming_info(
      player_, &is_live, PLAYER_ADAPTIVE_INFO_IS_LIVE);
    LOG_ERROR("[MediaPlayer] player_get_adaptive_streaming_info failed: %s",
              get_error_message(ret));
    return false;
  }
  return is_live != 0;
}
```
you can refer to :
https://github.com/flutter-tizen/plugins/blob/master/packages/video_player_videohole/tizen/src/media_player_proxy.h

JSUYA: Thank you for information. 
However, video_player_tizen should work on common profile(rpi4) and, unless there are special cases, should be able to work on tizen emulator(8.0) . `player_get_adaptive_streaming_info()` has been confirmed to work only on TV devices. I will make a proxy class of mediaplayer and modify the code like videohole. but, the `duration == 0` code can be kept for rpi (common profile).

xiaowei-guan (approved):

PR #748, opened 2024-09-20
[shared_preferences] Update shared_preferences to 2.3.2

JSUYA:
* Update shared_preferences to 2.3.2.
* Update shared_preferences_interface to 2.4.0.
* Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
* Supports `SharedPreferencesAsync` and `SharedPreferencesWithCache` APIs.


packages/shared_preferences/example/pubspec.yaml
```diff
@@ -3,13 +3,13 @@ description: Demonstrates how to use the shared_preferences_tizen plugin.
 publish_to: "none"
 
 environment:
-  sdk: ">=3.1.0 <4.0.0"
-  flutter: ">=3.13.0"
+  sdk: ^3.3.0
+  flutter: ">=3.19.0"
```
seungsoo47, line 7: I have a question. Could you explain why you set the versions lower than original plugin's version?

JSUYA: I did refer to the version of the platform interface. And saw a comment that this was related to N-1 test(?) in the 3.19 and 3.22 updates, but I couldn't find the exact details about this, so I went with 3.19.
However, I decided that it would be better to sync with the latest version, so I modified the patch.

packages/shared_preferences/lib/src/shared_preferences_tizen.dart
```diff
@@ -0,0 +1,198 @@
+// Copyright 2020 Samsung Electronics Co., Ltd. All rights reserved.
```
seungsoo47, line 1:
```suggestion
// Copyright 2024 Samsung Electronics Co., Ltd. All rights reserved.
```

packages/shared_preferences/example/pubspec.yaml
seungsoo47: Thank you for your detailed reply :)

seungsoo47 (approved):

xiaowei-guan (approved):

PR #756, opened 2024-10-23
[tools] Bump up sdk version

JSUYA:
Older Dart SDK Versions are deprecated. So sdk version must be upgraded to 3.0 or higher.
https://groups.google.com/g/flutter-announce/c/JQHzM3FbBGI/m/B-i8IVa1AQAJ

+) For wearable_rotary, it is no longer updated, but it runs together with the build test of CI due to the change in tools.
Since it causes a break in CI, so temporarily skip the build of wearable_rotary.

### DIFF
--- a/tools/lib/src/build_examples_command.dart
+++ b/tools/lib/src/build_examples_command.dart
@@ -23,16 +23,6 @@ class BuildExamplesCommand extends PackageLoopingCommand {
 
     bool builtSomething = false;
     for (final RepositoryPackage example in package.getExamples()) {
-      final String packageName = getRelativePosixPath(
-        example.directory,
-        from: packagesDir,
-      );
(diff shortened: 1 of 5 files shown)


tools/lib/src/build_examples_command.dart
```diff
@@ -23,6 +23,16 @@ class BuildExamplesCommand extends PackageLoopingCommand {
 
     bool builtSomething = false;
     for (final RepositoryPackage example in package.getExamples()) {
+      final String packageName = getRelativePosixPath(
+        example.directory,
+        from: packagesDir,
+      );
+      // TODO(jsuya): Temporarily skip build test of wearable_rotary.
+      if (packageName == 'wearable_rotary/example') {
```
swift-kim, line 31: Will `build-examples --exclude` work?

swift-kim (approved):

PR #759, opened 2024-10-28
[video_player_avplay] Add start position in player options when creating player

xiaowei-guan:
Add start position in player options when creating player,this is useful for resuming playback from last viewed position.
code sample:
```dart
    _controller = VideoPlayerController.network(
        'https://media.w3.org/2010/05/bunny/trailer.mp4',
         // startPosition in millisecond
        playerOptions: {'startPosition': 10 * 1000});
```

### DIFF
--- a/packages/video_player_avplay/tizen/src/plus_player.cc
+++ b/packages/video_player_avplay/tizen/src/plus_player.cc
@@ -141,7 +141,9 @@ int64_t PlusPlayer::Create(const std::string &uri,
       create_message.player_options(), "startPosition", 0);
   if (start_position > 0) {
     LOG_INFO("[PlusPlayer] Start position: %lld", start_position);
-    Seek(player_, start_position);
+    if (!Seek(player_, start_position)) {
+      LOG_INFO("[PlusPlayer] Fail to seek, it's a non-seekable content");
+    }
   }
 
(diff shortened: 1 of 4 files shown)


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -137,6 +137,13 @@ int64_t PlusPlayer::Create(const std::string &uri,
     is_prebuffer_mode_ = true;
   }
 
+  int64_t start_position = flutter_common::GetValue(
+      create_message.player_options(), "startPosition", 0);
+  if (start_position > 0) {
+    LOG_INFO("[PlusPlayer] Start position: %lld", start_position);
+    Seek(player_, start_position);
```
JSUYA, line 144: As far as I know, non-seekable content is returned as false, so you can add as the log that the start position didn't work.

JSUYA (approved):
It seems that the condition of the state calling the Seek() API could be `kIdle` or higher. The state condition of SeekTo() could also be modified to avoid confusion.

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: Done

PR #773, opened 2024-11-12
[video_player_avplay] Fix start position out of range issue

xiaowei-guan:
Because the type of 0 is int_32 on ARM32,  convert the type of 0 to int64_t.


xuelian-bai (approved):
looks good to me

Issue #758, opened 2024-10-25
Plugins that need to be updated


JSUYA:
As of 10/22/2024, I have reviewed the versions of plugins and organized the plugins that need to be updated.

| plugin | current (tizen) |  current (frontend)| target (frontend) |  note | 
|:-|:-:|:-:|:-:|:-|
| audioplayers_tizen | 3.0.1 | 5.1.0 | 6.1.0 | Deprecated onCurrentPosition event |
| connectivity_plus_tizen |1.1.4 | 4.0.1 | 6.1.0| Added support for multiple connectivity results |
| flutter_tts_tizen | 1.4.0 | 3.6.3 | 4.0.2 | Added pitch related features | 
| google_maps_flutter_tizen | 0.1.8 |  2.7.0 | 2.9.0 | Added clustering related features | 
| in_app_purchase_tizen | 0.1.0 |  3.1.4 | 3.2.0 | Added countryCode |
| url_launcher_tizen |  2.1.2 | 6.1.0 | 6.3.1 | There are many APIs that are not implemented, such as `closeInAppWebView`. Currently, they are at least working.|
| webview_flutter_lwe / webview_flutter_tizen | 0.3.3 / 0.9.2 | 4.4.2 | 4.10.0 | Added JavaScriptLogLevel, HTTP basic authentication, WebViewController.setOnJavaScriptAlertDialog, WebViewController.setOnJavaScriptConfirmDialog WebViewController.setOnJavaScriptTextInputDialog, onHttpError, etc. |

(`firebase_core_tizen`, `flutter_webrtc_tizen`, `google_sign_in_tizen`, and `sqflite_tizen` packages have been excluded from the priority review due to their complex history and implementation. These should be reviewed separately for updates.)

JSUYA:
Most of plugins have been updated. Thanks to everyone. See you next quarter~

PR #775, opened 2024-11-15
[flutter_tts] Update flutter_tts to 4.2.0

seungsoo47:
Update flutter_tts 3.6.3 to 4.2.0.
Update the example app.
Add isLanguageAvailable, getVoices and setVoice APIs which were not implemented.

### DIFF
--- a/packages/flutter_tts/example/lib/main.dart
+++ b/packages/flutter_tts/example/lib/main.dart
@@ -42,6 +42,8 @@ class _MyAppState extends State<MyApp> {
 
     _getDefaultVoice();
 
+    _getSupportedVoices();
+
     flutterTts.setStartHandler(() {
       setState(() {
         print("Playing");
@@ -90,7 +92,14 @@ class _MyAppState extends State<MyApp> {
   Future<void> _getDefaultVoice() async {
     var voice = await flutterTts.getDefaultVoice;
     if (voice != null) {
-      print(voice);
+      print('_getDefaultVoice(): $voice');
+    }
+  }
+
+  Future<void> _getSupportedVoices() async {
+    var voices = await flutterTts.getVoices;
+    if (voices != null) {
+      print('_getSupportedVoices(): $voices');
     }
   }
 
--- a/packages/flutter_tts/example/lib/main.dart
+++ b/packages/flutter_tts/example/lib/main.dart
@@ -24,10 +24,10 @@ class _MyAppState extends State<MyApp> {
 
   TtsState ttsState = TtsState.stopped;
 
-  get isPlaying => ttsState == TtsState.playing;
-  get isStopped => ttsState == TtsState.stopped;
-  get isPaused => ttsState == TtsState.paused;
-  get isContinued => ttsState == TtsState.continued;
+  bool get isPlaying => ttsState == TtsState.playing;
+  bool get isStopped => ttsState == TtsState.stopped;
+  bool get isPaused => ttsState == TtsState.paused;
+  bool get isContinued => ttsState == TtsState.continued;
 
   @override
   initState() {
@@ -35,7 +35,7 @@ class _MyAppState extends State<MyApp> {
     initTts();
   }
 
-  initTts() {
+  dynamic initTts() {
     flutterTts = FlutterTts();
 
     _setAwaitOptions();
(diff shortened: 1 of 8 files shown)


packages/flutter_tts/example/lib/main.dart
```diff
@@ -175,23 +187,27 @@ class _MyAppState extends State<MyApp> {
...
-      padding: EdgeInsets.only(top: 25.0, left: 25.0, right: 25.0),
-      child: TextField(
-        maxLines: 11,
-        minLines: 6,
-        onChanged: (String value) {
-          _onChange(value);
-        },
-      ));
+  Widget _inputSection() {
+    _newVoiceText = 'Hello everyone. This is a flutter tts example app.';
```
JSUYA, line 198: I couldn't find this code in flutter_tts/example. Do we need it?

+) 208 line

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -107,8 +137,8 @@ void TextToSpeech::UnregisterCallbacks() {
   tts_unset_error_cb(tts_);
 }
 
-std::vector<std::string> &TextToSpeech::GetSupportedLanaguages() {
-  if (supported_lanaguages_.size() == 0) {
+void TextToSpeech::InitializeSupportedLanaguagesAndVoiceType() {
+  if (supported_lanaguages_.size() == 0 || supported_voice_types_.size() == 0) {
```
JSUYA, line 141: This function was executed for every API call,
but now it is executed once during the initialization phase.
This condition and lines 163-168 seem unnecessary. what do you think about this?

packages/flutter_tts/tizen/src/text_to_speech.cc
```diff
@@ -245,6 +273,14 @@ bool TextToSpeech::GetSpeedRange(int32_t *min, int32_t *normal, int32_t *max) {
   return true;
 }
 
+bool TextToSpeech::IsLanguageAvailable(const std::string &language) {
+  if (std::find(supported_lanaguages_.begin(), supported_lanaguages_.end(),
```
JSUYA, line 277: How about simply immediate return?
`return std::find(supported_lanaguages_.begin(), supported_lanaguages_.end(),language) != ported_lanaguages_.end();`

seungsoo47: Thank you for your review.

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -195,9 +215,11 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
     SendResult(flutter::EncodableValue(0));
   }
 
-  void OnGetLanguage() {
+  void OnGetLanguages() {
+    auto languages = tts_->GetSupportedLanaguages();
```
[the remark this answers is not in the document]
seungsoo47: Thank you for your review. I will revert them to original code.

packages/flutter_tts/tizen/src/text_to_speech.cc
seungsoo47: Thank you for your review.
Yes, I will remove the condition.
But, can you explain about "lines 163-168 seem unnecessary" in detail?

packages/flutter_tts/example/lib/main.dart
seungsoo47: It was very inconvenient to have to add text every time I tested. So I added the text. 
If I have to delete the text, I will.

JSUYA: We keep the frontend package's example code as much as possible for smooth version updates.
So I want to reduce unnecessary diffs as much as possible unless it is platform-specific code.
So my opinion is that it would be better to delete it.
https://pub.dev/packages/flutter_tts/example

packages/flutter_tts/tizen/src/text_to_speech.cc
JSUYA: Before the change, `TextToSpeech::GetSupportedLanaguages()` was called every time `getLanguages` was called.
So it cleared the vector list before adding items to `vector` using the callback of `tts_foreach_supported_voices`.

But now, `TextToSpeech::InitializeSupportedLanaguagesAndVoiceType()` is called only once via `TextToSpeech::Initialize().`
There is no need to clear `supported_lanaguages_` and `supported_voice_types_` as they will not be rewritten.
And `supported_lanaguages_` and `supported_voice_types_` are already empty lists.

I think it will very rare for `supported_language_` to change, although there may be issues with the list synchronization if system specifications change in the middle of the program.

packages/flutter_tts/example/lib/main.dart
seungsoo47: OK. I understood. Thank you for your review.

packages/flutter_tts/tizen/src/text_to_speech.cc
seungsoo47: I got it. Thanks for your review.

JSUYA (approved):

JSUYA:

PR #778, opened 2024-11-25
[ video_player_avplay] Make startPosition support int32_t

xiaowei-guan:
The startPosition could be int64_t or int32_t, we need to support both.

### DIFF
--- a/packages/video_player_avplay/tizen/src/plus_player.cc
+++ b/packages/video_player_avplay/tizen/src/plus_player.cc
@@ -139,6 +139,10 @@ int64_t PlusPlayer::Create(const std::string &uri,
 
   int64_t start_position = flutter_common::GetValue(
       create_message.player_options(), "startPosition", (int64_t)0);
+  if (start_position == 0) {
+    start_position = flutter_common::GetValue(create_message.player_options(),
+                                              "startPosition", (int32_t)0);
+  }
   if (start_position > 0) {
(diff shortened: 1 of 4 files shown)


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -139,6 +139,10 @@ int64_t PlusPlayer::Create(const std::string &uri,
 
   int64_t start_position = flutter_common::GetValue(
       create_message.player_options(), "startPosition", (int64_t)0);
+  if (start_position == 0) {
+    start_position = flutter_common::GetValue(create_message.player_options(),
+                                              "startPosition", (int32_t)0);
```
JSUYA, line 144: I was looking for a way to use std::variant<int32_t, int64_t> and std::holds_alternative<T>, std::get<int32_t>.
But I gave up because it seemed like it would only increase code complexity.

But I think it's a code that should be improved someday. Could you please leave a comment in code(//TODO?) why int32_t is needed?

JSUYA (approved):

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: I also realized this problem, we can release it for 3rd-party first. I will implement a new function for all types later.

PR #784, opened 2024-12-20
[video_player_avplay] Support Tizen 9.0

xiaowei-guan:


JSUYA (approved):
I tested that the app worked properly when I build and ran it on a Tizen 9.0 TV binary with the api-version set to 9.0.
There were no problems with overall playback.

However, when moving the tab, the following exception was detected.
This exception also occurs in the videohole package, so it is not affected by this PR.
```
══╡ EXCEPTION CAUGHT BY FOUNDATION LIBRARY ╞════════════════════════════════════════════════════════
The following assertion was thrown while dispatching notifications for VideoPlayerController:
setState() called after dispose(): _DashRomoteVideoState#ced13(lifecycle state: defunct, not
mounted)
This error happens if you call setState() on a State object for a widget that no longer appears in
the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error
can occur when code calls setState() from a timer or an animation callback.
The preferred solution is to cancel the timer or stop listening to the animation in the dispose()
callback. Another solution is to check the "mounted" property of this object before calling
setState() to ensure the object is still in the tree.
This error might indicate a memory leak if setState() is being called because another object is
retaining a reference to this State object after it has been removed from the tree. To avoid memory
leaks, consider breaking the reference to this object during dispose().

When the exception was thrown, this was the stack:
#0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1168:9)
#1      State.setState (package:flutter/src/widgets/framework.dart:1203:6)
#2      _DashRomoteVideoState.initState.<anonymous closure> (package:video_player_avplay_example/main.dart:144:7)
#3      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:437:24)
#4      ValueNotifier.value= (package:flutter/src/foundation/change_notifier.dart:559:5)
#5      VideoPlayerController._updatePosition (package:video_player_avplay/video_player.dart:833:5)
#6      VideoPlayerController._applyPlayPause.<anonymous closure> (package:video_player_avplay/video_player.dart:635:11)
<asynchronous suspension>

The VideoPlayerController sending notification was:
  VideoPlayerController#133aa(VideoPlayerValue(duration: DurationRange(start: 0:00:00.000000, end:
  0:11:58.998000), size: Size(3840.0, 2160.0), position: 0:00:01.023000, caption: Caption(number: 0,
  start: 0:00:00.000000, end: 0:00:00.000000, text: ), captionOffset: 0:00:00.000000, tracks: [],
  buffered: 99, isInitialized: true, isPlaying: true, isLooping: true, isBuffering: false, volume:
  1.0, playbackSpeed: 1.0, errorDescription: null, isCompleted: false),)
════════════════════════════════════════════════════════════════════════════════════════════════════

```

JSUYA:
https://github.com/flutter-tizen/flutter-tizen/issues/570

seungsoo47:
I tested the commits and it seems that the example is working properly.
And I will merge this plugin to resolve DF241202-00991.
Please let me know if you have any issues.

PR #790, opened 2025-01-06
[webview_flutter_lwe] Fix "channel sent a message from native to Flutter on a non-platform thread" error

seungsoo47:
When following code is run:
```
navigation_delegate_channel_->InvokeMethod(
       "onPageStarted", std::make_unique<flutter::EncodableValue>(args));
```
this message happens
```
[ERROR:flutter/shell/common/shell.cc(1065)] The 'plugins.flutter.io/lwe_webview_navigation_delegate_0' channel 
sent a message from native to Flutter on a non-platform thread. Platform channel messages must be sent on the 
platform thread. Failure to do so may result in data loss or crashes, and must be fixed in the plugin or application 
code creating that channel.
See https://docs.flutter.dev/platform-integration/platform-channels#channels-and-platform-threading for more information.
```
So, in order to resolve this problem, I added a function to dispatch tasks on the main thread.


packages/webview_flutter_lwe/tizen/src/webview.h
```diff
@@ -66,6 +68,7 @@ class WebView : public PlatformView {
   void InitWebView();
 
   LWE::WebContainer* webview_instance_ = nullptr;
+  MessageDispatcher* dispatcher = nullptr;
```
JSUYA, line 71: This object goes with the lifecycle of the WebView. I think it would be  more efficient to declare it as a unique_ptr?

packages/webview_flutter_lwe/tizen/src/webview.cc
```diff
@@ -212,9 +226,11 @@ void WebView::RegisterJavaScriptChannelName(const std::string& name) {
         {flutter::EncodableValue("channel"), flutter::EncodableValue(name)},
         {flutter::EncodableValue("message"), flutter::EncodableValue(message)},
     };
-    webview_channel_->InvokeMethod(
-        "javaScriptChannelMessage",
-        std::make_unique<flutter::EncodableValue>(args));
+    dispatcher->dispatchTaskOnMainThread([&]() {
```
JSUYA, line 229: How about narrowing the scope of the capture rather than using `&`? You could use `[this, &args]`.

JSUYA (approved):
LGTM

Issue #786, opened 2024-12-23
[video_player_avplay] Need to OS version check


JSUYA:
The avplay package must be built with the same OS version and application api-version.
in some cases, compatibility may be supported and execution may be possible, but it does not guarantee correct operation.

Therefore, when an app is executed that does not match the OS version or when the app is executed on an unsupported OS version, an appropriate exception and guidance must be provided.

We need to find the necessary ways to support this.

xiaowei-guan:
We implemented a version that didn't need version check, but it  would include all versions libs in TPK file, and makde the tpk file size larger.
https://github.com/hyue7/plugins/tree/avplay/muti_api/packages/video_player_avplay/

JSUYA:
As you all know, there is currently no way to solve the compatibility issue of avplay.

I want to compare the api-version(defined in the manifest) and platform version in the initialize step of avplay and, if they are different, cause the initialization of the video to fail.
This will prevent apps from unintentionally serving the tpk package on an OS version that is not supported.
I also think that it will prevent mistakes in the verification phase of the new Tizen OS.

So.. The platform version can be obtained from the device_info plugin. The api-version can be provided by adding an API in the [flutter-tizen plugin](https://github.com/flutter-tizen/flutter-tizen/blob/master/packages/flutter_tizen/lib/flutter_tizen.dart). I will work on this and plan to add a comparison step to avplay.

Please share your idea if you have any.

PR #793, opened 2025-01-07
[video_player_avplay] Add check platform and api version

JSUYA:
Add a step to the initialization phase that compares the API version of the app using avplay with the version of TizenOS.
This prevents TPKs built with avplay from running on TizenOS versions for which they are not intended.

related issue: https://github.com/flutter-tizen/plugins/issues/786


JSUYA:
rebased

xiaowei-guan (approved):

PR #804, opened 2025-01-22
[webview_flutter_lwe] Update webview_flutter to 4.10.0

seungsoo47:
* Update minimum Flutter and Dart version to 3.24 and 3.5.
* Update the example main app.
* Update webview_flutter to 4.10.0.
* Update webview_flutter_platform_interface to 2.10.0.
* Add removeJavaScriptChannel method call.
* Fix deadlock issue caused by calling setBackgroundColor() after the WebViewController is created.

### DIFF
--- a/packages/webview_flutter_lwe/example/lib/main.dart
+++ b/packages/webview_flutter_lwe/example/lib/main.dart
@@ -12,12 +12,6 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:webview_flutter/webview_flutter.dart';
-// #docregion platform_imports
-// Import for Android features.
-import 'package:webview_flutter_android/webview_flutter_android.dart';
-// Import for iOS/macOS features.
-import 'package:webview_flutter_wkwebview/webview_flutter_wkwebview.dart';
-// #enddocregion platform_imports
 
 void main() => runApp(const MaterialApp(home: WebViewExample()));
 
@@ -122,14 +116,7 @@ class _WebViewExampleState extends State<WebViewExample> {
 
     // #docregion platform_features
     late final PlatformWebViewControllerCreationParams params;
-    if (WebViewPlatform.instance is WebKitWebViewPlatform) {
-      params = WebKitWebViewControllerCreationParams(
-        allowsInlineMediaPlayback: true,
-        mediaTypesRequiringUserAction: const <PlaybackMediaTypes>{},
-      );
-    } else {
-      params = const PlatformWebViewControllerCreationParams();
-    }
+    params = const PlatformWebViewControllerCreationParams();
 
     final WebViewController controller =
         WebViewController.fromPlatformCreationParams(params);
@@ -137,7 +124,7 @@ class _WebViewExampleState extends State<WebViewExample> {
 
     controller
       ..setJavaScriptMode(JavaScriptMode.unrestricted)
-      ..setBackgroundColor(const Color(0x80000000))
+      ..setBackgroundColor(const Color(0x00000000))
       ..setNavigationDelegate(
         NavigationDelegate(
           onProgress: (int progress) {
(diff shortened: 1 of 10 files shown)


packages/webview_flutter_lwe/example/pubspec.yaml
```diff
@@ -3,18 +3,20 @@ description: Demonstrates how to use the webview_flutter_lwe plugin.
...
     sdk: flutter
   path_provider: ^2.0.7
   path_provider_tizen:
     path: ../../path_provider/
-  webview_flutter: ^4.4.2
+  webview_flutter: ^4.10.0
+  webview_flutter_android: ^4.0.0
   webview_flutter_lwe:
     path: ../
+  webview_flutter_wkwebview: ^3.13.0
```
JSUYA, line 19: webview_flutter_android and webview_flutter_wkwebview are unnecessary.
It seems like they were added because of example/lib/main.dart.
There is no need to keep code that is not related to the our platform.

packages/webview_flutter_lwe/example/lib/main.dart
```diff
@@ -108,15 +114,30 @@ class WebViewExample extends StatefulWidget {
...
+    }
+
+    final WebViewController controller =
+        WebViewController.fromPlatformCreationParams(params);
+    // #enddocregion platform_features
+
+    controller
       ..setJavaScriptMode(JavaScriptMode.unrestricted)
-      ..setBackgroundColor(const Color(0x00000000))
+      ..setBackgroundColor(const Color(0x80000000))
```
JSUYA, line 140: Please keep the code related of setBackgroundColor() in the example code of webview_flutter.
(I think it would be better to handle the issue of setBackgroundColor() not working after loadRequest() on a specific site separately. Or update it together in this PR.)

packages/webview_flutter_lwe/example/pubspec.yaml
```diff
@@ -25,6 +25,7 @@ dev_dependencies:
     sdk: flutter
   integration_test_tizen:
     path: ../../integration_test/
+  webview_flutter_platform_interface: ^2.10.0
```
JSUYA, line 28: This dependency is not necessary.

packages/webview_flutter_lwe/lib/src/lwe_webview_controller.dart
```diff
@@ -62,14 +62,6 @@ class LweWebViewController extends PlatformWebViewController {
           'LoadRequestParams#uri is required to have a scheme.');
     }
 
-    if (params.headers.isNotEmpty) {
-      throw ArgumentError('LoadRequestParams#headers is not supported.');
-    }
-
-    if (params.body != null) {
-      throw ArgumentError('LoadRequestParams#body is not supported.');
-    }
```
JSUYA, line 71: Just curious: is there a reason to skip this check? Are there cases when the body and header are empty?

packages/webview_flutter_lwe/example/lib/main.dart
```diff
@@ -157,6 +176,13 @@ Page resource error:
         },
       )
       ..loadRequest(Uri.parse('https://flutter.dev'));
+
+    // setBackgroundColor is not currently supported on macOS.
+    if (kIsWeb || !Platform.isMacOS) {
+      controller.setBackgroundColor(const Color(0x80000000));
+    }
+
```
JSUYA, line 184: This code is not new and does not apply to us, so you can delete it.

JSUYA (approved):
I left a some comments, please check it. Others are good.

JSUYA:
Analyze CI will be fixed with this PR: https://github.com/flutter-tizen/plugins/pull/808

packages/webview_flutter_lwe/lib/src/lwe_webview_controller.dart
seungsoo47: The params.headers and params.body variables are used in post method, and we already had UnimplementedError() to handle those exceptions. So I thought the conditions were redundant.

packages/webview_flutter_lwe/example/lib/main.dart
seungsoo47: Because Platform.isTizen is true and this condition affects to our webview, I did not remove it.

seungsoo47:

PR #805, opened 2025-01-22
[video_player_avplay] Revert 0.5.9 version

xiaowei-guan:
https://github.com/flutter-tizen/plugins/pull/788 cause another side effect.
MM team is preparing the new patch , and MM team can't complete it in a short time, so we want to revert this PR first.
After Mm team complete it, we make a new PR.


gin7773:
LGTM

gin7773 (approved):

JSUYA:
https://github.com/flutter-tizen/plugins/pull/788

JSUYA (commented):
9.0/libdash.so was not updated in #788. 
Is it possible that the problem is caused by 9.0/libdash.so not being updated?

xiaowei-guan:
> 9.0/libdash.so was not updated in #788. Is it possible that the problem is caused by 9.0/libdash.so not being updated?

Actually, I have update the libdash.so for the seek fail patch:
https://github.com/flutter-tizen/plugins/pull/784/commits/570a534fd577c58cc00ee7d0d7712d0183e071fc

JSUYA (approved):

PR #812, opened 2025-02-06
[video_player_avplay] Synchronize isPlaying state

gin7773:
Main changes:

- [PLAYER] Synchronize isPlaying state


packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -861,4 +864,7 @@ void PlusPlayer::OnCueOutContEvent(const char *cue_out_cont_data,
 
 void PlusPlayer::OnChangeSourceDone(bool ret, void *user_data) {}
 
-void PlusPlayer::OnStateChangedToPlaying(void *user_data) {}
+void PlusPlayer::OnStateChangedToPlaying(void *user_data) {
+  PlusPlayer *self = reinterpret_cast<PlusPlayer *>(user_data);
+  self->SendIsPlayingState(true);
```
JSUYA, line 869: Just question: Is this callback called after buffering is complete?

media player changes state to true when calling player_start().
Of course, according to the description of player_start(), it says that the state will changes to PLAYER_STATE_PLAYING when called function, but I think it may work differently from plus player.
In fact, it may not be a problem. hahaha..

packages/video_player_avplay/tizen/src/video_player.cc
```diff
@@ -163,6 +163,16 @@ void VideoPlayer::SendPlayCompleted() {
   PushEvent(flutter::EncodableValue(result));
 }
 
+void VideoPlayer::SendIsPlayingState(bool is_playing) {
+  flutter::EncodableMap result = {
+      {flutter::EncodableValue("event"),
+       flutter::EncodableValue("isPlayingStateUpdate")},
+      {flutter::EncodableValue("isPlayingState"),
```
JSUYA, line 170: For reduce confusion with other platforms, I recommend to use the same key name.
(`isPlayingState` -> `isPlaying`)
https://github.com/flutter/packages/pull/3261/files#diff-7bf3fe00ba4225d8c58e9e7525747626af4d510cf2faf3986b41374ee238bcfaR240

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: The plusplayer has onStateChangedToPlaying callback, it will be invoked when player state was changed to playing,
we just need add this callback to listen playing state.
The media player doesn't have such callback, we need to send a playing state when calling play_start.

JSUYA (approved):

xiaowei-guan (approved):

PR #816, opened 2025-02-11
[google_maps_flutter] Update google_maps_flutter to 2.10.0

JSUYA:
- Update google_maps_flutter to 2.10.0.
- Update google_maps_flutter_platform_interface to 2.10.0.
- Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.

### DIFF
--- a/packages/google_maps_flutter/example/integration_test/src/maps_inspector.dart
+++ b/packages/google_maps_flutter/example/integration_test/src/maps_inspector.dart
@@ -533,4 +533,106 @@ void runTests() {
       expect(myLocationButtonEnabled, true);
     });
   }, skip: !isIOS);
+
+  testWidgets('marker clustering', (WidgetTester tester) async {
+    final Key key = GlobalKey();
+    const int clusterManagersAmount = 2;
+    const int markersPerClusterManager = 5;
+    final Map<MarkerId, Marker> markers = <MarkerId, Marker>{};
+    final Set<ClusterManager> clusterManagers = <ClusterManager>{};
+
+    for (int i = 0; i < clusterManagersAmount; i++) {
+      final ClusterManagerId clusterManagerId =
+          ClusterManagerId('cluster_manager_$i');
+      final ClusterManager clusterManager =
+          ClusterManager(clusterManagerId: clusterManagerId);
+      clusterManagers.add(clusterManager);
+    }
+
+    for (final ClusterManager cm in clusterManagers) {
+      for (int i = 0; i < markersPerClusterManager; i++) {
+        final MarkerId markerId =
+            MarkerId('${cm.clusterManagerId.value}_marker_$i');
+        final Marker marker = Marker(
+            markerId: markerId,
+            clusterManagerId: cm.clusterManagerId,
+            position: LatLng(
+                kInitialMapCenter.latitude + i, kInitialMapCenter.longitude));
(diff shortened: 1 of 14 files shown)


JSUYA:
I implemented support for heatmap API but had a error in webview_flutter_lwe.
I will add this patch when the web engine is improved. https://github.com/flutter-tizen/plugins/commit/af8ea4e6308701bc15eaefe7988561dceebe1c7b

JSUYA:
rebased pr

packages/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -533,4 +533,106 @@ void runTests() {
       expect(myLocationButtonEnabled, true);
     });
   }, skip: !isIOS);
+
+  testWidgets('marker clustering', (WidgetTester tester) async {
```
seungsoo47, line 537: It seems to work properly as the ClusterManager class of the platform interface is added. 
Just question: Do we need to implement more of this feature?

packages/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
```diff
@@ -204,6 +207,271 @@ void runTests() {
...
+      ],
+    ),
+    maxIntensity: 1,
+    opacity: 0.5,
+    radius: HeatmapRadius.fromPixels(40),
+    minimumZoomIntensity: 1,
+    maximumZoomIntensity: 20,
+  );
+
+  testWidgets('set heatmap correctly', (WidgetTester tester) async {
```
seungsoo47, line 331: What do you think about that the patch which you implement and the test cases together are added?

packages/google_maps_flutter/example/integration_test/src/maps_inspector.dart
JSUYA: https://github.com/flutter-tizen/plugins/blob/777966de1d6585143b6ec573ac41828fd1218764/packages/google_maps_flutter/example/integration_test/google_maps_test.dart#L20

maps_inspector.dart is not used test. This change was added only for code sync.

> Just question: Do we need to implement more of this feature?

Yes of course, if it can works on the lwe engine, it should be implemented.

packages/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
JSUYA: https://github.com/flutter-tizen/plugins/pull/816#issuecomment-2649857786

heatmap does not work in the lwe engine and its implementation is not included in this PR.

seungsoo47: Yes. So I was curious about your thoughts on adding test cases along with a later patch since we don't know how long it will take to fix LWE.

packages/google_maps_flutter/example/integration_test/src/maps_inspector.dart
seungsoo47: It seemed to work on LWE, am I wrong?

seungsoo47 (approved):

JSUYA:

JSUYA:
manullay published https://pub.dev/packages/google_maps_flutter_tizen

PR #817, opened 2025-02-11
[video_player_avplay] Automatically rotates video player based on device orientation

JSUYA:
We already provide rotate API for video player,
but it seems like a bug that the video player's video image doesn't rotate according to the device orientation. So we should automatically change this inside video_player. When user call the API, the video image may rotate, but when the device's orientation changes, the rotation may change again.

### DIFF
--- a/packages/video_player_avplay/CHANGELOG.md
+++ b/packages/video_player_avplay/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.16
+
+* Automatically rotates video player based on device orientation.
+
 ## 0.5.15
 
 * Fix dash player fail to seek issue.
--- a/packages/video_player_avplay/README.md
+++ b/packages/video_player_avplay/README.md
@@ -12,7 +12,7 @@ To use this package, add `video_player_avplay` as a dependency in your `pubspec.
 
 ```yaml
 dependencies:
-  video_player_avplay: ^0.5.15
+  video_player_avplay: ^0.5.16
 ```
 
 Then you can import `video_player_avplay` in your Dart code:
--- a/packages/video_player_avplay/lib/video_player.dart
+++ b/packages/video_player_avplay/lib/video_player.dart
@@ -994,12 +994,16 @@ class _VideoPlayerState extends State<VideoPlayer> {
     };
   }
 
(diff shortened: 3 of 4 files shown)


JSUYA:
@xiaowei-guan 
The service team asked me about this issue. They want to handle the device that needs to support portrait mode inside the player, not in Guide. Can you please review this approach?
I'm worried that calling MediaQuery.orientationOf(context) will have a performance impact as build() is called repeatedly while the video player is playing.

xiaowei-guan:
The app side must call setState after initialized:
```dart
_controller.initialize().then((_) => setState()));
```

xiaowei-guan (approved):

JSUYA:
I waiting for https://github.com/flutter-tizen/plugins/pull/814's merge

JSUYA:
Rebased PR

xiaowei-guan:
> @xiaowei-guan The service team asked me about this issue. They want to handle the device that needs to support portrait mode inside the player, not in Guide. Can you please review this approach? I'm worried that calling MediaQuery.orientationOf(context) will have a performance impact as build() is called repeatedly while the video player is playing.

If we want to reduce the call times of MediaQuery.orientationOf, We can also use this：
```
OrientationBuilder(
        builder: (BuildContext context, Orientation orientation)
```

JSUYA:
> > @xiaowei-guan The service team asked me about this issue. They want to handle the device that needs to support portrait mode inside the player, not in Guide. Can you please review this approach? I'm worried that calling MediaQuery.orientationOf(context) will have a performance impact as build() is called repeatedly while the video player is playing.
> 
> If we want to reduce the call times of MediaQuery.orientationOf, We can also use this：
> 
> ```
> OrientationBuilder(
>         builder: (BuildContext context, Orientation orientation)
> ```

When I tested it, the build() calls from calling MediaQuery.orientationOf and OrientationBuilder both had the same call count. This is probably an effect of the prograssbar UI.

When we provided a guide about setDisplayRotate(), we did so by using OrientationBuilder().
However, the Orientation of OrientationBuilder doesn't change because it calculates the Orientation based on the size of the parent Container that contains it. (The ratio of the video screen are still the same.) That's why i had use  MediaQuery.orientationOf.

PR #830, opened 2025-03-14
[video_player_avplay] Add error event

xiaowei-guan:


JSUYA:
I applied this PR and tested to disconnect the network situation when dash video playing. (unpluged cable)
I expected `_controller.value.playerError` of `addListener` to be `connectionFailed`, but `_controller.value.playerError` is always `none`. Is there a specific os version or plusplayer lib version that does not receive the error message? (my device is tizen 9.0)

And the existing hasError/errorDescription and this playerError looks duplicated.
Isn't it better to set errorDescription than to set playerError value? Is it because of the enum type?

xiaowei-guan:
> I applied this PR and tested to disconnect the network situation when dash video playing. (unpluged cable) I expected `_controller.value.playerError` of `addListener` to be `connectionFailed`, but `_controller.value.playerError` is always `none`. Is there a specific os version or plusplayer lib version that does not receive the error message? (my device is tizen 9.0)
> 
> And the existing hasError/errorDescription and this playerError looks duplicated. Isn't it better to set errorDescription than to set playerError value? Is it because of the enum type?

Sorry, I just verified the PR with HSL stream, I also can't get connfication failed with Dash stream, I will check with MM team.

PR #843, opened 2025-03-27
[ci] Change integration test tv target version 8.0 to 9.0

seungsoo47:


.github/workflows/integration_test.yml
```diff
@@ -17,6 +17,11 @@ jobs:
       - uses: subosito/flutter-action@v2
         with:
           channel: stable
+      - name: Install GNOME Packages
+        run: |
+          sudo add-apt-repository ppa:deadsnakes/ppa
```
JSUYA, line 22: Could you remove this line? this for python packages

JSUYA:
Due to the flutter-tizen optimization patch, we need to update tizen-studio on our host pc. The latest tizen-studio releases TV extension 9.0. Therefore, we updated the recipe.

JSUYA (approved):

PR #849, opened 2025-04-21
[integration_test] Add --no-dds option for test

JSUYA:
This option seems to be useful for test environments that don't need access to DDS(Dart Development Service).

https://github.com/flutter-tizen/plugins/actions/runs/14534140709/job/40780256549
```
2025-04-18T11:44:48.5393409Z Bad state: Failed to connect to VM Service at ws://localhost:40951/U0gfMA8WRbw=/ws.
2025-04-18T11:44:48.5395053Z This may happen if DDS is enabled. If this test was launched via `flutter drive`, try adding
2025-04-18T11:44:48.5396346Z `--no-dds`.
```

### DIFF
--- a/tools/lib/src/device.dart
+++ b/tools/lib/src/device.dart
@@ -25,8 +25,8 @@ class Device {
     this.profile, {
     required TizenSdk tizenSdk,
     ProcessRunner processRunner = const ProcessRunner(),
-  })  : _tizenSdk = tizenSdk,
-        _processRunner = processRunner;
+  }) : _tizenSdk = tizenSdk,
+       _processRunner = processRunner;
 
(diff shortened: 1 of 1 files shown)


xiaowei-guan (approved):

seungsoo47:
I have a question. 
So far, there is no issue related to the option.
Why did this problem occur?
Is it because it's version 3.29?

seungsoo47 (approved):

JSUYA:
> I have a question. So far, there is no issue related to the option. Why did this problem occur? Is it because it's version 3.29?

I couldn't find reason why only the test in integration_test fails to connect to dds.

PR #855, opened 2025-05-12
[battery_plus] update version

seungsoo47:


packages/battery_plus/LICENSE
```diff
@@ -10,7 +10,7 @@ are permitted provided that the following conditions are met:
       copyright notice, this list of conditions and the following
       disclaimer in the documentation and/or other materials provided
       with the distribution.
-    * Neither the names of the copyright holders nor the names of the
+    * Neither the name of the copyright holder nor the names of the
```
JSUYA, line 13: This change is to fix a problem where the license of a package was not recognized by pub.dev and was displayed as unknown.
https://pub.dev/packages/battery_plus_tizen/score
[The original text of bsd-3-clause](https://opensource.org/license/bsd-3-clause) is used as is.

JSUYA (approved):

PR #858, opened 2025-05-19
[video_player_videohole] Add suspend and restore api for handing onPause and onResume lifecyle event

gin7773:
### Add suspend and restore interface:

- Plugin native side need to re-create player when power instant off/on.
- If plugin plays a live stream, plugin native side need to re-create player when receiving lifecycle onPause and onResume.
- If plugin plays a vod stream, plugin native side need to call suspend and restore when receiving lifecycle onPause and onResume.

#### Scenario 1: When the video sample in the example is playing/pause, if you turn off the TV power and turn it back on within 5 seconds.
- If the video type is LIVE, the video should start playing/pause from the LIVE point. (Live)
- If you change the video's URL and resume time, the video should playing from the resume time. (ChangeURLTest)
- Else, the video should start playing/pause from the point you were watching. (MP4, HLS, DASH, and others)
#### Scenario 2: When the video sample in the example is playing/pause, if the player preempted by other apps and then returns.
- Same with 'Scenario 1


JSUYA (approved):

xiaowei-guan (approved):

JSUYA:
This plugin was published manually.

PR #871, opened 2025-07-15
[audioplayers] Fix resource cleanup bugs

seungsoo47:
- Handled player resource cleanup in destructor
- Renamed Release() to ReleaseMediaSource() for clearer responsibility
- Handled errors by replacing exceptions in GetDuration() and GetPosition()
- Added logging for unsupported features
- Fixed range limitation of SetVolume() and SetPlaybackRate()

### DIFF
--- a/packages/audioplayers/example/lib/tabs/controls.dart
+++ b/packages/audioplayers/example/lib/tabs/controls.dart
@@ -96,7 +96,7 @@ class _ControlsTabState extends State<ControlsTab>
         ),
         WrappedListTile(
           leading: const Text('Rate'),
-          children: [0.5, 1.0, 2.0].map((it) {
+          children: [0.0, 0.5, 1.0, 2.0].map((it) {
             final formattedVal = it.toStringAsFixed(1);
             return Btn(
               key: Key('control-rate-$formattedVal'),
--- a/packages/audioplayers/tizen/src/audio_player.cc
+++ b/packages/audioplayers/tizen/src/audio_player.cc
@@ -194,11 +194,6 @@ void AudioPlayer::SetVolume(double volume) {
 void AudioPlayer::SetPlaybackRate(double playback_rate) {
   // TODO(seungsoo47): The player_set_playback_rate() API has a limitation of
   // 0.5-2x on TV and is not supported on RPI.
-  if (playback_rate < 0.5) {
-    playback_rate = 0.5;
-  } else if (playback_rate > 2.0) {
-    playback_rate = 2.0;
-  }
   playback_rate_ = playback_rate;
   player_state_e state = GetPlayerState();
   if (state == PLAYER_STATE_READY || state == PLAYER_STATE_PLAYING ||
--- a/packages/audioplayers/README.md
+++ b/packages/audioplayers/README.md
@@ -70,3 +70,4 @@ For detailed information on Tizen privileges, see [Tizen Docs: API Privileges](h
 
 - `onPlayerComplete` event will not be fired when `ReleaseMode` is set to loop which differs from the behavior specified in the [documentation](https://pub.dev/documentation/audioplayers/latest/audioplayers/AudioPlayer/onPlayerComplete.html). And playback rate will reset to 1.0 when audio is replayed.
 - `setVolume` will have no effect on TV devices.
+- `setPlaybackRate` is limited to values between 0.5 and 2.0 on TV and is not supported on RPI.


packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -164,8 +176,15 @@ void AudioPlayer::SetDataSource(std::vector<uint8_t> &data) {
...
+  if (volume > 1) {
+    volume = 1;
+  } else if (volume < 0) {
+    volume = 0;
+  }
   volume_ = volume;
-  if (GetPlayerState() != PLAYER_STATE_NONE) {
+
+  player_state_e state = GetPlayerState();
+  if (state != PLAYER_STATE_NONE) {
```
JSUYA, line 187: This `state` won't be reused so, i don't think it's necessary.
+) 87,220 lines are  also.

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -175,6 +194,13 @@ void AudioPlayer::SetVolume(double volume) {
 }
 
 void AudioPlayer::SetPlaybackRate(double playback_rate) {
+  // TODO(seungsoo47): The player_set_playback_rate() API has a limitation of
+  // 0.5-2x on TV and is not supported on RPI.
+  if (playback_rate < 0.5) {
+    playback_rate = 0.5;
+  } else if (playback_rate > 2.0) {
+    playback_rate = 2.0;
+  }
```
JSUYA, line 203: According to [API guide...](https://docs.tizen.org/iot/api/4.0/tizen-iot-headed/group__CAPI__MEDIA__PLAYER__MODULE.html#gaa2cc9972379f7f2225c6ec2ca1463a25)
It looks like playback_rate should accept values less than 0 and greater than 2.0. In particular, it looks like values less than 0.5 are also possible. Can you confirm?

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -211,31 +239,26 @@ void AudioPlayer::SetLatencyMode(bool low_latency) {
...
 int AudioPlayer::GetDuration() {
+  // TODO(seungsoo47): Get duration in milliseconds, returns -1 if no duration
+  // is available. The frontend package will handle these.
   int32_t duration;
   int ret = player_get_duration(player_, &duration);
   if (ret != PLAYER_ERROR_NONE) {
-    throw AudioPlayerError("player_get_duration failed",
-                           get_error_message(ret));
+    OnLog("player_get_duration failed, " + std::string(get_error_message(ret)));
+    return -1;
```
JSUYA, line 248: um... I don't know what value `player_get_duration()` will return (I don't think millisecond will be less than 0), but is it okay to return -1 even though type is `int *`...?

JSUYA (commented):
and please update CHANGELOG and version.

packages/audioplayers/tizen/src/audio_player.cc
JSUYA: > my test results show that the TV only supports 0.5-2.0, unlike the specifications.

Does your test return PLAYER_ERROR_INVALID_* when the rate value is below 0.5 and above 2.0?

This API is also available in the RPI environment (Common profile), so we cannot consider only the TV case.
If an invalid value is provided, PLAYER_ERROR_INVALID_PARAMETER or PLAYER_ERROR_INVALID_OPERATION will be returned, so this can be confirmed in the error log. 
Therefore, if values between 0.0 and 0.5 do not work, this should be confirmed through the error return value.

According to API guide, values below 0.0 and above 2.0 will be muted. I don't think this means that the API is not working. According to the internal API reference related to the TV API, the range of rates that can be assigned varies depending on the resource type.
Therefore, I don't think it is necessary to limit the minimum and maximum values.

I think the same applies to volume. 
If entering a value outside the range causes a system crash or serious problem, it is possible to limit that value, but if it is already being handled within the platform API, I don't think there is any need to limit it.

What do you think?

packages/audioplayers/example/lib/tabs/controls.dart
```diff
@@ -96,7 +96,7 @@ class _ControlsTabState extends State<ControlsTab>
         ),
         WrappedListTile(
           leading: const Text('Rate'),
-          children: [0.0, 0.5, 1.0, 2.0].map((it) {
+          children: [0.5, 1.0, 2.0].map((it) {
```
JSUYA, line 99: I think it would be better to use the same example as audioplayers.
https://github.com/bluefireteam/audioplayers/blob/main/packages/audioplayers/example/lib/tabs/controls.dart#L104
If this value is limited, I think it should be passed as an error through the platform code.
(There is no need to change it unless there are problems running or building the example, or unless it is platform-dependent code.)

JSUYA (approved):

packages/audioplayers/tizen/src/audio_player.cc
seungsoo47: I'll remove the condition and leave a comment in the README.

PR #875, opened 2025-07-22
[video_player] Fix app crash issue

xiaowei-guan:
Because pause method thorws VideoPlayerError, if call pause method, we muse be add catch the exception, otherwise the process will call terminate.

```
E/ConsoleMessage( 9835): terminate called after throwing an instance of 'VideoPlayerError'
E/ConsoleMessage( 9835): DotNET onSigabrt called on com.yeelight.station_tv_global
E/ConsoleMessage( 9835): DotNET onSigsegv called
```

### DIFF
--- a/packages/video_player/tizen/src/video_player_tizen_plugin.cc
+++ b/packages/video_player/tizen/src/video_player_tizen_plugin.cc
@@ -237,6 +237,7 @@ void VideoPlayerTizenPlugin::SeekTo(
                          [result]() -> void { result(std::nullopt); });
   } catch (const VideoPlayerError &error) {
     result(FlutterError(error.code(), error.message()));
+    return;
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/video_player/tizen/src/video_player_tizen_plugin.cc
```diff
@@ -237,6 +237,7 @@ void VideoPlayerTizenPlugin::SeekTo(
                          [result]() -> void { result(std::nullopt); });
   } catch (const VideoPlayerError &error) {
     result(FlutterError(error.code(), error.message()));
+    return;
```
JSUYA, line 240: i have a question.
SeekTo() ends after catch(). Is there a reason why `return;` is necessary? How is this related to crash?

xiaowei-guan: Sorry, I misunderstand the meaning of try....catch. 
I have added try...catch in onCompleted callback for pause method.

PR #876, opened 2025-07-23
[webview_flutter_lwe] Fix bugs for Tizen 10.0 support

seungsoo47:


packages/webview_flutter_lwe/tizen/src/buffer_pool.h
```diff
@@ -8,6 +8,7 @@
 #include <flutter_texture_registrar.h>
 #include <tbm_surface.h>
 
+#include <cstdio>
```
JSUYA, line 11: `sprintf()` is only used in DEBUG, so it would be better to move it inside `#ifndef NDEBUG~#endif` in buffer_pool.c.

JSUYA:
And please update changelog.

PR #881, opened 2025-08-08
Fix integration test bugs

seungsoo47:


packages/tizen_package_manager/lib/tizen_package_manager.dart
```diff
@@ -18,6 +18,9 @@ enum PackageType {
   /// Tizen web/hybrid application package.
   wgt,
 
+  /// Tizen resource package.
+  rpk,
```
JSUYA, line 22: This is an API change, so you need to update plugin version. probably update to 0.3.0.

packages/tizen_app_manager/example/integration_test/tizen_app_manager_test.dart
```diff
@@ -27,7 +27,11 @@ void main() {
 
     final String appId = await AppManager.currentAppId;
     final AppRunningContext context = AppRunningContext(appId: appId);
-    expect(context.appState, AppState.foreground);
+
+    // TODO(seungsoo47): The context.appState API always returns the appropriate
+    // value. However, we decided to comment out this testcase because the test
+    // farm does not support the 'foreground' state.
+    // expect(context.appState, AppState.foreground);
```
JSUYA, line 34: Could you please update the changelog? (audio_manager too).

PR #889, opened 2025-08-18
[video_player] Temporarily comment out 'asset audio' test case

seungsoo47:
We are building a kind of TCT that runs on Tizen Platform Test Farm. 
So I temporarily remove the problematic test case in order to add integration_test for this plugin. 
I will add the removed test case after the issue is resolved.


packages/video_player/example/integration_test/video_player_test.dart
```diff
@@ -316,67 +316,67 @@ void main() {
...
+  //     expect(controller.value.position, Duration.zero);
+  //     expect(controller.value.isPlaying, false);
+  //     // Due to the duration calculation accuracy between platforms,
+  //     // the milliseconds on Web will be a slightly different from natives.
+  //     // The audio was made with 44100 Hz, 192 Kbps CBR, and 32 bits.
+  //     expect(
+  //       controller.value.duration,
+  //       const Duration(seconds: 5, milliseconds: kIsWeb ? 42 : 41),
+  //     );
+  //   });
```
JSUYA, line 339: You can use `skip` parameter. Please refer to 315 line.

seungsoo47: Done.

packages/video_player/example/integration_test/video_player_test.dart
```diff
@@ -318,65 +318,70 @@ void main() {
...
-      expect(controller.value.position, pausedPosition);
-    });
-  });
+        // Verify that we stopped playing after the pause.
+        expect(controller.value.isPlaying, false);
+        expect(controller.value.position, pausedPosition);
+      });
+    },
+    // NOTE(seungsoo47): Unknown errors sometimes occur in RPI.
+    skip: Platform.environment['ELM_PROFILE'] == 'common',
```
JSUYA, line 385: Could you use isTizenProfile?
https://github.com/flutter-tizen/flutter-tizen/commit/94c3447bafc35b91d16c336fb4dc1c4a52e7d775

PR #891, opened 2025-08-18
Fix ci build bug

seungsoo47:


JSUYA (approved):
CI fails : https://github.com/flutter-tizen/plugins/actions/runs/17040191537/job/48301820363

Our CI always uses the latest Flutter version by default.
This PR may be reverted when flutter-tizen supports Flutter 3.35.
https://github.com/flutter-tizen/flutter-tizen/issues/657

PR #906, opened 2025-09-19
Update whether plugins work on RPI

seungsoo47:


seungsoo47:
> I think we can remove `Watch` and `Watch emulator` from this PR. you can also remove `wearable_rotary` from the list.

I have removed the columns. And as we discussed offline, later let's discuss in more details for how we should handle plugins that don't work properly on Tizen devices and emulators.

seungsoo47:
[[camera_tizen] camera plugin has some issues](https://github.com/flutter-tizen/plugins/issues/907)
[[image_picker_tizen] Plugin doesn't work](https://github.com/flutter-tizen/plugins/issues/908)

JSUYA (approved):

Issue #820, opened 2025-02-19
[video_player_avplay] Add missing APIs


xiaowei-guan:
Refer to [Tizen avplay](https://developer.samsung.com/smarttv/develop/api-references/samsung-product-api-references/avplay-api.html), add missing APIs.

JSUYA:
Could you consider adding a [`bool Suspend(PlusplayerRef player);`](https://github.com/flutter-tizen/plugins/blob/master/packages/video_player_avplay/tizen/inc/plusplayer/plusplayer_wrapper.h#L464) 

It looks like a similar spec to[ suspend() in WebApp API.](https://developer.samsung.com/smarttv/develop/api-references/samsung-product-api-references/avplay-api.html#AVPlayManager-suspend)
If it is the same spec, I think we can utilize this in the pause/resume scenario of AppLifecycle.

https://github.com/flutter-tizen/plugins/blob/a9ecdd07e29787cca9d75d87aa678241fd07e330/packages/video_player_avplay/lib/video_player.dart#L976-L985

+) Of course, if it doesn't meet the specifications we expect, it doesn't necessarily need to be added.

xiaowei-guan:
> Could you consider adding a [`bool Suspend(PlusplayerRef player);`](https://github.com/flutter-tizen/plugins/blob/master/packages/video_player_avplay/tizen/inc/plusplayer/plusplayer_wrapper.h#L464)
> 
> It looks like a similar spec to[ suspend() in WebApp API.](https://developer.samsung.com/smarttv/develop/api-references/samsung-product-api-references/avplay-api.html#AVPlayManager-suspend) If it is the same spec, I think we can utilize this in the pause/resume scenario of AppLifecycle.
> 
> [plugins/packages/video_player_avplay/lib/video_player.dart](https://github.com/flutter-tizen/plugins/blob/a9ecdd07e29787cca9d75d87aa678241fd07e330/packages/video_player_avplay/lib/video_player.dart#L976-L985)
> 
> Lines 976 to 985 in [a9ecdd0](/flutter-tizen/plugins/commit/a9ecdd07e29787cca9d75d87aa678241fd07e330)
> 
>  void didChangeAppLifecycleState(AppLifecycleState state) { 
>    if (state == AppLifecycleState.paused) { 
>      _wasPlayingBeforePause = _controller.value.isPlaying; 
>      _controller.pause(); 
>    } else if (state == AppLifecycleState.resumed) { 
[...]

PR #909, opened 2025-09-24
[video_player_avplay] Add new features for DASH

gin7773:
Main changes:
* Add new features for DASH video player
  1. Support update token before and after player prepare done.
  2. Support setting resolution, frame rate and whether to update same language code.
  3. Support obtaining streaming information such as audio, video, and subtitle.
  4. Support getting the http response header of the requested url.
  5. Support capturing ad information in the stream.


xiaowei-guan:
https://github.com/flutter-tizen/plugins/issues/820

packages/video_player_avplay/example/lib/main.dart
```diff
@@ -127,23 +127,49 @@ class _DashRomoteVideo extends StatefulWidget {
...
+      // update token [after] dash-player prepare done.
+      _controller.updateToken('YWJyVHlwZT1CUi1BVkMtREFTSC');
+      // New features: get the following properties.
+      _controller.getStreamingProperty(StreamingPropertyType.audioStreamInfo);
+      _controller.getStreamingProperty(
+        StreamingPropertyType.subtitleStreamInfo,
+      );
+      _controller.getStreamingProperty(StreamingPropertyType.videoStreamInfo);
+      _controller.getData(<DashPlayerProperty>{DashPlayerProperty.httpHeader});
+    });
```
JSUYA, line 172: Is it possible to retrieve the result of adInfo when this property related to this AD is applied to the current resource? (is that token string valid?)
Is there a suitable resource?
I understand this feature is intended for CP, but I'd appreciate a more detailed explanation for future users.
If you have a tight schedule, you can write a description of this feature in another PR.

packages/video_player_avplay/lib/video_player.dart
```diff
@@ -973,6 +992,15 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
   }
 
+  /// Update token of DASH at any time after initialization is complete.
+  Future<bool> updateToken(String token) async {
+    if (_isDisposedOrNotInitialized) {
+      return false;
+    }
+
+    return _videoPlayerPlatform.updateToken(playerId, token);
+  }
```
JSUYA, line 1002: When I set UpdateDashToken() when the current resource type or format is not "dash" the following error will continue to occur through listener.
Since PlusPlayer's API explicitly states that it's Dash-only(`::UpdateDashToken(player_, token);`),
I think the interface should also indicate that it's a Dash-only API.
(And if it's not a dash, it can output related exception or error messages.)

```
[E] [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PlatformException(channel-error, Unable to establish connection on channel: "dev.flutter.pigeon.video_player_avplay.VideoPlayerAvplayApi.updateToken"., null, null)
#0      VideoPlayerAvplayApi.updateToken (package:video_player_avplay/src/messages.g.dart:1782:7)
<asynchronous suspension>
```

packages/video_player_avplay/README.md
```diff
@@ -12,7 +12,7 @@ To use this package, add `video_player_avplay` as a dependency in your `pubspec.
 
 ```yaml
 dependencies:
-  video_player_avplay: ^0.6.0
+  video_player_avplay: ^0.6.1
```
JSUYA, line 15: Since this pr is a feature addition, 0.7.0 seems more appropriate.

gin7773: Done.

packages/video_player_avplay/example/lib/main.dart
gin7773: The detailed description has been added.

Because we don't have a dedicated media server, we can only test using URLs from third-party and internal servers, which cannot be uploaded. The URL in the sample code is not compatible, so this can only be used as an example(adInfo and update token), and the results cannot be seen at now.

We are still looking for a suitable resource for AD. If so, I will submit it next time.

packages/video_player_avplay/example/lib/main.dart
```diff
@@ -127,23 +127,49 @@ class _DashRomoteVideo extends StatefulWidget {
...
+      setState(() {});
+      // update token [after] dash-player prepare done.
+      _controller.updateDashToken('YWJyVHlwZT1CUi1BVkMtREFTSC');
+      // New features: get the following properties.
+      _controller.getStreamingProperty(StreamingPropertyType.audioStreamInfo);
+      _controller.getStreamingProperty(
+        StreamingPropertyType.subtitleStreamInfo,
+      );
+      _controller.getStreamingProperty(StreamingPropertyType.videoStreamInfo);
+      _controller.getData(<DashPlayerProperty>{DashPlayerProperty.httpHeader});
```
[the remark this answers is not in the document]
gin7773: Done.

packages/video_player_avplay/example/lib/main.dart
```diff
@@ -127,23 +127,49 @@ class _DashRomoteVideo extends StatefulWidget {
...
-    _controller.initialize().then((_) => setState(() {}));
+    _controller.initialize().then((_) {
+      setState(() {});
+      // update token [after] dash-player prepare done.
+      _controller.updateDashToken('YWJyVHlwZT1CUi1BVkMtREFTSC');
+      // New features: get the following properties.
+      _controller.getStreamingProperty(StreamingPropertyType.audioStreamInfo);
+      _controller.getStreamingProperty(
+        StreamingPropertyType.subtitleStreamInfo,
+      );
```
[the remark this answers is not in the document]
gin7773: Done.

packages/video_player_avplay/example/lib/main.dart
```diff
@@ -127,23 +127,49 @@ class _DashRomoteVideo extends StatefulWidget {
...
+  // Optional
+  final Map<StreamingPropertyType, String> _streamingProperties =
+      <StreamingPropertyType, String>{
+    StreamingPropertyType.unwantedResolution: '3840X2160',
+    StreamingPropertyType.unwantedFramerate: '60',
+    StreamingPropertyType.updateSameLanguageCode: '1',
+
+    /// update token [before] dash-player prepare done.
+    StreamingPropertyType.dashToken: 'YWJyVHlwZT1CUi1BVkMtREFTSC',
+    StreamingPropertyType.openHttpHeader: 'TRUE',
```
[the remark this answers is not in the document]
gin7773: If use the ‘_streamingProperties’ setting, there will be no prompt on the user side, but the plugin log will show a prompt ‘[PlusPlayer] Only support streaming property type: %s for DASH!’.

If use the 'setStreamingProperty()' API setting, there will throw an error.

xiaowei-guan (approved):

PR #912, opened 2025-09-29
[video_player_avplay] Fix type missmatch

JSUYA:
fix to below error on tizen 10.0 rootstrap

```
src/plus_player.cc:1366:9: error: cannot initialize a variable of type 'char *' with an rvalue of type 'const char *'
  char *data = strstr(ad_data, prefix);
        ^      ~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/packages/video_player_avplay/CHANGELOG.md
+++ b/packages/video_player_avplay/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.7.2
+* Fix type missmatch.
+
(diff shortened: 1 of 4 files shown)


seungsoo47:
> Sorry for my mistake, I have verified it on 10.0. Regarding the `ad_data` content, I am confirming it with MM.

Oh I see.
I will approve this commit.

Issue #918, opened 2025-10-17
[tizen_package_manager] Support installed app size data


JSUYA:
```cpp
int package_manager_get_package_size_info(const char *package_id, package_manager_size_info_receive_cb callback, void *user_data);
int package_size_info_get_app_size(package_size_info_h handle, long long *app_size);
```
We can use these two APIs to obtain the installed app size.
This can help users display the app size.

PR #919, opened 2025-10-17
[tizen_package_manager] Add retrieve package size information

JSUYA:
* Add retrieve package size information using `PackageManager.getPackageSizeInfo`.

related issue: https://github.com/flutter-tizen/plugins/issues/918


JSUYA:
+) @shyunMin

Issue #916, opened 2025-10-15
New plugin supporting the Window API


JSUYA:
We may need to provide a Dart API for controlling Windows.
(We are already providing a way to control the window using the internal method channel. https://github.com/flutter-tizen/embedder/pull/122)

However, some plugins already obtain the window native handle, open to libecore_wl2.so using dlopen, and make EcoreWl2WindowProxy.
https://github.com/flutter-tizen/plugins/blob/master/packages/video_player_avplay/tizen/src/ecore_wl2_window_proxy.h

Therefore, we can provide APIs as plugins for several APIs.

```
ecore_wl2_window_show
ecore_wl2_window_hide

ecore_wl2_window_activate
ecore_wl2_window_raise
ecore_wl2_window_lower

ecore_wl2_window_alpha_get
ecore_wl2_window_alpha_set

ecore_wl2_window_title_set   // ??
ecore_wl2_window_title_get  //??
```

If you have any opinions, please feel free.

xiaowei-guan:
@JSUYA Do you need me to implement it?

JSUYA:
> [@JSUYA](https://github.com/JSUYA) Do you need me to implement it?

No, thank you. I'll do it when I have time later. It's not urgent.

PR #922, opened 2025-10-27
Introduce tizen_window_manager plugin

JSUYA:
https://github.com/flutter-tizen/plugins/issues/916

~~This plugin controls windows and obtains geometry information. It only works when the RendererType is EGL.~~
(EvasGL has been removed.)

### DIFF
--- a/packages/tizen_window_manager/tizen/src/tizen_window_manager_plugin.cc
+++ b/packages/tizen_window_manager/tizen/src/tizen_window_manager_plugin.cc
@@ -36,10 +36,9 @@ class TizenWindowManagerPlugin : public flutter::Plugin {
     plugin_registrar->AddPlugin(std::move(plugin));
   }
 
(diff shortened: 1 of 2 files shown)


seungsoo47 (commented):
a minor issue

packages/tizen_window_manager/tizen/src/tizen_window_manager_plugin.cc
```diff
@@ -0,0 +1,88 @@
...
+      window_manager_->Lower();
+      result->Success();
+    } else if (method_name == "getGeometry") {
+      result->Success(flutter::EncodableValue(window_manager_->GetGeometry()));
+    } else {
+      result->NotImplemented();
+    }
+  }
+
+  FlutterDesktopPluginRegistrarRef registrar_ref_;
```
xiaowei-guan: I think we can remove registrar_ref_, because the plugin just uses it onw time.

JSUYA: done

PR #942, opened 2025-11-28
[video_player_avplay] Fix black line issue when playing the video

xiaowei-guan:
Because the native player api SetDisplayRoi only support int type:
```
void SetDisplayRoi(int32_t x, int32_t y, int32_t width, int32_t height)
```
But flutter widget widget postion type is double, There may be offset during type conversion, we need to make the video player ROI size slightly larger.

### DIFF
--- a/packages/video_player_avplay/CHANGELOG.md
+++ b/packages/video_player_avplay/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.7
+
+* Fix black line issue when playing video.
+
 ## 0.7.6
 * Fix video doesn't scale issue.
 
--- a/packages/video_player_avplay/README.md
+++ b/packages/video_player_avplay/README.md
@@ -12,7 +12,7 @@ To use this package, add `video_player_avplay` as a dependency in your `pubspec.
 
 ```yaml
 dependencies:
-  video_player_avplay: ^0.7.6
+  video_player_avplay: ^0.7.7
 ```
 
 Then you can import `video_player_avplay` in your Dart code:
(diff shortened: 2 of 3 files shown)


gin7773 (approved):

packages/video_player_avplay/lib/video_player.dart
```diff
@@ -1246,26 +1246,29 @@ class _VideoPlayerState extends State<VideoPlayer> {
...
+        final double offsetWidth = rect.width.ceil() - rect.width;
+        final double offsetHeight = rect.height.ceil() - rect.height;
+        final int left = _isInvalid(rect.left) ? 0 : rect.left.floor();
+        final int top = _isInvalid(rect.top) ? 0 : rect.top.floor();
+        final int width = _isInvalid(rect.width)
+            ? 1
+            : rect.width.ceil() + ((offsetLeft > offsetWidth) ? 1 : 0);
+        final int height = _isInvalid(rect.height)
+            ? 1
+            : rect.height.ceil() + ((offsetTop > offsetHeight) ? 1 : 0);
```
JSUYA: Is this logic the same as PlusPlayer's surface logic ?
In some cases, it might be 1 pixel larger.
umm.. wouldn't it be more appropriate to use just `round() +- 1`? 


```dart
void main() {
  double left =   9.999999999;   // 10
  double width = 20.00000001;  // 20

  final double offsetLeft =
      left - left.floor(); // 9.9999 - 9 = 0.9999...
  final double offsetWidth =
      width.ceil() - width; // 21 - 20.00001 = 0.9999...

  var final_left = left.floor(); // 9
  var final_width = width.ceil() + ((offsetLeft > offsetWidth) ? 1 : 0); // 21 + 0 or 1
  
  print(left);
  print(width);
  
  print(final_left);
  print(final_width);
}
```

PR #1009, opened 2026-04-24
[webview_flutter_tizen] Update webview_flutter to 4.13.1

seungsoo47:
* Update minimum supported SDK version to Flutter 3.32/Dart 3.8.
* Update webview_flutter to 4.13.1.
* Update webview_flutter_platform_interface to 2.13.0.
* Add support for setVerticalScrollBarEnabled, setHorizontalScrollBarEnabled, supportsSetScrollBarsEnabled APIs

https://github.com/flutter-tizen/plugins/issues/982

### DIFF
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -706,6 +706,8 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
     const auto* value = std::get_if<bool>(arguments);
     if (value) {
       scrollbar_enabled_ = *value;
+      EwkInternalApiBinding::GetInstance().view.MainFrameScrollbarVisibleSet(
+          webview_instance_, scrollbar_enabled_);
       result->Success();
     } else {
       result->Error("Invalid argument", "The argument must be a bool.");
--- a/packages/webview_flutter/tizen/src/webview.cc
+++ b/packages/webview_flutter/tizen/src/webview.cc
@@ -701,6 +701,15 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
     } else {
       result->Error("Invalid argument", "The argument must be a string.");
     }
+  } else if (method_name == "setVerticalScrollBarEnabled" ||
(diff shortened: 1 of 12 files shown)


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -773,6 +784,9 @@ void WebView::OnFrameRendered(void* data, Evas_Object* obj, void* event_info) {
 
 void WebView::OnLoadStarted(void* data, Evas_Object* obj, void* event_info) {
   WebView* webview = static_cast<WebView*>(data);
+  EwkInternalApiBinding::GetInstance().view.MainFrameScrollbarVisibleSet(
+      webview->webview_instance_, webview->scrollbar_enabled_);
```
JSUYA, line 788: Should this API always be called every time the page is reloaded?
You are already calling the set API on line 709.

seungsoo47: This API should be called only after the webview is loaded.
After that,  the API call on line 709 is intended to immediately reflect the requested state of the UI.

JSUYA: Does the setting get reset when the page reloads?

If not, I think it would be better to use `std:optional<bool>` to call the function only when the setting is configured, and to ensure it is not called again once it has already been invoked.

While this code will likely have little impact on performance or behavior, if you want to be clearer, it might be better to treat this as an API constraint and display a warning message if it is called before the page loads.

JSUYA (approved):

packages/webview_flutter/tizen/src/webview.cc
seungsoo47: Thank you for your review.

When I tried it, the `std::optional<bool>` variable seems to work correctly only on the first attempt.
In other words, its has_value() function always returns true from the second call onwards.
Therefore, I think it would be better to keep the current code as is.

And this API on the Flutter side can be called while the WebViewController is set up or after it is set up.

Thank you again for your review.

seungsoo47:
https://github.com/flutter-tizen/plugins/pull/1009

PR #1036, opened 2026-06-16
[audioplayers] Update integration tests to upstream v6.6.0

seungsoo47:
Updates the audioplayers integration tests to match upstream v6.6.0, and fixes
the plugin issues surfaced while validating them on device (Tizen TV / RPI4).

### Integration tests
- Port the upstream v6.6.0 suite (`lib_test` + `platform_test`) into a single
`audioplayers_test.dart`, keeping only Tizen-runnable cases.
- Restrict data-driven tests to local asset sources, since remote URL/stream
playback does not emit reliable complete/position events on Tizen.
- Drop cases unsupported on Tizen (bytes source, AudioContext, LOW_LATENCY
semantics, `setBalance`, invalid-file failure event).

### Plugin fixes
- Return `null` for duration/position when no source is prepared (e.g. after
`release()`), matching other platforms.
- Fix a crash where `Stop()` → `Seek(0)` threw `Invalid state` for network
sources; the position reset after stop is now best-effort.
- Emit `OnError` / `OnInterrupted` on the platform thread to avoid the
"channel sent a message ... on a non-platform thread" warning.

Tested on Tizen TV (10.0) and Raspberry Pi 4: all integration tests pass
(23 passed, 0 skipped, 0 failed).

#1039


packages/audioplayers/example/integration_test/lib/lib_source_test_data.dart
```diff
@@ -0,0 +1,141 @@
...
+      // Cache not working for web
+      LibSourceTestData(
+        source: BytesSource(
+          await AudioCache.instance.loadAsBytes(wavAsset2),
+          mimeType: 'audio/wav',
+        ),
+        duration: const Duration(seconds: 1, milliseconds: 068),
+      ),*/
+  ];
+}
```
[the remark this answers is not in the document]
seungsoo47: On Tizen, hasBytesSource is false so mp3BytesTestData() is never called - no download occurs.
I will not modify the code to maintain upstream code.

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -101,6 +104,15 @@ void AudioPlayer::Stop() {
...
     }
+    // Reset the play position to 0 to match other platforms. This is
+    // best-effort: on some devices (e.g. TV with network sources)
+    // player_set_play_position right after stop can fail with an invalid
+    // state, which must not crash the app.
+    try {
+      Seek(0);
+    } catch (const AudioPlayerError &error) {
+      OnLog("Failed to reset position on stop: " + error.message());
+    }
```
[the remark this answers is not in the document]
seungsoo47: This is code to maintain the INVALID_STATE state, which only occurs under certain conditions (network TV).

packages/audioplayers/example/integration_test/audioplayers_test.dart
```diff
@@ -1,188 +1,601 @@
...
+    },
+    skip: kIsWeb,
+  );
 
-      final position = await player.getCurrentPosition();
-      expect(duration, isNotNull);
-      expect(position!.inMilliseconds, 0);
+  testWidgets(
+    'test url source with no extension',
+    (WidgetTester tester) async {
```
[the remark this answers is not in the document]
seungsoo47: This is from upstream code.

packages/audioplayers/example/integration_test/audioplayers_test.dart
```diff
@@ -1,188 +1,601 @@
...
+        eventStream,
+        emitsThrough(
+          const AudioEvent(
+            eventType: AudioEventType.log,
+            logMessage: 'SomeLog',
+          ),
+        ),
+      );
+      await platform.emitLog(playerId, 'SomeLog');
+    });
```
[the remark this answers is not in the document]
seungsoo47: This is from upstream code.

packages/audioplayers/tizen/src/audio_player.h
```diff
@@ -54,6 +54,10 @@ class AudioPlayer {
   int32_t GetCurrentPosition();
   std::string GetPlayerId() const { return player_id_; }
   bool IsPlaying();
+  // Whether a source is prepared, i.e. duration and position are meaningful.
+  // After release (or before a source is prepared) this is false, so the
+  // plugin returns null for duration/position to match other platforms.
+  bool IsSourcePrepared();
```
JSUYA: The function's internal code and name sufficiently explain its behavior, so automatically generated comments are not necessary. Please delete any automatically generated comments unnecessary.

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -419,17 +437,31 @@ void AudioPlayer::OnPlayCompleted(void *data) {
...
+  ecore_main_loop_thread_safe_call_async(
+      [](void *data) {
+        auto *context =
+            reinterpret_cast<std::pair<AudioPlayer *, std::string> *>(data);
+        AudioPlayer *player = context->first;
+        player->log_listener_(player->player_id_, context->second);
+        delete context;
+      },
+      context);
 }
```
JSUYA, line 517: please rebase pr

packages/audioplayers/tizen/src/audio_player.cc
```diff
@@ -101,6 +104,15 @@ void AudioPlayer::Stop() {
     if (ret != PLAYER_ERROR_NONE) {
       throw AudioPlayerError("player_stop failed", get_error_message(ret));
     }
+    // Reset the play position to 0 to match other platforms. This is
+    // best-effort: on some devices (e.g. TV with network sources)
+    // player_set_play_position right after stop can fail with an invalid
+    // state, which must not crash the app.
+    try {
+      Seek(0);
```
JSUYA, line 124: This affects existing behavior. I think, if it is a platform-dependent issue, it should be handled within the player_stop() API itself.

packages/audioplayers/example/integration_test/platform_features.dart
```diff
@@ -0,0 +1,162 @@
...
+  defaultValue: true,
+);
+
+const testIsAndroidMediaPlayer = bool.fromEnvironment(
+  'TEST_ANDROID_MEDIAPLAYER',
+);
+
+/// Specify supported features for a platform.
+class PlatformFeatures {
+  static const webPlatformFeatures = PlatformFeatures(
```
JSUYA: We do not need to have test code related to other platforms within our test code.

packages/audioplayers/example/integration_test/platform_features.dart
```diff
@@ -0,0 +1,162 @@
...
+            : defaultTargetPlatform == TargetPlatform.iOS
+                ? iosPlatformFeatures
+                : defaultTargetPlatform == TargetPlatform.macOS
+                    ? macPlatformFeatures
+                    : defaultTargetPlatform == TargetPlatform.linux
+                        ? linuxPlatformFeatures
+                        : defaultTargetPlatform == TargetPlatform.windows
+                            ? windowsPlatformFeatures
+                            : const PlatformFeatures();
+  }
```
JSUYA: It seems it will return Linux here. You need to check if the support range is correct. For example, `hasBalance`

packages/audioplayers/tizen/src/audio_player.cc
seungsoo47: Could you clarify which existing behavior you're concerned about? 
This is an intentional behavioral fix to match the audioplayers platform interface, not a Tizen-specific workaround. The audioplayers documentation specifies that after stop(), the position should reset to 0. Other platform implementations (Android, iOS) handle this automatically at the OS level, but Tizen's player_stop() does not reset the position, so we do it explicitly here.
Moving this into a lower-level API is not feasible since player_stop() is a Tizen system API we don't control. 
If you'd prefer, we can add a comment at the caller clarifying that this is a platform interface requirement rather than a Tizen-internal detail.

JSUYA: Ok, I understand.
Could you provide a link to a document containing the relevant guide? I couldn't find it.
If you have a shareable link to the document, I recommend writing it in the comments.

packages/audioplayers/example/integration_test/platform_features.dart
```diff
@@ -0,0 +1,56 @@
...
+    this.hasForceSpeaker = true,
+    this.hasDuckAudio = true,
+    this.hasRespectSilence = true,
+    this.hasStayAwake = true,
+    this.hasRecordingActive = true,
+    this.hasPlayingRoute = true,
+    this.hasDurationEvent = true,
+    this.hasPlayerStateEvent = true,
+    this.hasErrorEvent = true,
+  });
```
JSUYA, line 53: It looks that all features are currently in an opened.
For unsupported features (e.g., setBalance/hasBalance), you must explicitly mark them as false, and the actual test should not run.

JSUYA (approved):

Issue #1039, opened 2026-06-17
Review and update integration tests for Tizen plugins


seungsoo47:
## Background / Purpose

  The integration tests of several Tizen plugins have drifted from their
  upstream counterparts or only cover a thin slice of the API. This issue tracks
  a systematic pass to bring each testable plugin's integration tests up to date,
  so that they reflect the current behavior and are actually runnable on Tizen
  devices/emulators.

  ## Scope (per plugin)

  - Identify the upstream package and the exact version the Tizen plugin targets.
  - For plugins that wrap an upstream package: port the test cases from that
    upstream version that can actually run on Tizen, excluding platform-specific
    cases (Android/iOS/macOS-gated) that don't apply, and consolidate them into
    the single `example/integration_test/<plugin>_test.dart`.
  - For Tizen-specific plugins (or where upstream tests don't apply): add
    regression tests over the public API.
  - Validate on a Tizen device and/or emulator.

  ## Target plugins (testable in `.github/recipe.yaml`)

  - [x] audioplayers — #1036
  - [x] connectivity_plus — already covered; no Tizen-runnable cases to add
  - [x] device_info_plus — #1038 
  - [x] flutter_secure_storage — #1040 
  - [x] flutter_tts — #1041 
  - [x] integration_test — already covered; no Tizen-runnable cases to add
  - [x] keyboard_detection — #1042 
  - [x] messageport — #1043 
  - [x] package_info_plus — already covered; no Tizen-runnable cases to add
  - [x] path_provider — already covered; no Tizen-runnable cases to add
  - [x] shared_preferences — #1047 
  - [x] sqflite — already covered; no Tizen-runnable cases to add
  - [x] tizen_app_manager — #1049 
  - [x] tizen_audio_manager — #1050 
  - [x] tizen_bundle — #1051 
  - [x] tizen_package_manager — #1055 
  - [x] tizen_rpc_port — #1054 
  - [x] tizen_window_manager — #1053 
  - [x] url_launcher — already covered; no Tizen-runnable cases to add
  - [x] wakelock_plus — already covered; no Tizen-runnable cases to add

## Other plugins (in `.github/recipe.yaml`)
  - [ ] permission_handler
  - [ ] google_sign_in
  - [ ] firebase_core
  - [ ] tizen_log
  - [ ] tizen_notification
  - [ ] flutter_webrtc
  - [ ] geolocator
  - [ ] in_app_purchase
  - [ ] network_info_plus
[...]

PR #1041, opened 2026-06-17
[flutter_tts] Add integration tests for the supported TTS APIs

seungsoo47:
flutter_tts has no integration tests upstream, so add regression tests covering
the Tizen-supported parts of the flutter_tts v4.2.5 API (per the plugin's
"Supported APIs"):

- speak / stop / pause return 1
- getLanguages returns a non-empty list
- isLanguageAvailable is true for a supported language
- setLanguage returns 1 for a supported language
- getVoices returns a non-empty list
- getDefaultVoice returns a voice
- setVoice returns 1
- setSpeechRate returns 1
- setVolume returns 1
- getMaxSpeechInputLength is positive (calls stop() first so the engine is in
  the ready state that tts_get_max_text_size requires)

The isLanguageAvailable test is fixed to return a boolean value.

Validated on a Tizen TV (10.0) and Raspberry Pi 4: all tests pass
(12 passed, 0 skipped, 0 failed).

#1039

### DIFF
--- a/packages/flutter_tts/example/integration_test/flutter_tts_test.dart
+++ b/packages/flutter_tts/example/integration_test/flutter_tts_test.dart
@@ -22,13 +22,7 @@ void main() {
   });
 
   testWidgets('stop returns 1', (WidgetTester tester) async {
-    await flutterTts.speak('Hello, world!');
     expect(await flutterTts.stop(), 1);
-  });
-
-  testWidgets('pause returns 1', (WidgetTester tester) async {
-    await flutterTts.speak('Hello, world!');
-    expect(await flutterTts.pause(), 1);
   });
 
   testWidgets('getLanguages returns a non-empty list', (
--- a/packages/flutter_tts/example/integration_test/flutter_tts_test.dart
+++ b/packages/flutter_tts/example/integration_test/flutter_tts_test.dart
@@ -7,10 +7,14 @@ void main() {
 
   late FlutterTts flutterTts;
 
-  setUp(() async {
+  setUpAll(() async {
     flutterTts = FlutterTts();
     // Give the native TTS engine time to initialize.
     await Future<void>.delayed(const Duration(seconds: 2));
+  });
+
+  tearDown(() async {
+    await flutterTts.stop();
   });
 
   testWidgets('speak returns 1', (WidgetTester tester) async {
@@ -39,6 +43,7 @@ void main() {
     WidgetTester tester,
   ) async {
     final languages = (await flutterTts.getLanguages as List).cast<String>();
+    expect(languages, isNotEmpty);
     expect(await flutterTts.isLanguageAvailable(languages.first), isTrue);
   });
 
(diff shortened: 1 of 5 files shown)


packages/flutter_tts/example/integration_test/flutter_tts_test.dart
```diff
@@ -5,10 +5,96 @@ import 'package:integration_test/integration_test.dart';
...
+
+  testWidgets('stop returns 1', (WidgetTester tester) async {
+    await flutterTts.speak('Hello, world!');
+    expect(await flutterTts.stop(), 1);
+  });
+
+  testWidgets('pause returns 1', (WidgetTester tester) async {
+    await flutterTts.speak('Hello, world!');
+    expect(await flutterTts.pause(), 1);
+  });
```
JSUYA: After calling tts_speak(), tts_stop and tts_pause are called immediately in c-api code.
If a timing issue occurs and tts_speak is delayed while in the TTS_STATE_READY state, it is unlikely that it will always return 1.

packages/flutter_tts/example/integration_test/flutter_tts_test.dart
```diff
@@ -5,10 +5,96 @@ import 'package:integration_test/integration_test.dart';
...
+    expect(languages, isNotEmpty);
+    expect(await flutterTts.setLanguage(languages.first), 1);
+  });
+
+  testWidgets('getVoices returns a non-empty list', (
+    WidgetTester tester,
+  ) async {
+    final voices = await flutterTts.getVoices;
+    expect(voices, isNotNull);
+    expect(voices, isNotEmpty);
```
JSUYA, line 57: I'm just wondering, I think `getVoices` might not be available depending on the environment. (Currently, our target always has voice, but there could be multiple building blocks.) Is this test case valid? Usually, in cases like this, wouldn't you test by mocking `getVoices`?

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
```diff
@@ -298,11 +298,11 @@ class FlutterTtsTizenPlugin : public flutter::Plugin {
     if (std::holds_alternative<std::string>(arguments)) {
       std::string language = std::move(std::get<std::string>(arguments));
       if (!language.empty() && tts_->IsLanguageAvailable(language)) {
-        SendResult(flutter::EncodableValue(1));
+        SendResult(flutter::EncodableValue(true));
         return;
       }
     }
-    SendResult(flutter::EncodableValue(0));
+    SendResult(flutter::EncodableValue(false));
```
JSUYA, line 305: Since this part is a change to the interface (`Future<int>` -> `Future<Bool>`), I think the version needs to be changed to 1.7.0.

packages/flutter_tts/example/integration_test/flutter_tts_test.dart
seungsoo47: Integration tests run against real devices or specific emulators where the TTS engine is always present. If 'getVoices' returns empty, it means the device or emulator is not properly configured for TTS, which is a valid test failure.

packages/flutter_tts/tizen/src/flutter_tts_tizen_plugin.cc
seungsoo47: Done.

JSUYA (approved):

PR #1042, opened 2026-06-18
[keyboard_detection] Add regression integration tests

seungsoo47:
Expand the integration test suite for the keyboard detection controller from 2 to 20 cases, driving the input-panel event channel with simulated platform messages so the tests are deterministic and need no real keyboard:

- Initial unknown state and unknown fallback for unrecognized events.
- All keyboard states: visibling, visible, hiding, hidden.
- stateAsBool mapping, including the transitional states with and without includeTransitionalState.
- Keyboard metrics (width, height/size, position) updates on show with dimensions, reset on hide, and isSizeLoaded / ensureSizeLoaded.
- Robustness against malformed events (non-map / missing or wrong-typed state).
- Notifications via the state stream and the onChanged callback.
- Registered callbacks: delivery, self-unregister on returning false, unregisterCallback and unregisterAllCallbacks.
- No further state changes are reported after dispose.

This is a Tizen-exclusive plugin with no upstream package. Test-only change, recorded under a NEXT changelog entry without a version bump. Validated on a Raspberry Pi 4 (Tizen) device: 20 passed, 0 failed.

### DIFF
--- a/packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
+++ b/packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
@@ -42,6 +42,8 @@ void main() {
       await emit(tester, <String, Object?>{'state': 'will_show'});
       await tester.pump();
       expect(controller.state, KeyboardState.visibling);
+      expect(controller.stateAsBool(), isFalse);
+      expect(controller.stateAsBool(true), isTrue);
       await controller.dispose();
     });
 
@@ -64,6 +66,8 @@ void main() {
       await emit(tester, <String, Object?>{'state': 'will_hide'});
       await tester.pump();
       expect(controller.state, KeyboardState.hiding);
+      expect(controller.stateAsBool(), isTrue);
+      expect(controller.stateAsBool(true), isFalse);
       await controller.dispose();
     });
 
@@ -107,30 +111,6 @@ void main() {
       await emit(tester, <String, Object?>{'state': 123});
       await tester.pump();
       expect(controller.state, KeyboardState.visible);
-      await controller.dispose();
-    });
-  });
-
-  group('stateAsBool', () {
-    testWidgets('maps the transitional visibling state',
-        (WidgetTester tester) async {
-      final KeyboardDetectionController controller =
-          KeyboardDetectionController();
-      await emit(tester, <String, Object?>{'state': 'will_show'});
-      await tester.pump();
-      expect(controller.stateAsBool(), isFalse);
-      expect(controller.stateAsBool(true), isTrue);
-      await controller.dispose();
-    });
-
-    testWidgets('maps the transitional hiding state',
-        (WidgetTester tester) async {
-      final KeyboardDetectionController controller =
-          KeyboardDetectionController();
-      await emit(tester, <String, Object?>{'state': 'will_hide'});
-      await tester.pump();
-      expect(controller.stateAsBool(), isTrue);
-      expect(controller.stateAsBool(true), isFalse);
       await controller.dispose();
     });
   });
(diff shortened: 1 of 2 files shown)


packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
```diff
@@ -22,25 +26,319 @@ void main() {
...
+        (WidgetTester tester) async {
+      final KeyboardDetectionController controller =
+          KeyboardDetectionController();
+      await emit(tester, <String, Object?>{'noState': true});
+      await tester.pump();
+      await emit(tester, <String, Object?>{'state': 123});
+      await tester.pump();
+      expect(controller.state, KeyboardState.unknown);
+      await controller.dispose();
+    });
```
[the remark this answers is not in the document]
seungsoo47: Done.

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
```diff
@@ -22,25 +26,319 @@ void main() {
...
+      controller.registerCallback((KeyboardState state) {
+        seen.add(state);
+        return true;
+      });
+      await emit(tester, <String, Object?>{'state': 'show'});
+      await tester.pump();
+      await tester.pump();
+      await emit(tester, <String, Object?>{'state': 'hide'});
+      await tester.pump();
+      await tester.pump();
```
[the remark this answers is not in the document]
seungsoo47: Done.

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
```diff
@@ -22,25 +26,317 @@ void main() {
...
+    testWidgets('maps the transitional visibling state',
+        (WidgetTester tester) async {
+      final KeyboardDetectionController controller =
+          KeyboardDetectionController();
+      await emit(tester, <String, Object?>{'state': 'will_show'});
+      await tester.pump();
+      expect(controller.stateAsBool(), isFalse);
+      expect(controller.stateAsBool(true), isTrue);
+      await controller.dispose();
+    });
```
JSUYA: This is similar to the "reports visibling on will_show event" test.
Since it verifies the values ​​of each data point for the "will_show" state, I believe it can be integrated into a one test.
Please combine similar tests into one. They are essentially duplicate tests.

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
```diff
@@ -13,7 +15,9 @@ void main() {
   const String channelName = 'tizen/internal/inputpanel';
   const StandardMethodCodec codec = StandardMethodCodec();
 
-  Future<void> emit(WidgetTester tester, Map<String, Object?> payload) async {
+  // Injects an event into the input-panel [EventChannel] the controller
+  // listens to, simulating a message coming from the flutter-tizen embedder.
```
JSUYA: That is an unnecessary comment.

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
```diff
@@ -13,7 +15,9 @@ void main() {
...
+  Future<void> emit(WidgetTester tester, Object? payload) async {
```
JSUYA: `payload`'s type has changed, but we are using `Map<String, Object?>` in all cases. Is there a reason you changed it to `Object?`

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
seungsoo47: Done.

packages/keyboard_detection/example/integration_test/keyboard_detection_tizen_test.dart
seungsoo47: Done.

JSUYA (approved):

PR #1049, opened 2026-06-30
[tizen_app_manager] Add integration tests

seungsoo47:
Fix a memory leak in AppRunningContext.packageId where the native string returned by app_context_get_package_id was not freed, and add 9 integration test cases covering AppManager and AppRunningContext APIs.

### DIFF
--- a/packages/tizen_app_manager/CHANGELOG.md
+++ b/packages/tizen_app_manager/CHANGELOG.md
@@ -1,4 +1,4 @@
-## NEXT
+## 0.2.5
 
 * Fix memory leak in `AppRunningContext.packageId` by freeing the native string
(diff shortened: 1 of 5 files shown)


packages/tizen_app_manager/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## NEXT
```
JSUYA: I recommend updating the version because the memory leak has been fixed.

JSUYA (approved):

packages/tizen_app_manager/CHANGELOG.md
seungsoo47: done

PR #1050, opened 2026-06-30
[tizen_audio_manager] Add regression integration tests

seungsoo47:
Add 2 integration test cases for previously untested public API:
- currentPlaybackType returns a valid AudioVolumeType: verifies the getter returns a recognized enum value.
- onChanged emits VolumeChangedEvent when volume is changed: subscribes to the stream, triggers a volume change via setLevel, and asserts the emitted event carries the correct type and level.

All 19 tests pass on the Samsung TV emulator (T-samsung-10.0-x86_64).

### DIFF
--- a/packages/tizen_audio_manager/example/integration_test/audio_manager_test.dart
+++ b/packages/tizen_audio_manager/example/integration_test/audio_manager_test.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:tizen_audio_manager/tizen_audio_manager.dart';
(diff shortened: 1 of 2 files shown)


packages/tizen_audio_manager/example/integration_test/audio_manager_test.dart
```diff
@@ -204,4 +206,44 @@ void main() {
     level = await AudioManager.volumeController.getLevel(AudioVolumeType.voip);
     expect(level, equals(0));
   });
+
+  testWidgets('currentPlaybackType returns a valid AudioVolumeType',
+      (WidgetTester tester) async {
+    final AudioVolumeType type =
+        await AudioManager.volumeController.currentPlaybackType;
+    expect(AudioVolumeType.values, contains(type));
+  });
```
JSUYA, line 215: packages/tizen_audio_manager/lib/tizen_audio_manager.dart
```
case 'none':
default:
  return AudioVolumeType.none;
```
This test looks to always be true because it always returns one of the Types.
Please check if this is a valid test.

seungsoo47: Yes. You're right that the test is always true because _stringToAudioVolumeType has a default case that maps any unknown string to AudioVolumeType.none, so the returned value will always be within AudioVolumeType.values.
However, the intent of this test is to verify that the native getCurrentPlaybackType channel call completes successfully on the device without throwing an exception. It's a smoke test for the native side. If you think this is too weak to justify keeping, I'll remove it.

JSUYA: If this is about maintaining native tests, let's leave it as is.

JSUYA (approved):

PR #1055, opened 2026-07-01
[tizen_package_manager] Add integration tests and fix icon path handling

seungsoo47:
Adds regression integration tests for the `tizen_package_manager` plugin and fixes package icon path handling.

  ### Plugin fix
  - Return a null icon path (instead of an empty string) when a package has no icon, and only use `package_info_get_icon`'s result when the call succeeds (`PACKAGE_MANAGER_ERROR_NONE`) and the path is non-empty; always free the returned buffer. The icon is optional, so a failure is not treated as fatal.
  - Bumps the version to `0.4.3`.

  ### Integration tests
  - Add tests for `getPackageInfo`, `getPackageSizeInfo`, `getPackagesInfo`, package/size-info field validation, error paths (empty / invalid IDs), and the install/uninstall/update event streams.

  ## Test

  All tests pass on:
  - tv-9.0 emulator (x86)
  - tv-10.0 emulator (x86_64)
  - Raspberry Pi 4 (armv7 / ARM)
 
### DIFF
--- a/packages/tizen_package_manager/example/integration_test/tizen_package_manager_test.dart
+++ b/packages/tizen_package_manager/example/integration_test/tizen_package_manager_test.dart
@@ -42,14 +42,6 @@ void main() {
   });
 
   group('PackageInfo fields', () {
-    testWidgets('installedStorageType is a valid StorageType value', (
-      WidgetTester tester,
-    ) async {
-      final PackageInfo info =
-          await PackageManager.getPackageInfo(currentPackageId);
-      expect(StorageType.values, contains(info.installedStorageType));
-    });
-
     testWidgets('iconPath is null or a non-empty string', (
       WidgetTester tester,
     ) async {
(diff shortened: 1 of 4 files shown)


packages/tizen_package_manager/example/integration_test/tizen_package_manager_test.dart
```diff
@@ -37,4 +40,139 @@ void main() {
...
     expect(infos.isNotEmpty, true);
   });
+
+  group('PackageInfo fields', () {
+    testWidgets('installedStorageType is a valid StorageType value', (
+      WidgetTester tester,
+    ) async {
+      final PackageInfo info =
+          await PackageManager.getPackageInfo(currentPackageId);
+      expect(StorageType.values, contains(info.installedStorageType));
```
JSUYA: I think this test is invalid because the type of info.installedStorageType is StorageType and it must return one of them.

packages/tizen_package_manager/example/integration_test/tizen_package_manager_test.dart
```diff
@@ -37,4 +40,139 @@ void main() {
...
+      await subscription.cancel();
+    });
+
+    testWidgets('onUpdateProgressChanged can be listened to without error', (
+      WidgetTester tester,
+    ) async {
+      final StreamSubscription<PackageEvent> subscription =
+          PackageManager.onUpdateProgressChanged.listen(null);
+      await subscription.cancel();
+    });
```
JSUYA, line 171: This test does not have an except process. Is it simply checking whether it works? If it is an await, you must check the result through a timeout or separate processing and verify the successful call status.

JSUYA (approved):

PR #1056, opened 2026-07-07
[webview_flutter_lwe] Add integration test based on upstream v4.13.1

seungsoo47:
Port the 'can receive url changes' NavigationDelegate integration test from upstream webview_flutter v4.13.1, and fix three bugs found while making it pass on Tizen:

  * Implement `onUrlChange` for the navigation delegate.
  * Fix navigation delegate callbacks being dropped after the delegate is replaced.
  * Fix a raster thread crash on WebView disposal (TBM surface freed while still being composited).
  * Fix a use-after-free when async callbacks run after the WebView is disposed.
  * Update lightweight web engine (1.3.40)

Bumped version 0.4.1 → 0.4.2.

All tests pass on TV/RPi4 emulator or device

> **Note**: This PR depends on the corresponding embedder PR(https://github.com/flutter-tizen/embedder/pull/182) and LWE PR and must be merged only after those are applied.


packages/webview_flutter_lwe/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: webview_flutter_lwe
 description: Tizen implementation of the webview_flutter plugin backed by Lightweight Web Engine.
 homepage: https://github.com/flutter-tizen/plugins
 repository: https://github.com/flutter-tizen/plugins/tree/master/packages/webview_flutter_lwe
-version: 0.4.1
+version: 0.4.2
```
JSUYA: I recommend updating to 0.5.0 because the implementation of the `onUrlChange` feature has been added.

packages/webview_flutter_lwe/tizen/src/webview.cc
```diff
@@ -257,12 +285,51 @@ std::string WebView::GetNavigationDelegateChannelName() {
...
+      GetTextureId(), [pool, dispatcher = dispatcher_]() {
+        // Invoked on the render thread once the texture is fully unregistered.
+        // Hand the pool to a main-thread task for destruction; by this point
+        // the raster thread has released all GPU resources and the LWE engine
+        // has unmapped all TBM surfaces, so no in-flight frames remain.
+        dispatcher->dispatchTaskOnMainThread([pool]() {
+          // Pool destructor calls tbm_surface_destroy on all buffers.
+        });
+      });
 }
```
JSUYA: Not all code needs to be commented on. Be careful with automatically generated comments.
If this code change is unusual, please write it briefly.
And if a detailed explanation is needed, writing it in the PR body allows us to track the history.

packages/webview_flutter_lwe/lib/src/lwe_webview_controller.dart
```diff
@@ -383,14 +394,23 @@ class LweNavigationDelegate extends PlatformNavigationDelegate {
   LweNavigationDelegate(super.params) : super.implementation();
 
   late final MethodChannel _navigationDelegateChannel;
+  bool _isChannelCreated = false;
```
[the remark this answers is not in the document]
JSUYA: @seungsoo47 Could you please review this suggestion once again? It looks that there is no need to add _isChannelCreated.

packages/webview_flutter_lwe/tizen/src/webview.cc
```diff
@@ -153,28 +164,39 @@ WebView::WebView(flutter::PluginRegistrar* registrar, int view_id,
...
+        dispatcher_->dispatchTaskOnMainThread([this, args,
+                                               alive = is_alive_]() {
+          if (!*alive) return;
           navigation_delegate_channel_->InvokeMethod(
               "onPageStarted", std::make_unique<flutter::EncodableValue>(args));
+          // The lightweight web engine has no dedicated URL-change
+          // callback, so report a URL change whenever a navigation
+          // starts.
+          navigation_delegate_channel_->InvokeMethod(
+              "onUrlChange", std::make_unique<flutter::EncodableValue>(args));
```
JSUYA, line 173: Why should we call urlChange on PageStarted instead of PageLoaded(or onPageFinished)? Is there a reason?

packages/webview_flutter_lwe/lib/src/lwe_webview_controller.dart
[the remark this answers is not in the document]
seungsoo47: I have applied it.

packages/webview_flutter_lwe/tizen/src/webview.cc
seungsoo47: On Android, 'onUrlChange' originates from 'WebViewClient.doUpdateVisitedHistory'. It triggers almost simultaneously with 'onPageStarted' rather than at the 'onPageFinished' stage. On iOS (WKWebView), 'onUrlChange' is triggered by monitoring the URL property; this value updates immediately when navigation starts, well before the page finishes loading. Therefore, on both platforms, 'onUrlChange' signifies that the browser has begun navigating to a new URL, not that the page has fully loaded.
Since LWE does not have a dedicated 'URL change' callback, I think the event that best corresponds to that moment is 'PageStarted'.

JSUYA (approved):

PR #1057, opened 2026-07-08
[tizen_window_manager] Add tizen-core-wl window backend

JSUYA:
Introduce a WindowProxyInterface abstraction with two backends selected at runtime by WindowProxy:
  - TizenCoreWlWindowProxy (tizen-core-wl) on Tizen API version >= 11.0 when the library is available
  - EcoreWl2WindowProxy (ecore_wl2) otherwise, including as a fallback when tizen-core-wl cannot be loaded on 11.0+

Both backends implement WindowProxyInterface (GetGeometry/Activate/Lower) and resolve their symbols at runtime via dlopen/dlsym. TizenWindowManager constructs WindowProxy, which picks the backend by querying the platform version at startup.

### DIFF
--- a/packages/tizen_window_manager/tizen/src/tizen_core_wl_window_proxy.cc
+++ b/packages/tizen_window_manager/tizen/src/tizen_core_wl_window_proxy.cc
@@ -1,0 +1,62 @@
+// Copyright 2025 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "tizen_core_wl_window_proxy.h"
+
+#include <dlfcn.h>
+
+#include "log.h"
+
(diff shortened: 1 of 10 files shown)


packages/tizen_window_manager/tizen/src/tizen_core_wl_window_proxy.cc
```diff
@@ -0,0 +1,62 @@
...
+  }
+  get_geometry_(window, x, y, width, height);
+}
+
+void TizenCoreWlWindowProxy::Activate(void *window) {
+  if (!handle_) {
+    LOG_ERROR("tizen-core-wl handle not valid");
+    return;
+  }
+  activate_(window);
```
xiaowei-guan: get_geometry
activate
lower
the return values from these methods never checked. Suggest to add a check when return an error code.

JSUYA:
https://github.com/flutter-tizen/plugins/pull/1136

PR #1059, opened 2026-07-09
[permission_handler] Add regression integration tests based on upstream v12.0.1

seungsoo47:
Upstream permission_handler v12.0.1 has no integration tests. Added Tizen regression tests for the public permission API covering all privileges declared in the example app's tizen-manifest.xml:

- Permission.status: camera, microphone, location, mediaLibrary, storage, contacts
- Permission.serviceStatus: location service status is valid
- Permission.request: requesting camera permission returns granted
- Permission.request: requesting multiple permissions returns granted
- openAppSettings (skipped, not supported on emulators)

Note: only testable on Galaxy Watch (wearable profile) where runtime permission dialogs are applicable. On TV devices, all declared permissions are granted by default.

### DIFF
--- a/packages/permission_handler/CHANGELOG.md
+++ b/packages/permission_handler/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Add 8 integration test cases.
+
 ## 1.4.4
 
 * Remove Ecore API.
--- a/packages/permission_handler/example/integration_test/permission_handler_test.dart
+++ b/packages/permission_handler/example/integration_test/permission_handler_test.dart
@@ -11,27 +11,27 @@ void main() {
 
   group('Permission.status', () {
     testWidgets('camera permission is granted', (tester) async {
-      expect(await Permission.camera.status.isGranted, true);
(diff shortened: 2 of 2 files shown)


JSUYA:
> Note: only testable on Galaxy Watch (wearable profile) where runtime permission dialogs are applicable. On TV devices, all declared permissions are granted by default.

I am not sure if adding this test is a meaningful task. It seems like a package that should effectively be discontinued. What do you think?
I believe the test is meaningless because all values ​​return true.

seungsoo47:
> > Note: only testable on Galaxy Watch (wearable profile) where runtime permission dialogs are applicable. On TV devices, all declared permissions are granted by default.
> 
> I am not sure if adding this test is a meaningful task. It seems like a package that should effectively be discontinued. What do you think? I believe the test is meaningless because all values ​​return true.

As you know, this plugin does not run on the TV profile. However, as we discussed offline, since we can still check Tizen's Permission APIs on a Tizen platform (like a RPi), I will keep the plugin and the tests.

JSUYA (approved):

PR #1069, opened 2026-07-15
[webview_flutter] Implement clearLocalStorage/onHttpError and add integration tests based on upstream v4.13.1

seungsoo47:
- Implement native Tizen support for `clearLocalStorage` and `onHttpError`, and bump `webview_flutter_tizen` to 0.10.1.
- Port the remaining runnable upstream v4.13.1 integration tests (onHttpError x2, clearLocalStorage) and align async style (await instead of unawaited()) with upstream.
- Stabilize the scroll position test by polling for the settled value instead of reading it once.
- Fix a WebView disposal race where engine-owned TBM surfaces could be freed while a raster-thread frame still read them, and work around a chromium-efl crash on the TV 10.0 emulator during teardown.
- `flutter-tizen test` (or example integration tests) on TV 10.0 emulator — 19/19 passing, including previously-crashing disposal cases

### DIFF
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Implement `clearLocalStorage`.
 * Implement `onHttpError` for the navigation delegate.
+* Fix races and use-after-frees on WebView disposal, and avoid a web engine
+  teardown crash on the Tizen 10.0 TV emulator.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.10.1
+
+* Implement `clearLocalStorage`.
+* Implement `onHttpError` for the navigation delegate.
+
 ## 0.10.0
 
 * Update minimum supported SDK version to Flutter 3.32/Dart 3.8.
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 * Implement `clearLocalStorage`.
 * Implement `onHttpError` for the navigation delegate.
-* Fix races and use-after-frees on WebView disposal, and avoid a web engine
-  teardown crash on the Tizen 10.0 TV emulator.
+* Fix races and use-after-frees on WebView disposal, including a buffer-pool
+  use-after-free on the raster thread.
+* Replace the Ecore main loop API with GLib.
+* Narrow the TV emulator teardown workaround to a runtime profile check.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -5,9 +5,10 @@
 * Fix races and use-after-frees on WebView disposal, including a buffer-pool
   use-after-free on the raster thread.
 * Replace the Ecore main loop API with GLib.
-* Remove the TV emulator's evas_object_del() workaround; drain all pending
-  WebView teardowns before ewk_shutdown() instead, fixing a native crash on
-  app exit and enabling simultaneous use of multiple WebViews.
+* Remove the TV emulator's evas_object_hide() workaround (which skipped
+  evas_object_del() to avoid a crash); drain all pending WebView teardowns
+  before ewk_shutdown() instead, fixing a native crash on app exit and
+  enabling simultaneous use of multiple WebViews.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Implement `onHttpError` for the navigation delegate.
 * Fix races and use-after-frees on WebView disposal, including a buffer-pool
   use-after-free on the raster thread.
-* Replace the Ecore main loop API with GLib.
+* Use GLib to schedule the deferred WebView teardown.
 * Drain all pending WebView teardowns before ewk_shutdown(), fixing a native
   crash on app exit and enabling simultaneous use of multiple WebViews.
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -6,7 +6,7 @@
   use-after-free on the raster thread.
 * Use GLib to schedule the deferred WebView teardown.
 * Drain all pending WebView teardowns before ewk_shutdown(), fixing a native
-  crash on app exit and enabling simultaneous use of multiple WebViews.
+  crash on app exit.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -5,10 +5,8 @@
 * Fix races and use-after-frees on WebView disposal, including a buffer-pool
   use-after-free on the raster thread.
 * Replace the Ecore main loop API with GLib.
-* Remove the TV emulator's evas_object_hide() workaround (which skipped
-  evas_object_del() to avoid a crash); drain all pending WebView teardowns
-  before ewk_shutdown() instead, fixing a native crash on app exit and
-  enabling simultaneous use of multiple WebViews.
+* Drain all pending WebView teardowns before ewk_shutdown(), fixing a native
+  crash on app exit and enabling simultaneous use of multiple WebViews.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -2,11 +2,9 @@
 
 * Implement `clearLocalStorage`.
 * Implement `onHttpError` for the navigation delegate.
-* Fix races and use-after-frees on WebView disposal, including a buffer-pool
-  use-after-free on the raster thread.
+* Fix races and use-after-frees on WebView disposal, including a buffer-pool use-after-free on the raster thread.
 * Use GLib to schedule the deferred WebView teardown.
-* Drain all pending WebView teardowns before ewk_shutdown(), fixing a native
-  crash on app exit.
+* Drain all pending WebView teardowns before ewk_shutdown(), fixing a native crash on app exit.
 
 ## 0.10.0
 
--- a/packages/webview_flutter/CHANGELOG.md
+++ b/packages/webview_flutter/CHANGELOG.md
@@ -5,7 +5,9 @@
 * Fix races and use-after-frees on WebView disposal, including a buffer-pool
   use-after-free on the raster thread.
 * Replace the Ecore main loop API with GLib.
-* Narrow the TV emulator teardown workaround to a runtime profile check.
+* Remove the TV emulator's evas_object_del() workaround; drain all pending
+  WebView teardowns before ewk_shutdown() instead, fixing a native crash on
+  app exit and enabling simultaneous use of multiple WebViews.
 
 ## 0.10.0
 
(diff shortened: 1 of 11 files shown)


packages/webview_flutter/tizen/src/webview.h
```diff
@@ -117,6 +118,14 @@ class WebView : public PlatformView {
...
   std::unique_ptr<BufferPool> tbm_pool_;
   bool disposed_ = false;
+  // Set under mutex_ at the start of Dispose(). The raster thread checks it
+  // in ObtainGpuSurface() and stops being handed engine-owned TBM surfaces
+  // that are about to be freed by the deferred evas_object_del().
+  bool is_disposing_ = false;
+  // Set to false at the start of Dispose(). A pending "navigationRequest"
+  // reply from Dart (resolved asynchronously) captures a copy and checks it
+  // before dereferencing this WebView, avoiding a use-after-free.
+  std::shared_ptr<bool> is_alive_ = std::make_shared<bool>(true);
```
JSUYA: Please delete all unnecessary comments.
If you wish to explain how disposing and is_alive work, please leave an explanation in the PR.
Please only leave comments for parts that must be referenced for functionality (TODO, NOTE) or descriptions that need to be published.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -173,33 +191,111 @@ void WebView::Dispose() {
...
+  // pool) is deferred until then. The callback fires on the render thread;
+  // evas_object_del() must run on the main thread, so hop back via
+  // ecore_main_loop_thread_safe_call_async().
+  struct TeardownContext {
+    Evas_Object* instance;
+    std::unique_ptr<BufferPool> pool;
+  };
+  auto* context = new TeardownContext{instance, std::move(pool)};
+  texture_registrar_->UnregisterTexture(GetTextureId(), [context]() {
+    ecore_main_loop_thread_safe_call_async(
```
JSUYA: The ecore API is no longer in use.

JSUYA (commented):
Do the parts related to dispose and evas_object_del ensure that it works without issues even when displaying multiple views?

JSUYA:
@codex review

seungsoo47:
> Do the parts related to dispose and evas_object_del ensure that it works without issues even when displaying multiple views?

While the WebView can functionally support multiple views, this isn't currently supported in practice due to memory-leak and crash issues. I will create issues regarding these problems.
https://github.com/flutter-tizen/plugins/issues/1076
https://github.com/flutter-tizen/plugins/issues/1077

seungsoo47:
@codex review

seungsoo47:
@codex review

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -173,33 +232,128 @@ void WebView::Dispose() {
...
+  if (!g_shared_canvas) {
+    // "wayland_shm", not "wayland_egl": this canvas only hosts the ewk_view
+    // smart object (content arrives via tbm_surface), and wayland_egl raced
+    // libtpl-egl's wl_egl_thread teardown on disposal.
+    g_shared_canvas = ecore_evas_new("wayland_shm", 0, 0, 1, 1, 0);
+  }
+  return g_shared_canvas;
+}
+
+// static
```
JSUYA: Please delete all unnecessary comments.

packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.10.1
+
+* Implement `clearLocalStorage`.
+* Implement `onHttpError` for the navigation delegate.
+* Fix races and use-after-frees on WebView disposal, including a buffer-pool
+  use-after-free on the raster thread.
+* Replace the Ecore main loop API with GLib.
```
JSUYA: Is "Replace" correct? There is no that previously used the ecore main loop.

packages/webview_flutter/tizen/src/webview.h
```diff
@@ -58,6 +59,16 @@ class WebView : public PlatformView {
...
                                                        size_t height);
 
+  // Blocks until every WebView's deferred delete (queued in Dispose()) has
+  // run, force-deleting stragglers after a timeout. Must be called before
+  // ewk_shutdown(), which fatally CHECKs on any live Ewk_View.
+  static void FlushPendingTeardowns();
+
+  static Ecore_Evas* GetSharedCanvas();
+
+  // Must be called after FlushPendingTeardowns() and before ewk_shutdown().
```
JSUYA: Please delete all unnecessary comments.

packages/webview_flutter/tizen/src/webview_flutter_tizen_plugin.cc
```diff
@@ -4,27 +4,38 @@
...
+// run once.
 class WebviewFlutterTizenPlugin : public flutter::Plugin {
  public:
   static void RegisterWithRegistrar(flutter::PluginRegistrar* registrar) {
     auto plugin = std::make_unique<WebviewFlutterTizenPlugin>();
     registrar->AddPlugin(std::move(plugin));
   }
 
-  WebviewFlutterTizenPlugin() {}
+  WebviewFlutterTizenPlugin() { ewk_init(); }
```
JSUYA: It is not appropriate to call the ewk_ API directly from webview_flutter_tizen_plugin.cc.

packages/webview_flutter/tizen/src/webview_flutter_tizen_plugin.cc
```diff
@@ -4,27 +4,38 @@
...
 
-  virtual ~WebviewFlutterTizenPlugin() {}
+  virtual ~WebviewFlutterTizenPlugin() {
+    // See WebView::FlushPendingTeardowns() for why this must run first.
+    WebView::FlushPendingTeardowns();
+    WebView::FreeSharedCanvas();
+    ewk_shutdown();
```
JSUYA: It is not appropriate to call the ewk_ API directly from webview_flutter_tizen_plugin.cc.

packages/webview_flutter/tizen/src/webview_flutter_tizen_plugin.cc
```diff
@@ -4,27 +4,38 @@
...
     auto plugin = std::make_unique<WebviewFlutterTizenPlugin>();
     registrar->AddPlugin(std::move(plugin));
   }
 
-  WebviewFlutterTizenPlugin() {}
+  WebviewFlutterTizenPlugin() { ewk_init(); }
 
-  virtual ~WebviewFlutterTizenPlugin() {}
+  virtual ~WebviewFlutterTizenPlugin() {
+    // See WebView::FlushPendingTeardowns() for why this must run first.
```
JSUYA: Please delete unnecessary comments.

packages/webview_flutter/tizen/src/webview_flutter_tizen_plugin.cc
JSUYA: I think it can be handled in a single function within the WebView class.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -87,6 +107,45 @@ bool GetValueFromEncodableMap(const flutter::EncodableValue* arguments,
...
+  Evas_Object* instance = nullptr;
+  // Guards against the queued callback and a force-delete both deleting
+  // the same instance.
+  std::atomic<bool> completed{false};
+};
+
+std::mutex g_pending_teardown_mutex;
+std::vector<std::shared_ptr<PendingTeardown>> g_pending_teardowns;
+
+Ecore_Evas* g_shared_canvas = nullptr;
```
JSUYA: Why is it "shared" canvas? and it's not even canvas.

packages/webview_flutter/tizen/src/webview.h
```diff
@@ -115,8 +127,14 @@ class WebView : public PlatformView {
   std::unique_ptr<FlMethodChannel> navigation_delegate_channel_;
   std::unique_ptr<flutter::TextureVariant> texture_variant_;
   std::mutex mutex_;
-  std::unique_ptr<BufferPool> tbm_pool_;
+  std::shared_ptr<BufferPool> tbm_pool_;
   bool disposed_ = false;
+  // Guarded by mutex_. Keeps the raster thread from being handed TBM surfaces
+  // during the deferred teardown.
+  bool is_disposing_ = false;
```
JSUYA: This PR tracks the "disposing state" in four separate places: disposed_, is_disposing_, is_alive_(shared_ptr), PendingTeardown+TeardownContext, and even the buffer_pool's active_buffers set.

Please minimize unnecessary implementations as much as possible.

JSUYA (commented):
There is too much excessive implementation. Please reduce duplicate implementations and delete all unnecessary comments.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -170,36 +229,134 @@ std::string WebView::GetNavigationDelegateChannelName() {
...
+      CompletePendingTeardown(pending);
+      continue;
+    }
+    // Non-blocking: pump the same context the queued g_timeout_add_full hop
+    // runs on, without stalling past the deadline check above.
+    if (!g_main_context_iteration(g_main_context_default(), FALSE)) {
+      g_usleep(1000);
+    }
+  }
 }
```
JSUYA: I understand calculating 2 seconds using g_get_monotonic_time(), but I do not quite understand why g_usleep 1 second needs to be used with g_main_context_iteration(). Also, when there are two or more conditions in the pending list, it seems that only the front of the deadline is checked, and subsequent ones are not. It appears structurally unstable.

I understand the need to manage the creation/destruction of webviews due to multiple creation requests, but by default, the chromieum efl library currently does not support multiple creations. I am not sure if this code change is valid.

packages/webview_flutter/tizen/src/webview.cc
JSUYA: remove all `// static` and more all unnecessary comments.

packages/webview_flutter/tizen/src/webview.cc
seungsoo47: Thank you for the review.

First question: The sleep is 1 millisecond, not 1 second. This sleep only happens when g_main_context_iteration() had nothing to do. We sleep a little bit so the loop does not spin at full speed while it waits for the g_timeout_add_full callback from Dispose() to run.

Second question: About the pending list, this is now fixed. Before, only the oldest item was checked. Now, once the 2-second deadline passes, we take a snapshot of the whole list and force-complete every item in it, not just the front one. So the structure issue you saw is not there anymore.

Multiple Webviews: I also tested this on three devices: rpi4, a TV emulator, and a real TV. I ran the test 5 times on each device. Every time, I could create and close multiple WebViews with no crash and no other problem.
However, there is a separate issue with focus when I show more than one WebView. I agree this is a real problem. I think it should not block this PR. I will track it in a new issue, #1092.
You said chromium-efl does not support creating more than one WebView by default. Can you share where you found this? I want to check it too. I have not found any document that says this yet.

packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.10.1
+
+* Implement `clearLocalStorage`.
+* Implement `onHttpError` for the navigation delegate.
+* Fix races and use-after-frees on WebView disposal, including a buffer-pool
+  use-after-free on the raster thread.
```
JSUYA: please no add new line

packages/webview_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+* Use GLib to schedule the deferred WebView teardown.
+* Drain all pending WebView teardowns before ewk_shutdown(), fixing a native
+  crash on app exit.
```
JSUYA: please no add new line

JSUYA (approved):

packages/webview_flutter/tizen/src/webview.cc
JSUYA: I remember we discussed in offline that creating more than one webview is not work(?). If there are no issues, I think it can be supported.

PR #1070, opened 2026-07-16
[tizen_rpc_port] Update RpcPort to support TIDL protocol version 2

pjh9216:
Add low level Parcel primitives to support the TIDL protocol version 2 Dart generator under packages/tizen_rpc_port.

Implement the dataSize, reserve, and reader getter/setter properties in the Parcel class using Dart FFI to map them to the native C APIs:
- rpc_port_parcel_get_reader
- rpc_port_parcel_set_reader
- rpc_port_parcel_get_data_size
- rpc_port_parcel_set_data_size
- rpc_port_parcel_reserve

### DIFF
--- a/packages/tizen_rpc_port/lib/src/parcel.dart
+++ b/packages/tizen_rpc_port/lib/src/parcel.dart
@@ -157,8 +157,8 @@ class Parcel {
 
   static final Finalizer<rpc_port_parcel_h> _finalizer =
       Finalizer<rpc_port_parcel_h>((rpc_port_parcel_h handle) {
-        tizen.rpc_port_parcel_destroy(handle);
-      });
+    tizen.rpc_port_parcel_destroy(handle);
+  });
 
   /// Gets a byte array backed by the raw data of this parcel.
   Uint8List asRaw() {
@@ -554,12 +554,13 @@ class Parcel {
 
 final DynamicLibrary _libRpcPort = DynamicLibrary.open('librpc-port.so.1');
 
-final int Function(rpc_port_parcel_h, Pointer<Uint32>)
-_rpcPortParcelGetReader = _libRpcPort
-    .lookup<NativeFunction<Int32 Function(rpc_port_parcel_h, Pointer<Uint32>)>>(
-      'rpc_port_parcel_get_reader',
-    )
-    .asFunction();
+final int Function(rpc_port_parcel_h, Pointer<Uint32>) _rpcPortParcelGetReader =
+    _libRpcPort
+        .lookup<
+            NativeFunction<Int32 Function(rpc_port_parcel_h, Pointer<Uint32>)>>(
+          'rpc_port_parcel_get_reader',
+        )
+        .asFunction();
 
 final int Function(rpc_port_parcel_h, int) _rpcPortParcelSetReader = _libRpcPort
     .lookup<NativeFunction<Int32 Function(rpc_port_parcel_h, Uint32)>>(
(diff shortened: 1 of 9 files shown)


packages/tizen_rpc_port/lib/src/parcel.dart
```diff
@@ -484,4 +484,94 @@ class Parcel {
...
+    if (ret != 0) {
+      throw PlatformException(
+        code: ret.toString(),
+        message: tizen.get_error_message(ret).toDartString(),
+      );
+    }
+  }
 }
+
+final DynamicLibrary _libRpcPort = DynamicLibrary.open('librpc-port.so.1');
```
JSUYA: Do these APIs have a dependency on the Tizen version? Are they only used in Tizen 11?

JSUYA (commented):
Please check gemini's review comments. If you think they are invalid, please resolve them.

And

Please modify the format using Dart Formatter.

https://github.com/flutter-tizen/plugins/actions/runs/29460673863/job/87506437916?pr=1070
```
2026-07-16T00:35:55.7848891Z Running command: "dart format example/client/integration_test/tizen_rpc_port_test.dart example/client/lib/main.dart example/client/lib/message_client.dart example/client/test_driver/integration_test.dart example/server/integration_test/tizen_rpc_port_test.dart example/server/lib/main.dart example/server/lib/message_server.dart example/server/test_driver/integration_test.dart lib/src/parcel.dart lib/src/port.dart lib/src/proxy_base.dart lib/src/stub_base.dart lib/tizen_rpc_port.dart" in /home/runner/work/plugins/plugins/packages/tizen_rpc_port
```

packages/tizen_rpc_port/lib/src/parcel.dart
pjh9216: Yes, It is dependent on tizen version. Since Tizen 10

JSUYA: Are these APIs public open ACR APIs?
If this API is changed in librpc-port.so in Tizen 11.0 or 10.1, platform compatibility cannot be guaranteed to users.
At the very least, there needs to be a way to verify that these APIs are managed.

pjh9216: It is internal API but It will not be changed to support backward compatibility. 
The native APIs is already  working in many apps using TIDL.

pjh9216: I made sample app to test action. Working well. (https://github.sec.samsung.net/jh9216-park/ActionSampleAppDart)
Please review and merge it ASAP

JSUYA: This API is only compatible with version 10.0, and there are absolutely no guards prepared for cases where it is used with other APIs.

Furthermore, there are no related integration_tests, and there are no modifications whatsoever to the README, CHANGELOG, or pubspec.yaml related to the tizen_rpc_port deployment.

flutter-tizen includes plugin code in the app packaging. It may also be deployed with the minimum supported version set (general guideline).

Therefore, version compatibility should be maintained whenever possible, and measures must be taken to prevent crashes when a non-functional API is called.

Please add relevant guards and ensure that appropriate exception messages are handled in the event of an unsupported API.

pjh9216: please see [tizen_rpc_port] Guard Parcel APIs for compatibility commit

JSUYA: Please check analyze and integration_test CI fails.

pjh9216: Fixed check analyze issue. Please run CI test. I don't have permission

JSUYA (approved):

PR #1078, opened 2026-08-05
[tizen_app_control] Fix potential race issue

seungsoo47:
Subscribe to `onAppControl` before calling `sendLaunchRequest()` to avoid a potential race where broadcast-stream events could be missed.

### DIFF
--- a/packages/tizen_app_control/example/integration_test/tizen_app_control_test.dart
+++ b/packages/tizen_app_control/example/integration_test/tizen_app_control_test.dart
@@ -39,9 +39,11 @@ void main() {
       appId: kAppId,
       operation: 'operation_1',
     );
+    final Future<ReceivedAppControl> receivedFuture =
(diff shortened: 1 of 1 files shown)


JSUYA:
@codex review

JSUYA (commented):
Could you clarify the specific scenario in which this issue occurs? The existing logic does not appear to have this problem. If the AppControl event arrives before Dart subscribes, the native embedder stores it in queue_ and delivers it when AppControl.onAppControl.first later subscribes.

seungsoo47:
> Could you clarify the specific scenario in which this issue occurs? The existing logic does not appear to have this problem. If the AppControl event arrives before Dart subscribes, the native embedder stores it in queue_ and delivers it when AppControl.onAppControl.first later subscribes.

Thanks. I will close this pr.

PR #1082, opened 2026-08-11
Fix "FocusManager used after being disposed" in integration tests

seungsoo47:
- Convert testWidgets to test for tests that don't use WidgetTester, resolving FocusManager lifecycle conflicts in integration tests
- Applied across 30+ packages (video_player, wakelock_plus, camera, geolocator, etc.)


JSUYA (commented):
The discontinued plugin does not need to be modified.
Please write CHANGELOG with *NEXT to indicate that test has been updated.

JSUYA (approved):

PR #1083, opened 2026-08-12
[flutter_inappwebview] Fix navigation/getUrl/scroll bugs and add integration tests

seungsoo47:
- Fix onTitleChanged to fire on later title changes (e.g. via JS document.title), not just once after load.
- Fix a getUrl() race with cancelled navigations, and skip the shouldOverrideUrlLoading round-trip for app-initiated navigations (loadUrl, goBack, reload, etc.).
- Fix scrollBy/getScrollX/getScrollY occasionally returning a stale position right after scrollTo/scrollBy (EWK applies scroll asynchronously).
- Add Tizen integration tests ported from upstream flutter_inappwebview v6.1.5.
- Bump flutter_inappwebview_tizen to 0.2.0.

### DIFF
--- a/packages/flutter_inappwebview/CHANGELOG.md
+++ b/packages/flutter_inappwebview/CHANGELOG.md
@@ -1,3 +1,15 @@
+## 0.2.0
+
+- Fix `onTitleChanged` to also fire when the page's title changes after the
+  initial load (e.g. when JavaScript updates `document.title`), instead of
+  only once when loading finishes.
+- Fix a race where `getUrl()` could return the URL of a navigation that was
+  cancelled via `shouldOverrideUrlLoading`, and skip the
+  `shouldOverrideUrlLoading` round-trip for app-initiated navigations
+  (`loadUrl`, `goBack`, `reload`, etc.).
+- Fix `scrollBy`/`getScrollX`/`getScrollY` occasionally returning a stale
+  scroll position right after `scrollTo`/`scrollBy`.
+
 ## 0.1.1
 
 - Fix a crash when a webview is disposed.
--- a/packages/flutter_inappwebview/tizen/src/webview.cc
+++ b/packages/flutter_inappwebview/tizen/src/webview.cc
@@ -354,6 +354,8 @@ void WebView::Dispose() {
                                    &WebView::OnNavigationPolicy);
     evas_object_smart_callback_del(webview_instance_, "url,changed",
                                    &WebView::OnUrlChange);
+    evas_object_smart_callback_del(webview_instance_, "title,changed",
+                                   &WebView::OnTitleChange);
     auto& ewk_view = EwkInternalApiBinding::GetInstance().view;
     if (ewk_view.OnJavaScriptAlert) {
       ewk_view.OnJavaScriptAlert(webview_instance_, nullptr, nullptr);
@@ -620,6 +622,8 @@ bool WebView::InitWebView() {
                                  &WebView::OnNavigationPolicy, this);
   evas_object_smart_callback_add(webview_instance_, "url,changed",
                                  &WebView::OnUrlChange, this);
+  evas_object_smart_callback_add(webview_instance_, "title,changed",
+                                 &WebView::OnTitleChange, this);
 
   Resize(width_, height_);
   evas_object_show(webview_instance_);
@@ -1110,6 +1114,18 @@ void WebView::OnUrlChange(void* data, Evas_Object* obj, void* event_info) {
   webview->webview_channel_->InvokeMethod(
       "onUpdateVisitedHistory",
       std::make_unique<flutter::EncodableValue>(args));
+}
+
+void WebView::OnTitleChange(void* data, Evas_Object* obj, void* event_info) {
+  WebView* webview = static_cast<WebView*>(data);
+  const char* title = static_cast<const char*>(event_info);
+  if (!title) {
+    return;
+  }
+  flutter::EncodableMap args = {
+      {flutter::EncodableValue("title"), flutter::EncodableValue(title)}};
+  webview->webview_channel_->InvokeMethod(
+      "onTitleChanged", std::make_unique<flutter::EncodableValue>(args));
 }
 
 void WebView::OnEvaluateJavaScript(Evas_Object* obj, const char* result_value,
--- a/packages/flutter_inappwebview/tizen/src/webview.cc
+++ b/packages/flutter_inappwebview/tizen/src/webview.cc
@@ -880,15 +880,12 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
     result->Success(flutter::EncodableValue(
         static_cast<bool>(ewk_view_forward_possible(webview_instance_))));
   } else if (method_name == "goBack") {
-    NavigateProgrammatically([this] {
-      ewk_view_back(webview_instance_);
-      return true;
-    });
+    NavigateProgrammatically(
+        [this] { return static_cast<bool>(ewk_view_back(webview_instance_)); });
     result->Success();
   } else if (method_name == "goForward") {
     NavigateProgrammatically([this] {
-      ewk_view_forward(webview_instance_);
-      return true;
+      return static_cast<bool>(ewk_view_forward(webview_instance_));
     });
     result->Success();
   } else if (method_name == "reload") {
@@ -968,18 +965,12 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
     int32_t x = 0, y = 0;
     ewk_view_scroll_pos_get(webview_instance_, &x, &y);
     if (target_scroll_x_ >= 0) {
-      if (x == target_scroll_x_) {
-        target_scroll_x_ = -1;
-      } else {
-        x = target_scroll_x_;
-      }
+      x = target_scroll_x_;
+      target_scroll_x_ = -1;
     }
     if (target_scroll_y_ >= 0) {
-      if (y == target_scroll_y_) {
-        target_scroll_y_ = -1;
-      } else {
-        y = target_scroll_y_;
-      }
+      y = target_scroll_y_;
+      target_scroll_y_ = -1;
     }
     result->Success(
         flutter::EncodableValue(method_name == "getScrollX" ? x : y));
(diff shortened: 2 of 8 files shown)


seungsoo47:
@codex review

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -901,6 +968,14 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
...
     int32_t x = 0, y = 0;
     ewk_view_scroll_pos_get(webview_instance_, &x, &y);
+    if (target_scroll_x_ >= 0) {
+      x = target_scroll_x_;
+      target_scroll_x_ = -1;
+    }
+    if (target_scroll_y_ >= 0) {
+      y = target_scroll_y_;
+      target_scroll_y_ = -1;
+    }
```
JSUYA, line 978: The getScrollX/getScrollY handler clears both axis targets on a single call. After scrollTo(30, 40), calling getScrollX() also resets target_scroll_y_, so an immediately following getScrollY() reads EWK's not-yet-applied stale position — reintroducing the exact symptom
  this PR fixes.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -1103,15 +1207,34 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
...
+  WebView* webview = static_cast<WebView*>(data);
+  const char* title = static_cast<const char*>(event_info);
+  if (!title) {
+    return;
+  }
+  flutter::EncodableMap args = {
+      {flutter::EncodableValue("title"), flutter::EncodableValue(title)}};
+  webview->webview_channel_->InvokeMethod(
+      "onTitleChanged", std::make_unique<flutter::EncodableValue>(args));
+}
```
JSUYA, line 1236: Is it guaranteed that onTitleChanged is called exactly once during the page load process?
Adding the title,changed callback while keeping the existing onTitleChanged emission in OnLoadFinished makes the event fire twice with the same title on a normal page load (once via title,changed during load, once at load,finished).

packages/flutter_inappwebview/tizen/src/webview.h
```diff
@@ -127,6 +137,12 @@ class WebView : public PlatformView {
   bool texture_registered_ = false;
   bool disposed_ = false;
   Ewk_Mouse_Button_Type mouse_button_type_ = (Ewk_Mouse_Button_Type)0;
+  bool is_programmatic_navigation_ = false;
+  bool is_navigation_cancelled_ = false;
```
JSUYA, line 141: It looks that `is_navigation_cancelled_` can be replaced by utilizing `committed_url_`. Currently, `is_navigation_cancelled_` looks to be an unnecessary member.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -307,9 +307,23 @@ void WebView::StopNavigation() {
...
 }
 
+bool WebView::NavigateProgrammatically(const std::function<bool()>& ewk_call) {
+  is_programmatic_navigation_ = true;
+  const bool started = ewk_call();
+  if (!started) {
+    is_programmatic_navigation_ = false;
+  }
+  return started;
+}
```
JSUYA, line 325: um.. I am not sure if "NavigateProgrammatically" or "Programmatic" is an appropriate name for the current behavior.
And, there must be a case where a page load is called, and another page or back/forward request is made before the page load is complete. Please check if this case is also blocked.

packages/flutter_inappwebview/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+- Fix `onTitleChanged` to also fire when the page's title changes after the
+  initial load (e.g. when JavaScript updates `document.title`), instead of
+  only once when loading finishes.
+- Fix a race where `getUrl()` could return the URL of a navigation that was
+  cancelled via `shouldOverrideUrlLoading`, and skip the
+  `shouldOverrideUrlLoading` round-trip for app-initiated navigations
+  (`loadUrl`, `goBack`, `reload`, etc.).
+- Fix `scrollBy`/`getScrollX`/`getScrollY` occasionally returning a stale
+  scroll position right after `scrollTo`/`scrollBy`.
```
JSUYA, line 11: Most md files do not use this type of line break. Also, the modifications must be explained briefly. Detailed explanations should be provided in the body of the PR.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -1080,6 +1162,26 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
...
+  }
+
+  // Snapshot the URL EWK is displaying before accepting the navigation
+  // below. EWK can fire "url,changed" for the new (possibly-to-be-cancelled)
+  // URL as soon as ewk_policy_decision_use() runs, racing with the async
+  // shouldOverrideUrlLoading round-trip. StopNavigation() reverts getUrl()
+  // using this snapshot rather than whatever "url,changed" reported last, so
+  // that race can't leave getUrl() stuck on a cancelled URL.
+  const std::string url_before_navigation =
+      GetViewUrl(webview->webview_instance_);
```
JSUYA, line 1183: It is recommended to declare it just before use.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -1080,6 +1162,26 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
...
+    ewk_policy_decision_use(policy_decision);
+    return;
+  }
+
+  // Snapshot the URL EWK is displaying before accepting the navigation
+  // below. EWK can fire "url,changed" for the new (possibly-to-be-cancelled)
+  // URL as soon as ewk_policy_decision_use() runs, racing with the async
+  // shouldOverrideUrlLoading round-trip. StopNavigation() reverts getUrl()
+  // using this snapshot rather than whatever "url,changed" reported last, so
+  // that race can't leave getUrl() stuck on a cancelled URL.
```
JSUYA, line 1181: I think it is an unnecessary comment.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -1080,6 +1162,26 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
...
+
+  // A new navigation decision means any previous cancellation is stale:
+  // getUrl() should stop overriding with the old committed_url_ snapshot.
+  webview->is_navigation_cancelled_ = false;
+
+  if (webview->is_programmatic_navigation_) {
+    webview->is_programmatic_navigation_ = false;
+    ewk_policy_decision_use(policy_decision);
+    return;
+  }
```
JSUYA, line 1174: Please explain a little more why "shouldOverrideUrlLoading" should not be called.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -307,9 +307,23 @@ void WebView::StopNavigation() {
   if (disposed_ || !webview_instance_) {
     return;
   }
+  is_navigation_cancelled_ = true;
+  if (!pending_navigation_revert_url_.empty()) {
+    committed_url_ = pending_navigation_revert_url_;
+  }
+  ewk_view_resume(webview_instance_);
```
JSUYA, line 314: Why is ewk_view_resume() necessary? ewk_view_stop() is called immediately after ewk_view_resume(). If this is due to internal behavior, you can leave a comment or note with //TODO on this seemingly abnormal code.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -1089,6 +1191,8 @@ void WebView::OnNavigationPolicy(void* data, Evas_Object* obj,
     return;
   }
 
+  webview->pending_navigation_revert_url_ = url_before_navigation;
```
JSUYA, line 1194: Why is it revert_url? This simply contains the previous URL information.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -836,18 +884,28 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
...
+      return true;
+    });
     result->Success();
   } else if (method_name == "getUrl") {
-    const char* url = ewk_view_url_get(webview_instance_);
-    result->Success(url ? flutter::EncodableValue(url)
-                        : flutter::EncodableValue());
+    if (is_navigation_cancelled_ && !committed_url_.empty()) {
+      result->Success(flutter::EncodableValue(committed_url_));
+    } else {
```
JSUYA, line 904: I believe it should operate according to the internal logic of ewk_view_url_get().
I do not think there is a need to separately save and return the previous URL. If the page is cancelled while loading, I believe it is correct to return that there is no URL, as the display of that page is not normal.

packages/flutter_inappwebview/tizen/src/webview.cc
seungsoo47: Confirmed — this can indeed be blocked incorrectly.

`is_programmatic_navigation_` is a single shared flag, so it can't tell which in-flight navigation a given `policy,navigation,decide` / `load,started` / `load,finished` / `load,error` event actually belongs to. If a second programmatic call (e.g. `goBack()`) arrives while an earlier one (e.g. `loadUrl()`) is still in flight, the abandoned first load's `load,error` can clear the flag before the second call's own policy decision arrives, so that second, genuinely-programmatic navigation ends up routed through `shouldOverrideUrlLoading` as if it were user-initiated (the app could be asked to approve its own `goBack()`). The reverse case — a real user click landing while a programmatic flag is still pending — is possible for the same reason.

A clean fix needs a per-navigation correlation id, which the EWK callbacks don't provide, so neither a bool nor a simple counter can fully disambiguate which event belongs to which call without deeper engine-side cooperation. I'd like to land the other fixes from this review as-is and track this specific race as a known limitation / follow-up issue rather than block this PR on it — let me know if you'd rather address it here instead.

On naming: leaving `NavigateProgrammatically` / `is_programmatic_navigation_` as-is for now since I don't have a clearly better name that captures the caveat above; happy to revisit if you have a suggestion.

seungsoo47: Filed #1098 to track this.

seungsoo47:
I've submitted a new PR based on the main branch.
All the review comments above have been addressed.
Please review #1099.

PR #1086, opened 2026-08-13
Fix CI failures: analysis_options drift and dart_analyze lint issues

seungsoo47:
These CI failures started appearing after upgrading to the latest stable Flutter SDK, which now auto-migrates `analysis_options.yaml` on `pub get` and enforces stricter lint rules.

- Exclude build/platform dirs in analysis_options.yaml
- camera: await stopVideoRecording() in try block
- tizen_bundle: ignore false-positive collection_methods_unrelated_type
- google_maps_flutter: use super.key instead of Key? param
- flutter_inappwebview: fix print/const/super-param/mounted-check lint issues

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -16,6 +16,13 @@ analyzer:
     # Ignore generated files
     - '**/*.g.dart'
     - '**/*.mocks.dart' # Mockito @GenerateMocks
+    - 'build/**'
+    - 'android/**'
+    - 'ios/**'
+    - 'web/**'
+    - 'windows/**'
+    - 'macos/**'
+    - 'linux/**'
 
 linter:
   rules:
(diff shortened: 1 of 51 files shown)


JSUYA (commented):
Basically, you must synchronize the root's analysis_options.yaml with flutter/packages(or _plus packages)'s analysis_options.yaml.
And, you must synchronize it with the analysis_options.yaml of each package's frontend package.

If a frontend package does not support version 3.47.0, you must describe this separately and fix the Analyze CI errors based on analysis rules.
We should not create unnecessary code diffs.

Of course, Tizen packages can modify their code.

JSUYA:
https://github.com/flutter-tizen/plugins/pull/1088

JSUYA:
#1088

PR #1088, opened 2026-08-14
Support Flutter 3.47.0 (fix analyze/format CI)

JSUYA:
The dart_analyze and format CI jobs have been failing since the stable Flutter 3.47.0 upgrade.
 
The root analysis_options.yaml is synced with the latest copy from flutter/packages, and the 4 diagnostics newly reported on 3.47.0 (use_super_parameters, collection_methods_unrelated_type, strict_top_level_inference, unintended_html_in_doc_comment) are suppressed with temporary rules marked TEMPORARY instead of code changes, to be re-enabled in per-package follow-up PRs.

The seven directory excludes are pre-applied to the 14 packages that have their own analysis_options.yaml so that the pub get migration becomes a no-op, and for flutter_inappwebview, temporary lint-disabling rules are added to the example's analysis_options.yaml instead of modifying the example code copied from upstream.

The sqflite and video_player packages are reformatted with the output of `dart format` from Dart 3.13.0, and video_player's two unused flutter_driver test files are removed in sync with flutter/packages#4347.

The CI format check now excludes the discontinued packages, which must not be modified, and the whole change set is verified with Dart 3.13.0 (the same version as CI): all packages pass `dart analyze --fatal-infos` and a repo-wide `dart format` rerun yields zero changes.


JSUYA:
instead of #1086

seungsoo47 (approved):

PR #1099, opened 2026-08-21
[flutter_inappwebview] Fix navigation/getUrl/scroll bugs and add integration tests

seungsoo47:
- Fix `onTitleChanged` to fire on later title changes (e.g. via JS document.title), not just once after load.
- Fix a getUrl() race with cancelled navigations, and skip the shouldOverrideUrlLoading round-trip for app-initiated navigations (loadUrl, goBack, reload, etc.).
- Fix shouldOverrideUrlLoading not being invoked for a hardware/remote Back-key navigation.
- Fix scrollBy/getScrollX/getScrollY occasionally returning a stale position right after scrollTo/scrollBy (EWK applies scroll asynchronously).
- Add Tizen integration tests ported from upstream flutter_inappwebview v6.1.5.
- Bump flutter_inappwebview_tizen to 0.2.0.

### DIFF
--- a/packages/flutter_inappwebview/tizen/src/webview.cc
+++ b/packages/flutter_inappwebview/tizen/src/webview.cc
@@ -971,11 +971,19 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
     }
     if (method_name == "scrollTo") {
       ewk_view_scroll_set(webview_instance_, x, y);
+      target_scroll_x_ = x;
+      target_scroll_y_ = y;
     } else {
-      ewk_view_scroll_by(webview_instance_, x, y);
+      int32_t current_x = 0, current_y = 0;
+      ewk_view_scroll_pos_get(webview_instance_, &current_x, &current_y);
+      int32_t base_x = (target_scroll_x_ >= 0) ? target_scroll_x_ : current_x;
+      int32_t base_y = (target_scroll_y_ >= 0) ? target_scroll_y_ : current_y;
+      target_scroll_x_ = base_x + x;
+      target_scroll_y_ = base_y + y;
+      ewk_view_scroll_set(webview_instance_, target_scroll_x_, target_scroll_y_);
     }
-    int32_t new_x = 0, new_y = 0;
-    ewk_view_scroll_pos_get(webview_instance_, &new_x, &new_y);
+    int32_t new_x = target_scroll_x_;
+    int32_t new_y = target_scroll_y_;
     flutter::EncodableMap args = {
         {flutter::EncodableValue("x"), flutter::EncodableValue(new_x)},
         {flutter::EncodableValue("y"), flutter::EncodableValue(new_y)},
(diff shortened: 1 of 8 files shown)


packages/flutter_inappwebview/tizen/src/webview.h
```diff
@@ -127,6 +137,12 @@ class WebView : public PlatformView {
   bool texture_registered_ = false;
   bool disposed_ = false;
   Ewk_Mouse_Button_Type mouse_button_type_ = (Ewk_Mouse_Button_Type)0;
+  bool is_programmatic_navigation_ = false;
+  bool is_navigation_cancelled_ = false;
```
JSUYA, line 144: It looks that is_navigation_cancelled_ can be replaced by utilizing committed_url_. Currently, is_navigation_cancelled_ looks to be an unnecessary member.

packages/flutter_inappwebview/tizen/src/webview.cc
```diff
@@ -836,18 +886,28 @@ void WebView::HandleWebViewMethodCall(const FlMethodCall& method_call,
...
-    const char* url = ewk_view_url_get(webview_instance_);
-    result->Success(url ? flutter::EncodableValue(url)
-                        : flutter::EncodableValue());
+    if (is_navigation_cancelled_ && !committed_url_.empty()) {
+      result->Success(flutter::EncodableValue(committed_url_));
+    } else {
+      const char* url = ewk_view_url_get(webview_instance_);
+      result->Success(url ? flutter::EncodableValue(url)
+                          : flutter::EncodableValue());
+    }
```
JSUYA, line 932: I believe it should operate according to the internal logic of ewk_view_url_get().
I do not think there is a need to separately save and return the previous URL. If the page is cancelled while loading, I believe it is correct to return that there is no URL, as the display of that page is not normal.

packages/flutter_inappwebview/tizen/src/webview.h
seungsoo47: This isn't interchangeable with committed_url_. During a normal (non-cancelled) navigation, committed_url_ still holds the previous URL until "url,changed" fires — but getUrl() needs to query EWK live during that window, since Android/iOS both report the in-progress URL there, not the last-committed one. is_navigation_cancelled_ is what tells getUrl() "don't trust EWK's live value right now"; committed_url_ alone can't signal that, since it's non-empty almost the entire time after the first page load, so !committed_url_.empty() can't distinguish a cancelled window from an ordinary in-flight navigation.

packages/flutter_inappwebview/tizen/src/webview.cc
seungsoo47: Checked ewk_view_url_get() in the chromium-ewk/ewk_view.h: it returns the view's internal "current URL", only guaranteed accurate after "url,changed" fires. This code has to call ewk_policy_decision_use() immediately (Tizen's policy-decision API needs a synchronous answer) and only then suspend the view to emulate waiting on Dart's async shouldOverrideUrlLoading response (see the comment above at L1182-1184) — so if the navigation gets cancelled while suspended, ewk_view_url_get() can already be reporting the destination URL that was never actually displayed. Returning "no URL" there would also break parity with Android's WebView.getUrl()/iOS's WKWebView.url, both of which keep reporting the previous URL when a load is stopped, since that's still what's on screen. Returning the last committed URL here is intentional.

JSUYA (approved):

PR #1106, opened 2026-08-25
[device_info_plus] Add duid getter

seungsoo47:
Add DeviceInfoPluginTizen.duid to read the device's DUID via vconf. This is a Tizen-specific API with no upstream counterpart in device_info_plus.

### DIFF
--- a/packages/device_info_plus/README.md
+++ b/packages/device_info_plus/README.md
@@ -10,7 +10,7 @@ Add `device_info_plus_tizen` as a dependency in your `pubspec.yaml` file.
 
 ```yaml
 dependencies:
-  device_info_plus_tizen: ^1.4.2
+  device_info_plus_tizen: ^1.4.3
 ```
 
 Then you can import `device_info_plus_tizen` in your Dart code.
--- a/packages/device_info_plus/README.md
+++ b/packages/device_info_plus/README.md
@@ -26,31 +26,32 @@ String modelName = tizenInfo.modelName;
 
 ## Supported properties
 
-| Property | Feature or system key |
-|-|-|
-| `modelName` | `http://tizen.org/system/model_name` |
-| `cpuArch` | `http://tizen.org/feature/platform.core.cpu.arch` |
-| `nativeApiVersion` | `http://tizen.org/feature/platform.native.api.version` |
-| `platformVersion` | `http://tizen.org/feature/platform.version` |
-| `webApiVersion` | `http://tizen.org/feature/platform.web.api.version` |
-| `profile` | `http://tizen.org/feature/profile` |
-| `buildDate` | `http://tizen.org/system/build.date` |
-| `buildId` | `http://tizen.org/system/build.id` |
-| `buildString` | `http://tizen.org/system/build.string` |
-| `buildTime` | `http://tizen.org/system/build.time` |
-| `buildType` | `http://tizen.org/system/build.type` |
-| `buildVariant` | `http://tizen.org/system/build.variant` |
-| `buildRelease` | `http://tizen.org/system/build.release` |
-| `deviceType` | `http://tizen.org/system/device_type` |
-| `manufacturer` | `http://tizen.org/system/manufacturer` |
-| `platformName` | `http://tizen.org/system/platform.name` |
-| `platformProcessor` | `http://tizen.org/system/platform.processor` |
-| `tizenId` | `http://tizen.org/system/tizenid` |
-| `freeDiskSize` | `storage_get_internal_memory_size()` |
-| `totalDiskSize` | `storage_get_internal_memory_size()` |
-| `physicalRamSize` | `runtime_info_get_system_memory_info()` |
-| `availableRamSize` | `runtime_info_get_system_memory_info()` |
-| `screenWidth` | `http://tizen.org/feature/screen.width` |
-| `screenHeight` | `http://tizen.org/feature/screen.height` |
+| Property | Feature or system key | TV only |
+|-|-|-|
+| `modelName` | `http://tizen.org/system/model_name` | |
+| `cpuArch` | `http://tizen.org/feature/platform.core.cpu.arch` | |
+| `nativeApiVersion` | `http://tizen.org/feature/platform.native.api.version` | |
+| `platformVersion` | `http://tizen.org/feature/platform.version` | |
+| `webApiVersion` | `http://tizen.org/feature/platform.web.api.version` | |
+| `profile` | `http://tizen.org/feature/profile` | |
+| `buildDate` | `http://tizen.org/system/build.date` | |
+| `buildId` | `http://tizen.org/system/build.id` | |
+| `buildString` | `http://tizen.org/system/build.string` | |
+| `buildTime` | `http://tizen.org/system/build.time` | |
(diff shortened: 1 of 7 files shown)


packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -23,6 +24,30 @@ namespace {
...
+  if (!handle) {
+    LOG_ERROR("Failed to open libvconf.so.0.3.1.");
+    return "";
+  }
+
+  std::string duid;
+  auto vconf_get_str =
+      reinterpret_cast<FuncVconfGetStr>(dlsym(handle, "vconf_get_str"));
+  if (vconf_get_str) {
+    if (char *value = vconf_get_str("db/comss/duid")) {
```
JSUYA: We need to check whether this path is maintained and managed identically in the common profile as well.
In the case of vconf_get_str used in in_app_purchase, it is used only in the TV profile, but this package is also used in the common profile.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -23,6 +24,30 @@ namespace {
...
 constexpr int64_t kUnknownMetric = -1;
 
+std::string GetDuid() {
+  using FuncVconfGetStr = char *(*)(const char *);
+
+  void *handle = dlopen("libvconf.so.0.3.1", RTLD_LAZY);
+  if (!handle) {
+    LOG_ERROR("Failed to open libvconf.so.0.3.1.");
+    return "";
+  }
```
JSUYA: I am well aware that it is used to the same way in `in_app_purchase`, but since `libvconf.so.3.1` includes the detailed version, it may be vulnerable to future changes. What about `libvconf.so.0`?
Would it actually be more vulnerable to potential issues regarding management or changes?

packages/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -77,4 +77,25 @@ void main() {
     expect(rebuilt.screenWidth, tizenInfo.screenWidth);
     expect(rebuilt.totalDiskSize, tizenInfo.totalDiskSize);
   }, skip: !Platform.isLinux);
+
+  // Tizen-only: duid has no upstream counterpart in device_info_plus.
+  test('duid does not throw and is a nullable string', () async {
+    final deviceInfoPlugin = DeviceInfoPluginTizen();
+    final duid = await deviceInfoPlugin.duid;
+    expect(duid, anyOf(isNull, isA<String>()));
```
JSUYA: `expect(duid, anyOf(isNull, isA<String>()))` holds for every possible value of a `Future<String?>`, so this assertion can never failed.
```
test('duid does not throw and is a non-empty string when available', () async {
    final deviceInfoPlugin = DeviceInfoPluginTizen();
    final duid = await deviceInfoPlugin.duid;
    if (duid != null) {
      expect(duid, isNotEmpty);
    }
  }, skip: !Platform.isLinux);
```

packages/device_info_plus/lib/device_info_plus_tizen.dart
```diff
@@ -183,4 +188,13 @@ class DeviceInfoPluginTizen {
...
       _cachedTizenDeviceInfo ??= await _platform.tizenInfo();
+
+  /// This information does not change from call to call. Cache it.
+  Future<String?>? _cachedDuid;
+
+  /// The DUID (Device Unique ID) that uniquely identifies this device.
+  ///
+  /// Returns `null` if the DUID is unavailable, for example because the
+  /// underlying `db/comss/duid` vconf key could not be read.
+  Future<String?> get duid => _cachedDuid ??= _platform.getDuid();
```
JSUYA: Just curios: If conditions are favorable for success after the first failure, the failure case is always cached. Is this intended behavior? Would that scenario never occur?

packages/device_info_plus/README.md
JSUYA: Please add the API you want to add in the table. If it is TV only, please add a new column.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
seungsoo47: It has been changed to be used only in the TV profile.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
seungsoo47: Thanks.
I have changed to libvconf.so.0.

packages/device_info_plus/lib/device_info_plus_tizen.dart
seungsoo47: It has been changed to return nullopt only when vconf key couldn't be read.
Therefore, in that case, a retry will be attempted next time.

packages/device_info_plus/README.md
seungsoo47: Thanks. Done

packages/device_info_plus/example/integration_test/device_info_plus_test.dart
seungsoo47: Thanks. Done

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -23,6 +25,55 @@ namespace {
...
+  char *value = vconf_get_str("db/comss/duid");
+  dlclose(handle);
+  if (!value) {
+    LOG_ERROR("Failed to read db/comss/duid.");
+    return std::nullopt;
+  }
+  std::string duid = value;
+  free(value);
+
+  LOG_ERROR("##### duid: %s", duid.data());
```
JSUYA: Please remove this line.

packages/device_info_plus/example/integration_test/device_info_plus_test.dart
```diff
@@ -77,4 +77,24 @@ void main() {
...
+    final first = await plugin1.duid;
+    final second = await plugin1.duid;
+    expect(second, first);
+    if (first != null) {
+      expect(second, same(first));
+    }
+
+    final plugin2 = DeviceInfoPluginTizen();
+    final third = await plugin2.duid;
+    expect(third, first);
```
JSUYA, line 100: `plugin2` does not get the cached duid because it is a new instance. This test looks invalid.

packages/device_info_plus/tizen/src/device_info_plus_tizen_plugin.cc
```diff
@@ -23,6 +25,55 @@ namespace {
...
+  int ret = system_info_get_platform_string("http://tizen.org/feature/profile",
+                                            &profile);
+  if (ret != SYSTEM_INFO_ERROR_NONE) {
+    LOG_ERROR("Failed to get the platform profile: %s", get_error_message(ret));
+    return false;
+  }
+  std::string profile_str(profile);
+  free(profile);
+  return profile_str == "tv";
+}
```
JSUYA: Other plugins use `TV_PROFILE` macro.

JSUYA (approved):

PR #1136, opened 2026-09-03
[tizen_window_manager] Add tizen-core-wl window backend

JSUYA:
Introduce a WindowProxyInterface abstraction with two backends selected at runtime by WindowProxy:
  - TizenCoreWlWindowProxy (tizen-core-wl) on Tizen API version >= 11.0 when the library is available
  - EcoreWl2WindowProxy (ecore_wl2) otherwise, including as a fallback when tizen-core-wl cannot be loaded on 11.0+

Both backends implement WindowProxyInterface (GetGeometry/Activate/Lower) and resolve their symbols at runtime via dlopen/dlsym. TizenWindowManager constructs WindowProxy, which picks the backend by querying the platform version at startup.

### DIFF
--- a/packages/tizen_window_manager/CHANGELOG.md
+++ b/packages/tizen_window_manager/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.2
+
+* Add tizen-core-wl window backend for Tizen 11.0 and later.
+
 ## 0.1.1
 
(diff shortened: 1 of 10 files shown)


JSUYA:
https://github.com/flutter-tizen/plugins/pull/1137

PR #1137, opened 2026-09-04
[tizen_window_manager] Use the embedder window channel

JSUYA:
The native code called the window APIs(ecore) directly, which only worked on the window backend the embedder happened to be built with.
Send the requests to the `tizen/internal/window` method channel instead and remove the native code.
The public Dart API is unchanged.


seungsoo47:
Could you document the minimum required flutter-tizen version in the README? With this change, activate() and lower() require flutter-tizen 3.35.3-tizen.1.1.0 or later and throw MissingPluginException on older SDKs. The Flutter version constraint in pubspec.yaml alone cannot enforce this requirement.

JSUYA:
> Could you document the minimum required flutter-tizen version in the README? With this change, activate() and lower() require flutter-tizen 3.35.3-tizen.1.1.0 or later and throw MissingPluginException on older SDKs. The Flutter version constraint in pubspec.yaml alone cannot enforce this requirement.

I added comments and explanations.

seungsoo47 (approved):

Issue #1089, opened 2026-08-14
Re-enable analyzer rules temporarily disabled for Flutter 3.47.0


JSUYA:
To unblock the analyze CI after the stable Flutter 3.47.0 upgrade (#1088), five diagnostics newly reported by the Dart 3.13.0 analyzer were temporarily suppressed in the root `analysis_options.yaml` instead of modifying code. Each suppression is marked with a `TODO(jsuya): Different from flutter/packages` comment referencing this issue.

This issue tracks re-enabling them. Each task below removes one suppression after fixing the affected package(s) in a follow-up PR.

### 1. `use_super_parameters` — google_maps_flutter
  - [x] Sync the example app with the upstream flutter/packages example, which already uses `super.key` (19 files under `example/lib/`), then re-enable the rule.

### 2. `collection_methods_unrelated_type` — tizen_bundle
  - [x] The two reports in `example/integration_test/tizen_bundle_test.dart` are false positives from intentional null-handling tests (`bundle[null]`, `bundle.remove(null)`). Add `// ignore: collection_methods_unrelated_type` to those lines (first-party code), then re-enable the rule.

### 3. `strict_top_level_inference` — video_player_avplay, video_player_videohole
  - [x] Add the missing parameter type in both packages (`void _afterFrameLayout(Duration _)` in `lib/video_player.dart`), then re-enable the rule.

### 4. `unintended_html_in_doc_comment` — in_app_purchase
  - [x] Wrap the angle-bracketed types in backticks in two doc comments of `lib/src/in_app_purchase_tizen_platform.dart` (`Map<Object?, Object?>?` → `` `Map<Object?, Object?>?` ``), then re-enable the rule.

## Deferred upstream settings (repo-wide migrations)

The following settings from the upstream flutter/packages `analysis_options.yaml` are also disabled with the same TODO markers. They require repo-wide migration rather than per-package fixes.

### 5. `omit_obvious_local_variable_types` — repo-wide
  - [x] Migrate the existing code, which still follows the old `always_specify_types` style (hundreds of violations), then enable the rule.

### 6. `formatter: page_width: 100` — repo-wide
  - [x] Reformat the repository with a line length of 100, then enable the setting.

JSUYA:
done

PR #1139, opened 2026-09-04
[tizen_bundle] Handle non-String keys in operator[] and remove()

JSUYA:
Map takes an Object? key and must not throw for unrelated keys, but Bundle guarded null only and cast the rest with `as String`.

Also re-enable the collection_methods_unrelated_type lint.

https://github.com/flutter-tizen/plugins/issues/1089


seungsoo47 (approved):
Please update README.

PR #1142, opened 2026-09-07
[webview_flutter_tizen] Add WV API backend

seungsoo47:
Add a WV API backend behind the existing `WebViewBackend` interface, alongside EWK. Backend is auto-selected by Tizen platform version: EWK through 10.0, EWK wrapper mode on 10.1, WV standalone from 11.0.


packages/webview_flutter/tizen/src/ewk_internal_api_binding.h
```diff
@@ -183,6 +185,8 @@ class EwkInternalApiBinding {
   EwkInternalApiBinding();
 
   void* handle_ = nullptr;
+  bool initialized_ = false;
+  bool initialize_result_ = false;
```
JSUYA: I think you can use `initialize_result_` instead of `initialized_`.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+void WvWebViewBackend::EvaluateJavaScript(
+    const std::string& javascript, std::function<void(const char*)> callback) {
+  auto* callback_ptr =
+      new std::function<void(const char*)>(std::move(callback));
+  if (!WvInternalApiBinding::GetInstance().view.ScriptExecute(
+          view_, javascript.c_str(), &WvWebViewBackend::OnEvaluateJavaScript,
+          callback_ptr)) {
+    LOG_WARN("wv_view_script_execute failed.");
+    (*callback_ptr)(nullptr);
+    delete callback_ptr;
```
JSUYA, line 437: If `ScriptExecute()` returns null, `(*callback_ptr)(nullptr);` is called, which finally calls `raw_result->Success();`. Please verify this along with the EWK logic.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  wv_key_event_s key_event = {};
+  key_event.key_name = key;
+  key_event.key = key;
+  key_event.string = string;
+  key_event.compose = compose;
```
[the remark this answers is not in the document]
JSUYA: It seems like key_event.modifiers is missing. Please check it.

packages/webview_flutter/tizen/src/wv_webview_backend.h
```diff
@@ -0,0 +1,130 @@
...
+  // before engine init. `standalone` adds --enable-wv-standalone to that
+  // argv. Returns whether wv_init() itself succeeded.
+  static bool GlobalInitialize(bool standalone);
+
+  // Must be called exactly once, after every WvWebViewBackend has been
+  // destroyed.
+  static void GlobalShutdown();
+
+ private:
+  static void FlushPendingTeardowns();
```
JSUYA: The implementation looks identical to ewk. If you can make it into a single method, it is better to do so.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+      [](const void* key1, int key1_length, const void* key2,
+         int key2_length) -> int {
+        return strcmp(static_cast<const char*>(key1),
+                      static_cast<const char*>(key2));
+      },
+      EINA_KEY_HASH(eina_hash_superfast), [](void* data) { free(data); }, 10);
+  for (const auto& header : headers) {
+    eina_hash_add(wv_headers, header.first.c_str(),
+                  strdup(header.second.c_str()));
+  }
```
JSUYA: I think you can create common functions by creating a mixin class and using multiple inheritance.

```
template <typename Backend>
class WebViewBackendMixin  {} ...
```

```
class WvWebViewBackend : public WebViewBackend,
                         public WebViewBackendMixin<WvWebViewBackend> {
```

or maybe, you can make common method within WebViewBackend class.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  // can drop it by shortening argc; wv_init() reads that switch to decide
+  // whether to run the WV implementation or forward everything to ewk_*.
+  const char* argv[] = {
+      "--disable-pinch", "--js-flags=--expose-gc", "--single-process",
+      "--no-zygote",     "--enable-wv-standalone",
+  };
+  int argc = sizeof(argv) / sizeof(argv[0]);
+  if (!standalone) {
+    --argc;
+  }
```
JSUYA: I think this code bad readability. I believe it would be better to add "--enable-wv-standalone" when standalone is true.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  // wv_set_arguments() returns a TIZEN_ERROR_* code (0 == success), while
+  // wv_init() returns ewk_init()'s reference count (> 0 == success).
+  int result = wv.main.SetArguments(argc, argv);
+  if (result != 0) {
+    LOG_WARN("wv_set_arguments() returned %d.", result);
+  }
```
JSUYA, line 98: It seems more appropriate not to call Init if SetArguments fails.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  result = wv.main.Init();
+  if (result <= 0) {
+    LOG_WARN("wv_init() returned %d.", result);
```
JSUYA: If a problem occurs during creation due to failure, it should be LOG_ERROR, not LOG_WARN.

Also, please delete all LOG_INFOs.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  const char* argv[] = {
+      "--disable-pinch", "--js-flags=--expose-gc", "--single-process",
+      "--no-zygote",     "--enable-wv-standalone",
+  };
+  int argc = sizeof(argv) / sizeof(argv[0]);
+  if (!standalone) {
+    --argc;
+  }
+  // wv_set_arguments() returns a TIZEN_ERROR_* code (0 == success), while
+  // wv_init() returns ewk_init()'s reference count (> 0 == success).
```
JSUYA: This is an unnecessary comment.

packages/webview_flutter/tizen/src/webview_backend_factory.cc
```diff
@@ -0,0 +1,110 @@
...
+    std::sscanf(value, "%d.%d", &major, &minor);
+    free(value);
+  }
+  if (major >= 11) {
+    return BackendKind::kWvStandalone;
+  }
+  if (major == 10 && minor >= 1) {
+    return BackendKind::kEwkWrapper;
+  }
+  return BackendKind::kEwk;
```
JSUYA, line 31: Please use https://pub.dev/packages/flutter_tizen#getting-api-version API

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  // wv_set_arguments() stores the argv used by wv_init(), so it must run
+  // before the engine boots. --enable-wv-standalone comes last so wrapper mode
+  // can drop it by shortening argc; wv_init() reads that switch to decide
+  // whether to run the WV implementation or forward everything to ewk_*.
+  const char* argv[] = {
+      "--disable-pinch", "--js-flags=--expose-gc", "--single-process",
+      "--no-zygote",     "--enable-wv-standalone",
+  };
+  int argc = sizeof(argv) / sizeof(argv[0]);
+  if (!standalone) {
```
JSUYA: standalone is always true in WvWebViewBackend::GlobalInitialize().
Therefore, the `bool standalone` parameter is not needed.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  LOG_INFO("wv_init() returned %d.", result);
+  return true;
+}
+
+void WvWebViewBackend::GlobalShutdown() {
+  FlushPendingTeardowns();
+  WvInternalApiBinding::GetInstance().main.Shutdown();
+}
+
+void WvWebViewBackend::FlushPendingTeardowns() {
```
JSUYA: I think duplicate functions can be combined into a single common interface.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  wv_settings_h settings = wv.view.SettingsGet(view_);
+  wv.settings.ImePanelEnabledSet(settings, true);
+  wv.settings.ForceZoomSet(settings, true);
+  wv.view.ImeWindowSet(view_, window_);
+  wv.view.KeyEventsEnabledSet(view_, true);
+#ifdef WEBVIEW_TIZEN_TOUCH_EVENTS_ENABLED
+  wv.view.TouchEventsEnabledSet(view_, 1);
+  wv.view.MouseEventsEnabledSet(view_, false);
+#else
+  wv.view.TouchEventsEnabledSet(view_, 0);
```
JSUYA: Is 1, 0 correct? Please check if it is true or false.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  }
+}
+
+bool WvWebViewBackend::Create(double width, double height, void* window,
+                              bool engine_policy) {
+  window_ = window;
+
+  if (engine_policy) {
+    LOG_WARN("engine_policy is not supported by the WV backend; ignored.");
+  }
```
JSUYA: This is an unnecessary log.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
```diff
@@ -0,0 +1,724 @@
...
+  wv.view.AddCallback(view_, "console,message",
+                      &WvWebViewBackend::OnConsoleMessage, this);
+  wv.view.AddCallback(view_, "policy,navigation,decide",
+                      &WvWebViewBackend::OnNavigationPolicy, this);
+  wv.view.AddCallback(view_, "policy,response,decide",
+                      &WvWebViewBackend::OnResponsePolicy, this);
+  wv.view.AddCallback(view_, "url,changed", &WvWebViewBackend::OnUrlChange,
+                      this);
+
+  wv.view.Resize(view_, static_cast<int>(width), static_cast<int>(height));
```
JSUYA: This is just just concern, but data loss may occur when converting a double type to an int using static_case. In most cases, you need to use round().

JSUYA (commented):
Please delete all unnecessary comments written throughout the code.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
seungsoo47: Thanks for the review.

WV already guards it (checks the return value, force-completes the callback with nullptr on failure). EWK didn't: ewk_view_script_execute also returns Eina_Bool and skips the callback on failure (chromium-efl ewk_view.h), so the old code would leave the Dart Future hanging forever. Fixed EWK to match WV's pattern.

runJavaScriptReturningResult() has no documented failure contract in the platform interface, and Android's own implementation does the exact same thing: evaluateJavascript() returning null becomes '', never an Error. None of the Dart-side callers here catch a PlatformException from this call either, so surfacing Error() would just be an uncaught exception, not a handled failure.

packages/webview_flutter/tizen/src/wv_webview_backend.cc
seungsoo47: WebViewBackendFactory passes kind == BackendKind::kWvStandalone, which is false for kEwkWrapper (Tizen 10.1). The parameter is needed to distinguish the two modes.

seungsoo47:
> Please delete all unnecessary comments written throughout the code.

I've addressed all the review comments above, thanks!

packages/webview_flutter/tizen/src/ewk_webview_backend.cc
```diff
@@ -0,0 +1,681 @@
...
+    return false;
+  }
+
+  Evas* evas = evas_object_evas_get(view_);
+  SyncEvasModifiers(evas, modifiers);
+  Evas_Modifier* evas_modifiers =
+      const_cast<Evas_Modifier*>(evas_key_modifier_get(evas));
+  Evas_Lock* evas_locks = const_cast<Evas_Lock*>(evas_key_lock_get(evas));
+
+  // TODO(swift-kim): Deal with other members of the structure.
```
JSUYA: Remove this comment

packages/webview_flutter/tizen/src/pending_teardown.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include "buffer_pool.h"
+#include "log.h"
+
+// Tracks views whose native destroy call has not run yet, shared by the EWK
+// and WV backends (both defer that call until after texture unregistration,
+// via g_timeout, and both must drain this before their engine shuts down —
+// EWK's ewk_shutdown() fatally CHECKs if any Ewk_View is still alive).
+template <typename Handle>
+class PendingTeardownRegistry {
```
JSUYA: This is quite different from my intention. I did not intend to create a class solely for Teardown, as I believe WebViewBackend can absorb this or declare it as an additional class within webview_backend.h.

Since WebViewBackend can have a common interface for both backends, I suggested a method to reduce duplicate code.

packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -1,70 +1,48 @@
-// Copyright 2021 Samsung Electronics Co., Ltd. All rights reserved.
+// Copyright 2026 Samsung Electronics Co., Ltd. All rights reserved.
```
JSUYA: Since it is an existing file, this is not necessary.

packages/webview_flutter/tizen/src/webview_backend_factory.cc
```diff
@@ -0,0 +1,99 @@
...
+
+namespace {
+
+enum class BackendKind { kEwk, kEwkWrapper, kWvStandalone };
+
+BackendKind DefaultBackendForPlatform() {
+  int major = 0, minor = 0;
+  if (const char* value = std::getenv("TIZEN_API_VERSION")) {
+    std::sscanf(value, "%d.%d", &major, &minor);
+  }
```
JSUYA, line 24: Please, do not use the TIZEN_API_VERSION environment variable; instead, use the API provided by the flutter-tizen plugin.

packages/webview_flutter/tizen/src/webview_backend.h
```diff
@@ -0,0 +1,108 @@
...
+  virtual void SetScrollbarVisible(bool visible) = 0;
+  virtual bool ClearCookies() = 0;
+
+ protected:
+  // Keeps |pool| alive and defers |destroy| until the returned closure runs.
+  static std::function<void()> RegisterPendingTeardown(
+      std::shared_ptr<BufferPool> pool, std::function<void()> destroy);
+
+  // Must be called before the engine shuts down.
+  static void FlushPendingTeardowns();
```
JSUYA: unnecessary comments

packages/webview_flutter/tizen/src/webview_backend_factory.cc
```diff
@@ -0,0 +1,99 @@
...
+  return kind;
+}
+
+bool g_wv_engine_initialized = false;
+
+}  // namespace
+
+std::unique_ptr<WebViewBackend> WebViewBackendFactory::Create(
+    WebViewBackend::Delegate* delegate) {
+  BackendKind kind = SelectedBackend();
```
JSUYA: SelectedBackend() is called in three places. I believe it can be stored as a single variable at startup.

packages/webview_flutter/README.md
```diff
@@ -61,6 +61,19 @@ class _WebViewExampleState extends State<WebViewExample> {
...
+The plugin chooses a backend from the platform version reported by the device:
+
+| Tizen version    | Backend                     |
+|------------------|-----------------------------|
+| 10.0 and earlier | EFL WebKit (EWK)            |
+| 10.1             | WV API, EWK wrapper mode    |
+| 11.0 and later   | WV API, standalone mode     |
+
+The WV backends are experimental. `WebViewController.tizenEnginePolicy` has no
+WV equivalent, so it is ignored (with a warning) on Tizen 10.1 and later.
```
JSUYA: No newline

packages/webview_flutter/tizen/src/webview_backend_factory.cc
JSUYA: invalid review, please ignore this comment

JSUYA (approved):

PR #1148, opened 2026-09-10
[tizen_stt] Add STT plugin

JSUYA:
Add speech-to-text support for Samsung TVs using the Tizen STT engine. Supports recognition results, language selection, and session control.


JSUYA:
While recording is possible by pressing the microphone button on the remote control, the user experience is compromised because Bixby launches simultaneously.
I will revisit this once there is a concrete usage plan or a way to perform recording independently. In the meantime, I will keep this as a draft.

PR #1150, opened 2026-09-14
[in_app_purchase] Fix SSO login info handling

seungsoo47:
* Fix the return type of `sso_get_login_info` and clear the login info after use.
* Fix a crash and a memory leak when reading the country code.
* Report an error from `getCustomId` when the SSO login info is unavailable.
* Load `libvconf` and `libcapi-system-info` by their major SONAME.


packages/in_app_purchase/tizen/src/billing_manager.h
```diff
@@ -18,6 +19,7 @@
 #include "rapidjson/document.h"
 
 #define SSO_API_MAX_STRING_LEN 128
+#define SSO_TRUE 0
```
xiaowei-guan: It is suggested to change "SSO_TRUE" to  "SSO_SUCCESS"

xiaowei-guan (approved):

PR #1151, opened 2026-09-16
[video_player_avplay] Replace Ecore pipes with GLib

JSUYA:
Replace Ecore pipes with GLib idle callbacks.


packages/video_player_avplay/tizen/src/video_player.cc
```diff
@@ -15,25 +15,21 @@ static int64_t player_index = 1;
...
+    g_source_remove(sink_event_source_);
   }
 }
 
 void VideoPlayer::ClearUpEventChannel() {
   is_initialized_ = false;
-  event_sink_ = nullptr;
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    event_sink_ = nullptr;
```
gin7773: In `SetUpEventChannel()`, `OnListen` assigns `event_sink_` and `OnCancel` clears it without taking the same lock. Would it be better to protect these operations with the same mutex as well?

xiaowei-guan (approved):

PR #1152, opened 2026-09-16
[video_player_avplay] Handle consecutive seekTo calls

gin7773:
Main changes:

Track active and pending `seek` operations in the Dart FFI implementation so only one `native seek` is in flight per player. Coalesce consecutive `seekTo` requests to the latest `pending` position, complete all waiting `futures` when the `seek` finishes, and fail `pending seeks` when the player is `disposed` or reports an error.

Send a `seekCompleted` event from native code so Dart can complete the `active seek` and start any `pending seek`.


packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan, line 388: do we need cancel seek when suspend?

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +131,85 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
+
+  void _handleSeekCompleted(int playerId) {
+    final _SeekOperation? op = _activeSeeks.remove(playerId);
+    if (op != null && !op.completer.isCompleted) {
+      op.completer.complete();
+    }
+    _startPendingSeekIfAny(playerId);
+  }
+
+  void _startPendingSeekIfAny(int playerId) {
```
xiaowei-guan: From the description, we only need seek the last position.

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +131,85 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
+  Future<void> _startSeek(int playerId, int position) async {
+    final _SeekOperation op = _SeekOperation(position);
+    _activeSeeks[playerId] = op;
+
+    try {
+      await _api.seekTo(
+        PositionMessage(playerId: playerId, position: position),
+      );
+    } catch (e) {
+      _completeSeekWithError(playerId, e);
```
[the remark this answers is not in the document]
gin7773: 👍Done

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +131,85 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
-    return _api.seekTo(
-      PositionMessage(playerId: playerId, position: position.inMilliseconds),
-    );
+    final int targetPosition = position.inMilliseconds;
+
+    if (_activeSeeks.containsKey(playerId)) {
+      final _SeekOperation op = _SeekOperation(targetPosition);
+      _pendingSeeks.putIfAbsent(playerId, () => <_SeekOperation>[]);
+      _pendingSeeks[playerId]!.add(op);
+      return op.completer.future;
```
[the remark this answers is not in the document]
gin7773: 👍Done

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +131,85 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
+          p.completer.completeError(error);
+        }
+      }
+    }
+  }
+
+  void _cancelAllSeeks(int playerId) {
+    final _SeekOperation? op = _activeSeeks.remove(playerId);
+    if (op != null && !op.completer.isCompleted) {
+      op.completer.completeError('Player was disposed.');
```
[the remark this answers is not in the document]
gin7773: 👍Done

packages/video_player_avplay/lib/src/video_player_tizen.dart
gin7773: Done

packages/video_player_avplay/lib/src/video_player_tizen.dart
gin7773: Done

xiaowei-guan (changes requested):
## Re-review of the latest changes (1f7b522)

Thanks for the updates. I verified each previous comment against the current head:

| Previous comment | Status |
|---|---|
| Coalesce pending seeks instead of replaying every position (Codex P2) | ✅ Fixed — only the latest pending op is retained and all waiting completers are carried over |
| Propagate native seek failures to the returned future (Codex P1) | ✅ Fixed for the synchronous-failure path via `return op.completer.future` |
| Cancel seek when suspend | ✅ Fixed — `suspend()` now calls `_cancelAllSeeks` |
| Only seek to the last position | ✅ Fixed |
| Complete the future actually returned for an active seek on dispose (Codex P2) | ❌ **Not actually fixed** — see issue 1 below |

However, I found the following issues that should be addressed before merging.

---

### 🔴 1. [Blocking] `seekTo()` future hangs forever when the player is disposed/suspended during an in-flight seek

In `_startSeek` (`lib/src/video_player_tizen.dart:180`), the returned future is gated on the Pigeon reply:

```dart
Future<void> _startSeek(int playerId, int position) async {
  ...
  await _api.seekTo(...);   // <-- never completes on dispose mid-seek
  return op.completer.future;
}
```

On the native side, the Pigeon `result` of `seekTo` is stored as `on_seek_completed_` and is only invoked when the seek actually finishes (`video_player_tizen_plugin.cc:297`). When the player is disposed mid-seek, `PlusPlayer::Dispose()` / `MediaPlayer::StopAndDestroy()` clear `on_seek_completed_` **without invoking it**, so the Dart-side `_api.seekTo` future never completes. Although `_cancelAllSeeks` correctly errors `op.completer`, `_startSeek` is stuck at the `await` forever — the future returned to the caller never completes.

**Suggested fix (Dart, preferred):** don't gate the returned future on the Pigeon reply. Drive the state machine from the method result instead:

```dart
Future<void> _startSeek(int playerId, int position) {
  final op = _SeekOperation(position);
  _activeSeeks[playerId] = op;
  _api.seekTo(PositionMessage(playerId: playerId, position: position))
      .then((_) => _handleSeekCompleted(playerId))
      .catchError((Object e) => _completeSeekWithError(playerId, e));
  return op.completer.future;
}
```

With this, `_handleSeekCompleted` is driven by the Pigeon result (idempotent with the new `seekCompleted` event, which then becomes optional), and `_cancelAllSeeks` can deliver the error to the caller immediately even if the native reply is dropped.

**Alternative fix (native):** invoke `on_seek_completed_` before clearing it in `Dispose`/`StopAndClose`/`StopAndDestroy`, so the Pigeon reply is always delivered.

### 🟡 2. Pending ops' private `completer` is never listened to — unhandled async errors on error paths

In the queue path of `seekTo` (`video_player_tizen.dart:163-175`), the caller receives `op.attach()` (a coalesced completer's future), while `op.completer.future` is never listened to by anyone. Every error path that touches it produces an unhandled asynchronous error in the app's zone:

- `_cancelAllSeeks` with a queued pending seek (dispose/suspend while scrubbing — always reproducible);
- `_completeSeekWithError` on a pending op (including each replaced intermediate op's `completer` carried in `_coalesced`).

**Suggested fix:** return `op.completer.future` to the first caller of each pending op and use `attach()` only for subsequent callers; or drop the `completer` field entirely and only keep the coalesced completer list (combined with the fix for issue 1, `_startSeek` would also `return op.attach()`).

### 🟡 3. Unrelated behavior changes bundled in this PR

`MediaPlayer::Play/Pause` and `PlusPlayer::Play` now return `true` instead of `false` for "already playing" / "not playing" (i.e., no more `PlatformException` to the app for redundant play/pause calls). This is unrelated to consecutive seek handling — please either justify it in the PR description/CHANGELOG or split it into a separate PR.


### 🟡 4. `PlusPlayer::StopAndClose` removed the `state < kReady` early return — suspend/resume regression risk

The old code returned `true` immediately when `player_state < kReady`; the new code calls `::Stop` even in `kIdle`/`kTrackSourceReady`. If `Stop` fails in those states, `Suspend()`/`Restore()` will now fail as a whole, which risks a resume regression on TV. Note that `MediaPlayer::StopAndDestroy` guards properly (stop only when `PLAYING`/`PAUSED`, unprepare only when prepared) — `PlusPlayer` should mirror that guard structure.

### 🟢 5. Minor / nits

- **Inconsistency:** `PlusPlayer::StopAndClose` sends `SendSeekCompleted()` when a seek is in flight, but `MediaPlayer::StopAndDestroy` doesn't. Moreover, since Dart's `dispose()`/`suspend()` always call `_cancelAllSeeks` first, this event is always a no-op on the Dart side — either remove it or document its purpose.
- **`restore()` doesn't cancel seeks** (`video_player_tizen.dart:278`): unlike `dispose()`/`suspend()`, an aborted seek may then be falsely reported as successful via `SendSeekCompleted` from `StopAndClose`, and a pending seek may be started on a closed player.
- **Error object is a String:** `_completeError('Player was disposed.')` — please use an `Exception` type; also `suspend()` reuses the "disposed" wording.
- **Fragile dependency on the event channel:** seek completion is now signaled through two channels (Pigeon result + `seekCompleted` event), but `_handleSeekCompleted` is only hooked to the event. `PushEvent` silently drops events when `event_sink_ == nullptr`, which would leave callers hanging. Driving completion from the Pigeon result (fix for issue 1) removes this dependency.
- **Tests:** please consider adding integration test cases in `example/integration_test/video_player_test.dart` for "consecutive `seekTo` lands on the final position only" and "dispose during an in-flight seek".
- Noise: an unrelated blank-line removal in the `bufferingUpdate` case.

---

The overall direction (single active seek + coalesced pending + event-driven completion) looks good, and the versioning (0.8.19 / CHANGELOG / README) is consistent. Requesting changes mainly for issue 1 (future hang on dispose mid-seek) and issue 2 (unhandled errors).

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +159,81 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
+    final op = _SeekOperation(position);
+    _activeSeeks[playerId] = op;
+
+    try {
+      await _api.seekTo(PositionMessage(playerId: playerId, position: position));
+    } catch (e) {
+      _completeSeekWithError(playerId, e);
+    }
+
+    return op.completer.future;
```
xiaowei-guan: 🔴 **[Blocking]** This future never completes when the player is disposed/suspended while a seek is in flight.

The Pigeon `result` of `seekTo` is stored natively as `on_seek_completed_` and is only invoked when the seek actually finishes (`video_player_tizen_plugin.cc:297`). `PlusPlayer::Dispose()` / `MediaPlayer::StopAndDestroy()` clear `on_seek_completed_` **without invoking it**, so `_api.seekTo` never completes, this line is never reached, and the future returned to the caller hangs forever — even though `_cancelAllSeeks` correctly errors `op.completer`.

Please don't gate the returned future on the Pigeon reply:

```dart
Future<void> _startSeek(int playerId, int position) {
  final op = _SeekOperation(position);
  _activeSeeks[playerId] = op;
  _api.seekTo(PositionMessage(playerId: playerId, position: position))
      .then((_) => _handleSeekCompleted(playerId))
      .catchError((Object e) => _completeSeekWithError(playerId, e));
  return op.completer.future;
}
```

Alternatively, invoke `on_seek_completed_` before clearing it on the native side so the reply is always delivered.

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -116,9 +159,81 @@ class VideoPlayerTizen extends VideoPlayerPlatform {
...
+      if (existing != null) {
+        final op = _SeekOperation(targetPosition);
+        op._coalesced.add(existing.completer);
+        op._coalesced.addAll(existing._coalesced);
+        _pendingSeeks[playerId] = op;
+        return op.attach();
+      }
+      final op = _SeekOperation(targetPosition);
+      _pendingSeeks[playerId] = op;
+      return op.attach();
```
xiaowei-guan, line 169: 🟡 `op.completer.future` of a pending op is never listened to (callers receive `attach()` futures), so every error path that calls `_completeError` on it (`_cancelAllSeeks`, `_completeSeekWithError`) raises an **unhandled asynchronous error** in the app's zone. This is always reproducible when disposing/suspending with a queued seek, and during rapid scrubbing each replaced intermediate op's `completer` leaks an unhandled error too.

Consider returning `op.completer.future` to the first caller of each pending op (use `attach()` only for subsequent callers), or dropping the `completer` field entirely and keeping only the coalesced completer list.

packages/video_player_avplay/lib/src/video_player_tizen.dart
```diff
@@ -11,18 +13,59 @@ import '../video_player_platform_interface.dart';
...
+    for (final Completer<void> c in _coalesced) {
+      if (!c.isCompleted) {
+        c.complete();
+      }
+    }
+  }
+
+  void _completeError(Object error) {
+    if (!completer.isCompleted) {
+      completer.completeError(error);
```
xiaowei-guan: 🟢 Nit: the error object passed here is a `String` (`'Player was disposed.'` in `_cancelAllSeeks`). Please use an `Exception` type so callers can catch it meaningfully. Also note `suspend()` reuses the "disposed" wording.

packages/video_player_avplay/tizen/src/media_player.cc
```diff
@@ -231,7 +231,7 @@ bool MediaPlayer::Play() {
   }
   if (state == PLAYER_STATE_PLAYING) {
     LOG_INFO("[MediaPlayer] Player already playing.");
-    return false;
+    return true;
```
xiaowei-guan, line 229: 🟡 This (and the identical `Pause` change at line 259) turns redundant play/pause calls into silent successes instead of a `PlatformException` — an unrelated behavior change not mentioned in the PR description. Please justify it in the description/CHANGELOG or split it into a separate PR.

packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -232,7 +234,7 @@ bool PlusPlayer::Play() {
     }
     return true;
   }
-  return false;
+  return true;
```
xiaowei-guan, line 237: 🟡 Same as the `MediaPlayer::Play`/`Pause` changes: returning `true` when the player is already playing is an unrelated behavior change. Please justify it in the description/CHANGELOG or split it into a separate PR.

packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -819,33 +823,37 @@ bool PlusPlayer::StopAndClose() {
...
 
   if (drm_manager_) {
     drm_manager_->StopDrmSession();
   }
 
-  if (!::Stop(player_)) {
-    LOG_ERROR("[PlusPlayer] Player fail to stop.");
-    return false;
-  }
+  if (player_state != plusplayer::State::kNone) {
```
xiaowei-guan, line 836: 🟡 Removing the `player_state < kReady` early return means `::Stop` is now called even in `kIdle`/`kTrackSourceReady`. If `Stop` fails in those states, `Suspend()`/`Restore()` fail as a whole — a suspend/resume regression risk on TV. `MediaPlayer::StopAndDestroy` guards this properly (stop only when `PLAYING`/`PAUSED`, unprepare only when prepared); please mirror that structure here.

packages/video_player_avplay/tizen/src/media_player.cc
```diff
@@ -809,36 +818,40 @@ bool MediaPlayer::StopAndDestroy() {
     return false;
   }
 
+  bool success = true;
   is_buffering_ = false;
+  on_seek_completed_ = nullptr;
```
xiaowei-guan, line 820: 🟡 Clearing `on_seek_completed_` here drops the still-pending Pigeon `result` callback of `seekTo` without invoking it, so the Dart-side `_api.seekTo` future never completes — this is the native root cause of the hanging-future issue commented in `video_player_tizen.dart`. Please invoke the callback before clearing it (or make the Dart side independent of the reply).

packages/video_player_avplay/tizen/src/plus_player.cc
```diff
@@ -819,33 +823,37 @@ bool PlusPlayer::StopAndClose() {
...
   }
 
+  bool success = true;
   is_buffering_ = false;
-  plusplayer::State player_state = GetState(player_);
-  if (player_state < plusplayer::State::kReady) {
-    LOG_INFO("[PlusPlayer] Player already stop, nothing to do.");
-    return true;
+  if (is_seeking_) {
+    SendSeekCompleted();
```
xiaowei-guan: 🟢 Dart's `dispose()`/`suspend()` always call `_cancelAllSeeks` first, so this `seekCompleted` event is always a no-op on the Dart side, and `MediaPlayer::StopAndDestroy` doesn't send it either. Please remove it or document the intended case (e.g., a native-initiated stop).

xiaowei-guan (commented):
Inline details for my previous review. Two items cannot be anchored to this diff: `restore()` doesn't cancel seeks (unchanged code, `video_player_tizen.dart:394`), and no integration tests were added for consecutive `seekTo` / dispose-during-seek.

gin7773:
> Inline details for my previous review. Two items cannot be anchored to this diff: `restore()` doesn't cancel seeks (unchanged code, `video_player_tizen.dart:394`), and no integration tests were added for consecutive `seekTo` / dispose-during-seek.

Verified on Tizen 9.0. All other comments have been addressed.

packages/video_player_avplay/tizen/src/video_player_tizen_plugin.cc
```diff
@@ -302,10 +302,11 @@ void VideoPlayerTizenPlugin::SeekTo(
     result(FlutterError("Invalid argument", "Player not found"));
     return;
   }
-  if (!player->SeekTo(msg.position(),
-                      [result]() -> void { result(std::nullopt); })) {
+  if (!player->SeekTo(msg.position(), nullptr)) {
```
xiaowei-guan, line 305: 🟢 Nit (follow-up): passing `nullptr` here means the whole `SeekCompletedCallback` / `on_seek_completed_` machinery is now dead code in both players (assignments, null-checks and clear-ups in `SeekTo`, `OnSeekDone`/`OnSeekCompleted`, `Dispose`, `StopAndClose`/`StopAndDestroy`, `OnError`). Consider removing the callback parameter and the field entirely in a follow-up PR.

xiaowei-guan (approved):
## Re-review of the latest changes (45bc293)

All comments from my previous review have been addressed — verified against the current head:

| Previous comment | Status |
|---|---|
| 🔴 `seekTo()` future hangs on dispose/suspend mid-seek | ✅ Fixed on both sides: Dart `_startSeek` returns `op.attach()` immediately (`unawaited(_runNativeSeek)`), and the native plugin now completes the Pigeon `result` as soon as the seek starts |
| 🟡 Pending ops' `completer` never listened → unhandled async errors | ✅ Fixed: `completer` field removed; only `attach()`-created waiters remain |
| 🟢 String as error object | ✅ Fixed: `StateError` with distinct messages for dispose/suspend/restore |
| 🟡 Unrelated play/pause behavior change | ✅ Documented in CHANGELOG |
| 🟡 `PlusPlayer::StopAndClose` removed the `kReady` guard | ✅ Fixed: `::Stop` guarded by `state >= kReady`, `::Close` by `state != kNone` |
| 🟡 Dropping the pending Pigeon `result` callback natively | ✅ Eliminated: callback is now `nullptr`, result is delivered immediately |
| 🟢 `SendSeekCompleted` in `StopAndClose` was a no-op | ✅ Removed; the event is now sent only from `OnSeekDone`/`OnSeekCompleted`, consistently in both players |
| `restore()` didn't cancel seeks | ✅ Fixed |

Nice touches: the `identical()` guard in `_runNativeSeek` against stale ops, and cancelling in-flight seeks on event-stream errors.

### Remaining nits (non-blocking, fine as follow-ups)

1. 🟢 The `SeekCompletedCallback` / `on_seek_completed_` machinery is now dead code — the only call site passes `nullptr` (see inline comment). Consider removing the callback parameter and all related assignments/checks in a follow-up.
2. 🟢 The `.handleError` on the event stream cancels in-flight seeks on *any* stream error (including cast failures of unrelated events). Conservative and safe — just noting the behavior.
3. 🟢 Seek completion now depends solely on the event channel; `PushEvent` silently drops events when `event_sink_ == nullptr`. Acceptable in practice since `VideoPlayerController` always listens, and dispose/suspend/restore all pre-cancel.
4. Still no automated tests for consecutive `seekTo` — noted that this was verified manually on Tizen 9.0.

LGTM — approving. Thanks for the quick iterations!

packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan: ✅ Verified fixed in 45bc293: `suspend()` now calls `_cancelAllSeeks(playerId, 'Player was suspended.')` before `_api.suspend()`.

packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan: ✅ Verified fixed in 45bc293: pending seeks coalesce to the latest `_SeekOperation`; waiting completers are carried over to the replacement op.

packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan: ✅ Fixed in 45bc293 on both sides: `_startSeek` now returns `op.attach()` immediately (`unawaited(_runNativeSeek(...))`), and the native plugin completes the Pigeon `result` right after the seek starts (callback is `nullptr`). Disposing mid-seek no longer hangs the returned future.

packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan: ✅ Fixed in 45bc293: the `completer` field is gone; only `attach()`-created waiter futures exist, all of which have listeners.

packages/video_player_avplay/lib/src/video_player_tizen.dart
xiaowei-guan: ✅ Fixed in 45bc293: `_cancelAllSeeks` now wraps the reason in `StateError`, with distinct messages for dispose/suspend/restore.

packages/video_player_avplay/tizen/src/media_player.cc
xiaowei-guan: ✅ Acceptable now — the behavior change is documented in CHANGELOG: "Ignore redundant play/pause calls instead of reporting an error."

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: ✅ Same as the `MediaPlayer` change — covered by the new CHANGELOG entry.

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: ✅ Fixed in 45bc293: `::Stop` is now guarded by `player_state >= kReady` and `::Close` by `state != kNone`, matching `MediaPlayer`'s guard philosophy.

packages/video_player_avplay/tizen/src/media_player.cc
xiaowei-guan: ✅ Obsolete in 45bc293: `VideoPlayerTizenPlugin::SeekTo` now passes `nullptr` and returns `result(std::nullopt)` immediately, so no pending callback can be dropped. (Left a follow-up nit about removing the now-dead callback machinery.)

packages/video_player_avplay/tizen/src/plus_player.cc
xiaowei-guan: ✅ Fixed in 45bc293: `SendSeekCompleted` removed from `StopAndClose`; the event is now sent only from `OnSeekDone`/`OnSeekCompleted` — consistent in both players.

Issue #1135, opened 2026-09-03
[webview_flutter_tizen] Update to 4.14.1


seungsoo47:

seungsoo47:
I will update this plugin after finishing the work (ewk to wv api migration) currently in progress.

PR #1153, opened 2026-09-16
[webview_flutter_tizen] Update webview_flutter to 4.14.1

seungsoo47:
* Update minimum supported SDK version to Flutter 3.38/Dart 3.10.
* Update webview_flutter to 4.14.1.
* Update webview_flutter_platform_interface to 2.15.1.
* Select the web engine backend from the device's platform version.
* Fix a crash and missing request headers on the WV backend.
* Correct the WV API declarations.

https://github.com/flutter-tizen/plugins/issues/1135


packages/webview_flutter/tizen/src/webview.cc
```diff
@@ -164,6 +159,10 @@ void WebView::Dispose() {
   }
   *is_alive_ = false;
 
+  if (g_current_webview == this) {
+    g_current_webview = nullptr;
+  }
```
JSUYA: Please explain why `current_window` is necessary.
If the intention is to handle multiple WebViews, this part seems a bit weird.
If you create WebView A and then WebView B, and WebView B is removed, `HandleCookieMethodCall` will stop working for WebView A.

packages/webview_flutter/tizen/src/webview.h
```diff
@@ -60,11 +60,14 @@ class WebView : public PlatformView, public WebViewBackend::Delegate {
   // Must be called exactly once, after every WebView has been destroyed.
   static void ShutdownEngine();
 
+  // NOTE: Routes to the most recently created WebView, and fails once that one
+  // is disposed.
```
JSUYA: Is this an issue with the platform API?

packages/webview_flutter/tizen/src/webview.cc
seungsoo47: As discussed, the cookie-related changes will be submitted in a separate PR.

JSUYA (approved):

PR #1157, opened 2026-09-18
[video_player_videohole] Use wrapper api

JSUYA:


packages/video_player_videohole/tizen/inc/ftpw_video_player_videohole.h
```diff
@@ -0,0 +1,121 @@
...
+    int *drm_handle, unsigned int len, unsigned char *pssh_data,
+    void *user_data);
+int ftpw_video_player_videohole_DMGRReleaseDRMSession(
+    DRMSessionHandle_t drm_session);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
xiaowei-guan: The closing `#endif` of the header guard is missing a trailing comment with the guard macro name, suggest change to :
```
#endif  // FLUTTER_TIZEN_PLUGINS_WRAPPER_VIDEO_PLAYER_AVPLAY_H_
```

xiaowei-guan (approved):

PR #1159, opened 2026-09-18
[tizen_rpc_port] Use wrapper API

JSUYA:


xiaowei-guan (commented):
.

packages/tizen_rpc_port/lib/src/parcel.dart
```diff
@@ -587,71 +577,33 @@ class Parcel {
   }
 }
 
-DynamicLibrary? _libRpcPort;
-bool _libRpcPortLoaded = false;
+const int _errorNotImplemented = -38; // -ENOSYS
```
xiaowei-guan: Verified against the shipped binaries: each of the 5 wrapper functions returns `-ENOSYS` (-38) when the underlying `librpc-port.so.1` symbol is unavailable (found the -38 return constant in both aarch64 and i586 builds), so this mapping to `UnsupportedError` preserves the previous behavior on Tizen < 10.0. 👍

packages/tizen_rpc_port/lib/src/parcel.dart
```diff
@@ -587,71 +577,33 @@ class Parcel {
...
-    try {
-      _libRpcPort = DynamicLibrary.open('librpc-port.so.1');
-    } catch (_) {
-      _libRpcPort = null;
-    }
-    _libRpcPortLoaded = true;
-  }
-  return _libRpcPort;
-}
+final DynamicLibrary _libRpcPort = DynamicLibrary.open('libftpw_tizen_rpc_port.so');
```
xiaowei-guan: Non-blocking note: unlike the previous lazy `DynamicLibrary.open('librpc-port.so.1')` (wrapped in try/catch), a missing `libftpw_tizen_rpc_port.so` would now surface as an `ArgumentError` from this initializer rather than `UnsupportedError` at API call time. Not a concern in practice — the wrapper is bundled and also linked into the app via `USER_LIBS += ftpw_tizen_rpc_port` — but worth being aware of if these APIs are ever exercised from host-side unit tests.

xiaowei-guan (commented):
Inline notes from review — both informational, non-blocking.

xiaowei-guan (approved):

PR #1166, opened 2026-09-18
[webview_flutter] Use wrapper API

JSUYA:


packages/webview_flutter/tizen/inc/ftpw_webview_flutter.h
```diff
@@ -0,0 +1,287 @@
...
+typedef struct {
+  const char *key_name;
+  const char *key;
+  const char *string;
+  const char *compose;
+  unsigned int timestamp;
+  wv_modifier_e modifiers;
+  int event_flags;
+  unsigned int key_code;
+} wv_key_event_s;
```
seungsoo47, line 106: wv_key_event_s must match the engine ABI. 
The actual struct has a trailing const char* device_name field.

JSUYA: fixed

